Issue #201, opened 2026-07-14
chore: migrate npm publishing to OIDC trusted publishing (deprecate 2FA-bypass tokens)


rtorcato:
## Why

GitHub/npm are deprecating **2FA-bypass tokens** (classic Automation + granular access tokens with 2FA bypass), which is what all our repos currently use to publish via the Changesets release job (`NPM_TOKEN` secret).

Per the [2026-07-08 changelog](https://github.blog/changelog/2026-07-08-npm-install-time-security-and-gat-bypass2fa-deprecation/):
- **early August 2026** — 2FA-bypass tokens lose account-management abilities.
- **~January 2027** — 2FA-bypass tokens lose **publishing** ability entirely.

npm's recommended replacement is **Trusted Publishing (OIDC)** — the workflow authenticates to npm via GitHub's OIDC identity, no long-lived token. It also gets us provenance for free.

Tokens still publish today (we just republished on an Automation token), so this is **not an outage — it's a deadline**. Target: migrate before Jan 2027.

## ⚠️ Blocking unknown — verify first

Our release path is `changeset publish` → **pnpm** publish. npm OIDC trusted publishing landed in the **npm CLI** (~11.5.1+). **We must confirm pnpm (+ the Changesets action) supports npm OIDC trusted publishing** before committing to the plan.

- [ ] Spike: confirm `pnpm publish` / `changesets/action` can publish via OIDC with no `NPM_TOKEN`.
  - If **yes** → roll out the OIDC workflow below.
  - If **no** → interim options: (a) swap the publish step to the `npm` CLI, (b) stay on a classic **Automation** token until pnpm ships OIDC support (deadline ~Jan 2027).

## Migration steps (per repo)

1. On npmjs.com, add a **Trusted Publisher** to each published package: GitHub Actions → org/repo, workflow filename (`ci.yml`), and environment if used.
2. Ensure the release job has `permissions: id-token: write` (react-common already does).
3. Update the release/publish step to publish via OIDC (drop `NPM_TOKEN`/`NODE_AUTH_TOKEN`); bump npm/pnpm to an OIDC-capable version.
4. Verify a real publish succeeds, then **remove the `NPM_TOKEN` secret** from the repo.

Since the release workflow is scaffolded from `@rtorcato/js-tooling`, do the workflow template change **here first**, then roll out to consumers.

## Repo checklist (npm publishers — confirm/adjust)

- [ ] `js-tooling` — update the scaffolded release workflow template (do first)
[...]

PR #267, opened 2026-07-22
fix(github-actions): drop NPM_TOKEN from generated release job (OIDC)

### LINKED ISSUE #266: fix(github-actions): generator still emits NPM_TOKEN, contradicting the OIDC doctor check (#265)
## Problem

The `github-actions` generator sets `id-token: write` on the release job **but still wires `NPM_TOKEN`/`NODE_AUTH_TOKEN` into the semantic-release step**, so it scaffolds a *half-migrated* OIDC workflow:

`src/cli/generators/github-actions.ts`
```yaml
# L260-264
permissions:
  contents: write
  id-token: write
# L297-302
- name: 🚀 Run semantic-release
  env:
    GITHUB_TOKEN: ${{ secrets.RELEASE_TOKEN || secrets.GITHUB_TOKEN }}
    NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
    NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
  run: npx semantic-release
```

This directly contradicts the `npm OIDC publish` doctor check added in #265, which flags any semantic-release workflow containing `secrets.NPM_TOKEN` as **drift**. The fixer's own hint says *"run `fix github-actions` to drop NPM_TOKEN"* — but the generator re-emits it. So:

- `doctor` reports drift → user runs `fix github-actions` → drift persists (or a clean OIDC repo regresses).
- The doctor check and the fixer disagree about the desired end state.

## Fix

Drop `NPM_TOKEN` and `NODE_AUTH_TOKEN` from the semantic-release step env (keep `id-token: write` and `registry-url`). OIDC trusted publishing needs no npm secret. Update the generator snapshot/tests accordingly.

## Downstream drift (scanned live, public published packages)

The generator was hand-fixed for js-tooling & swift-common in #264, but not the generator itself. Repos still on NPM_TOKEN:

- `shared-docs` — release.yml
- `js-common` — ci.yml
- `cf-common` — ci.yml
- `db-common` — ci.yml, publish-beta.yml
- `browser-common` — ci.yml

Once the generator is fixed, `fix github-actions` clears all five (each already has `id-token: write` + a Trusted Publisher). Until then it can't.

Refs #201, #264, #265.

rtorcato:
## What

The `github-actions` generator set `id-token: write` on the release job but still wired `NPM_TOKEN`/`NODE_AUTH_TOKEN` into the semantic-release step — scaffolding a half-migrated workflow that the `npm OIDC publish` doctor check (#265) flags as **drift**. Net effect: `fix github-actions` re-introduced the exact drift `doctor` reports, and the fixer's own hint ("run `fix github-actions` to drop NPM_TOKEN") was a no-op.

This drops the two token env vars from the generated release step. OIDC trusted publishing needs no npm secret; `id-token: write` stays.

## Downstream

Once released, `fix github-actions` clears the 5 repos still on NPM_TOKEN: `shared-docs`, `js-common`, `cf-common`, `db-common`, `browser-common` (each already has `id-token: write` + a Trusted Publisher).

## Tests

- Updated the release-job test to assert `id-token: write` present and `NPM_TOKEN` absent.
- Regenerated the workflow snapshot.
- Full suite green (452), typecheck + Biome clean.

Closes #266. Refs #201, #264, #265.

### DIFF
--- a/src/cli/generators/github-actions.ts
+++ b/src/cli/generators/github-actions.ts
@@ -295,10 +295,11 @@ ${
           git config --global user.email "github-actions[bot]@users.noreply.github.com"
 
       - name: 🚀 Run semantic-release
+        # npm publish authenticates via OIDC trusted publishing (id-token: write
+        # above) — no npm secret needed. Add a Trusted Publisher for the package
+        # on npmjs.com. See doctor's "npm OIDC publish" check (#201).
         env:
           GITHUB_TOKEN: \${{ secrets.RELEASE_TOKEN || secrets.GITHUB_TOKEN }}
-          NPM_TOKEN: \${{ secrets.NPM_TOKEN }}
-          NODE_AUTH_TOKEN: \${{ secrets.NPM_TOKEN }}
         run: npx semantic-release`
 		: ''
 }
--- a/tests/cli/generators/__snapshots__/snapshot.test.ts.snap
+++ b/tests/cli/generators/__snapshots__/snapshot.test.ts.snap
@@ -236,10 +236,11 @@ jobs:
           git config --global user.email "github-actions[bot]@users.noreply.github.com"
 
       - name: 🚀 Run semantic-release
+        # npm publish authenticates via OIDC trusted publishing (id-token: write
+        # above) — no npm secret needed. Add a Trusted Publisher for the package
+        # on npmjs.com. See doctor's "npm OIDC publish" check (#201).
         env:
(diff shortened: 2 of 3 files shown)


rtorcato:
Superseded — main already dropped NPM_TOKEN from the generated release job (shipped in 2.58.0) with the identical fix. Closing as redundant. Root cause tracked in #266.

Issue #316, opened 2026-07-29
decide: reusable CI workflow vs the generated ci.yml


rtorcato:
## Decision needed — no code until this is settled

Split out of #161, which proposed a reusable workflow consumers call:

```yaml
uses: rtorcato/repo-tooling/.github/workflows/ci.yml@main
```

Today consumers instead get a **generated** `.github/workflows/ci.yml` from `src/cli/generators/github-actions.ts`. It drifts the moment repo-tooling changes, and `doctor` can only flag the drift, not absorb it.

## The two models

**Generated (status quo)** — consumer owns the file, can edit it, works with any CI provider (we also generate GitLab CI). Drift is permanent and manual to reconcile.

**Reusable workflow** — one source of truth, upgrades land automatically. But it locks consumers to GitHub Actions, is hard to customize per repo, and pins them to a ref of *this* repo, which is a supply-chain surface.

## Complication

#315 proposes shipping an `action.yml` that runs `doctor`. If every consumer gates on doctor via that action, drift in a generated `ci.yml` matters much less — the audit catches it. **That may make this issue unnecessary.** Decide #315 first.

## Outcome

Either a design note plus a follow-up implementation issue, or close as wontfix with the reasoning recorded. Do not leave open indefinitely.

rtorcato:
## Decision write-up

Numbers below are measured, not estimated: I rendered `renderGitHubWorkflow(githubJobs(config))` for four real configs and counted, and grepped every `uses:` pin the repo emits.

---

### 1. What the generated `ci.yml` actually is

Rendered from the live generator (`src/base/ci.ts` + `src/languages/js/ci.ts`):

| config | lines |
|---|---|
| library + TS + vitest + bundler + publint + semantic-release (max) | **244** |
| app, no TS, no tests, no bundler, eslint (min) | **92** |
| app, everything but release | 189 |
| max but jest instead of vitest | 238 (8 lines differ from max) |

Inside the 244-line max render: **204 non-blank lines, 105 distinct** — so **99 lines (49%) are verbatim repeats within the single file**. 65 lines (27%) are the checkout → setup-node → pnpm → cache preamble, emitted five times.

**The crux number: exactly 2 of 244 lines (0.8%) contain a value chosen by consumer config.** They are:

```
run: pnpm check      # or `pnpm lint`
run: pnpm coverage   # or `pnpm test`
```

[...]

PR #350, opened 2026-08-08
feat(ci): ship doctor as a composite GitHub Action

### LINKED ISSUE #315: feat(ci): ship repo-tooling as a GitHub Action (action.yml) running doctor
## Why

`doctor` already does the hard part: it sets `process.exitCode = 1` when it finds drift or missing config (`src/cli/commands/doctor.ts:393`) and emits `--json`. `scripts/dogfood.mjs` already proves the pattern end-to-end by running the built CLI against this repo in CI (#273).

What's missing is the three-line on-ramp for *consumer* repos. Today adopting the audit means hand-writing a workflow step; the repo ships no `action.yml`. This is what turns the whole doctor story from advisory into enforced.

## Scope

A composite action at the repo root:

```yaml
- uses: rtorcato/repo-tooling@v2
  with:
    directory: .
    fail-on: drift   # drift | missing | none
```

- Composite, not container — it's `npx @rtorcato/repo-tooling doctor` plus node setup.
- Surface `--json` output as a step output so consumers can post it to a PR comment.
- `fail-on: none` for report-only adoption, so a repo can add the gate before it's clean.

## Notes

Needs a release story: actions are consumed by git ref, so this wants a floating `v2` tag maintained alongside the npm release. Decide whether semantic-release moves the major tag, or it's a separate step in `RELEASING.md`.

Related: #316 (reusable-workflow design call) — these overlap and should be decided together. It's plausible the action makes the reusable workflow unnecessary.

rtorcato:
Closes #315

## What

`action.yml` at the repo root, so a consumer repo gates on `doctor` in two lines:

```yaml
- uses: actions/checkout@v7
- uses: rtorcato/repo-tooling@v3.2.5
```

## Composite, not Docker

`doctor` is a read-only file audit already published as an npm CLI. A container adds an image to build, publish and pull, plus a volume mount to reach the checkout, and buys nothing — `npx` does it in seconds on the host filesystem.

## Surface (deliberately two inputs, one output)

| Name | Kind | Default | Why it exists |
|---|---|---|---|
| `directory` | input | `.` | Mirrors the CLI's `--directory`; monorepos need it. |
| `fail-on` | input | `drift` | The one thing doctor's exit code can't express. |
| `json` | output | — | `doctor --json`, for posting to a PR comment. |

`doctor` sets `process.exitCode = 1` on drift-or-missing, which **is** a usable CI gate — `fail-on: drift` is exactly that, and the default. What it can't express is the adoption path: a repo turning the audit on before it's clean needs `none` (annotate, never fail), and `missing` gates on absent config while tolerating drift. Findings become job annotations at every level (`::error::` for blocking, `::warning::` otherwise).

**No `node-version` input.** The action runs `actions/setup-node` at Node 22 (the CLI's `engines` floor). `doctor` never executes the audited project's code, so the runtime is the action's own concern, not a claim about the consumer's. Documented as a side effect on the calling job.

## Versioning

Consumers pin an exact release tag. The action reads `package.json` from its own checkout and runs that npm version, so the git ref is the only pin and the tag can't diverge from the CLI. No floating `v3` is maintained — a moving major is a second release channel to operate for a benefit Dependabot's `github-actions` ecosystem already delivers. Recorded with reasoning in `RELEASING.md`; adding a floating tag later is backwards-compatible, removing one isn't.

## Guards

- `action.yml` joins the emitted-pin scan from #342 — its `setup-node` pin drifts exactly like a generated workflow's.
- `tests/scripts/doctor-gate.test.ts` runs the gate as a subprocess, because the exit code is the feature; an assertion on a return value would pass while the step silently stopped failing.
- The `build` job smoke-tests the action through `uses: ./` with `fail-on: none` (this repo has genuine self-repo findings that `dogfood` already accounts for), then asserts the `json` output parses and is non-empty.

## Verified locally

`pnpm run verify` — 47 files / 598 tests passed. Simulating the composite's run block against the built CLI: version derived `3.2.5`, doctor exit `1`, `GITHUB_OUTPUT` 8675 bytes, gate exits `1 / 1 / 0` for `drift / missing / none` over 49 checks and 5 findings.

## Left out

Reusable workflow (#316 — this landing is the input to that call), a floating major tag, `node-version` / `json-output-path` / `annotate` inputs, and `$GITHUB_STEP_SUMMARY` rendering.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -335,6 +335,22 @@ jobs:
           # gracefully if the token lacks admin:read, so this never hard-fails.
           GH_TOKEN: ${{ github.token }}
 
+      # Smoke-test the composite action we ship (#315) the way a consumer uses
+      # it. `fail-on: none` because doctor has genuine self-repo findings (see
+      # scripts/dogfood.mjs) — this proves the wiring, dogfood proves the checks.
+      - name: 🎬 Smoke-test the composite action
+        id: action-smoke
+        uses: ./
+        with:
+          fail-on: none
+        env:
+          REPO_TOOLING_ALLOW_SELF: '1'
+
+      - name: 🎬 Assert the action emitted a report
+        run: jq -e '.results | length > 0' <<<"$REPORT" >/dev/null
+        env:
+          REPORT: ${{ steps.action-smoke.outputs.json }}
+
       - name: 📦 Upload build artifacts
         uses: actions/upload-artifact@v7
         with:
--- a/README.md
+++ b/README.md
@@ -72,6 +72,14 @@ See the [Getting Started guide](https://rtorcato.github.io/repo-tooling/guides/g
 | `doctor` | Diagnose an existing project for missing or drifted tooling. | `npx @rtorcato/repo-tooling doctor` |
 | `fix [target]` | Apply scaffolders for what `doctor` flagged (`--yes`, `--dry-run`, `--diff`). | `npx @rtorcato/repo-tooling fix` |
 
+Prefer to run the audit in CI? `doctor` also ships as a GitHub Action:
+
+```yaml
+- uses: rtorcato/repo-tooling@v3.2.5
+```
+
+See the [GitHub Actions reference](https://rtorcato.github.io/repo-tooling/reference/github-actions/#run-doctor-as-a-github-action) for its inputs and outputs.
+
 Every command takes `-d, --directory <path>`; run any with `--help` for its full flags. Run `list` (or `list --json`) for the full set of `fix` targets — it's the source of truth. Notable ones include `fix docs-site` (scaffold a [Docusaurus docs site](https://rtorcato.github.io/repo-tooling/guides/docs-site/)) and `fix bun` (Bun runtime config).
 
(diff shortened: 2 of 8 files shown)


rtorcato:
CI evidence for the action itself (run 31260149182, `build` job):

The first push **failed** at the smoke step with `sh: 1: repo-tooling: not found` (exit 127). Real bug, not a test artifact: `npx` treats a spec already satisfied by the *current project* as installed and runs its bin from `node_modules/.bin`, which a repo that only declares the dependency hasn't linked. Reproduced locally from this repo's root, fixed in 596ffe4 by resolving `directory` to an absolute path and running `npx` from `$RUNNER_TEMP`.

After the fix, all 14 checks pass. The smoke step logged:

```
doctor: 49 checks, 5 finding(s), 0 blocking at fail-on=none
```

and produced five job annotations — Commitlint, semantic-release and Vitest as `drift`, TypeScript as `missing`, package.json as `drift` — all at `warning` level, i.e. surfaced without failing, which is what `fail-on: none` promises. Those are the same five entries `scripts/dogfood.mjs` accepts as self-repo divergence. The follow-up step confirmed the `json` output parsed with a non-empty `.results`.

Issue #288, opened 2026-07-27
feat(swift): swift-library preset + docs


rtorcato:
Add `setup --preset swift-library` and a Swift guide to the docs site, mirroring the JS preset/docs pattern once the module and CI generation land.

Issue #286, opened 2026-07-27
feat(swift): src/languages/swift module end-to-end


rtorcato:
First non-JS language module — **proves the base/module seams** (#139). `swift-common` is the real consumer target.

- `detect`: `Package.swift` (already a marker in detect-language.ts).
- `checks`: SwiftFormat / SwiftLint config presence + drift, `.gitignore` (Swift template), `.swift-version`, `Package.swift` hygiene.
- `fixers`: scaffold the above.

Depends on the core-architecture milestone (registry + de-noded generators).

PR #356, opened 2026-08-09
feat(python): add src/languages/python module

### LINKED ISSUE #290: feat(python): src/languages/python module
Python language module on the same template as Swift (#139).

- `detect`: `pyproject.toml` / `setup.py` (already markers).
- `checks`/`fixers`: ruff + black + pytest + mypy config, `.gitignore` (Python), CI steps + CodeQL `language: python`.

Depends on the core-architecture seams; follows the Swift module as the reference.

rtorcato:
Closes #290.

Python language module on the Swift template (#286): `doctor`/`fix` dispatch to it for any repo with a `pyproject.toml` or `setup.py`, and `python` flips to `supported: true` in the registry.

## What lands

| Check | Absent | Fix target |
|---|---|---|
| `pyproject.toml` | `missing` | — (project metadata) |
| Ruff | `missing` | `ruff` |
| mypy | `missing` | `mypy` |
| pytest | `missing` | `pytest` |
| Python `.gitignore` | `missing` | `python-gitignore` |
| Python tests | `missing` | — (write tests, or `python-ci`) |
| Git hooks / Pre-push | `optional-missing` | `python-git-hooks` |

Plus `python-ci` (ruff / mypy / pytest, matrixed over the interpreters `requires-python` declares) and `python-gitlab-ci`. CodeQL `language: python` and Dependabot `package-ecosystem: pip` already came from the registry — the shared `fix codeql` / `fix dependabot` targets needed no change.

Three presets under `tooling/python/`: `ruff.toml`, `mypy.ini`, `pytest.ini`, all reachable via `copy` too.

## Two deliberate deviations from the issue

**No black.** The issue lists "ruff + black". Ruff's formatter is black-compatible and already runs in the pre-commit hook and the lint job; a second *rewriting* formatter fights the first. That's the same call the Swift module already makes about SwiftFormat, so this follows existing precedent rather than inventing one. flake8 / isort / pyupgrade are likewise absent — the shipped `select` list covers all three. Documented in the module header and the docs page. Say the word and I'll add black as an optional slot the way `swift-format` is.

**No `python-lockfile` fixer.** Swift's writes the config `setup --preset swift-library` would have produced, and there's no `python-library` preset (the Swift one was its own issue, #288). Fabricating a `ProjectConfig` here would put invented JS tool choices in a Python repo's lockfile — worse than doctor reporting the lockfile absent, which is true. So `lockfile` shows as uncovered for Python; the test asserts that explicitly.

## Two things worth a look

**The tool-config check isn't a plain `FileCheck`.** Every Python repo has a `pyproject.toml`, and `checkFile` reads "candidate exists but doesn't match" as drift — so a repo that simply hasn't configured ruff reported `pyproject.toml found but is not a valid Ruff configuration`. Caught it in the end-to-end smoke run, not the unit tests. Now dedicated files (`ruff.toml`, `mypy.ini`, `pytest.ini`) can produce drift; shared files (`pyproject.toml`, `setup.cfg`, `tox.ini`) can only upgrade the result to `ok`. Dedicated wins on precedence, matching what the tools actually do at runtime.

**Shared hooks installer.** Swift and Python want the same write-chmod-`git config core.hooksPath` dance, so it moved to `src/base/git-hooks.ts` and Swift now calls it. Net -48 lines in the Swift module, no behaviour change.

## Verification

`check` + `typecheck` + 697 tests + `build-cli` all green (pre-push ran the full gate). Also smoke-tested the built CLI end to end against a scratch Python repo: `doctor` → `fix --yes` → `doctor` leaves only `lockfile` and `README badges`, both documented as uncovered.

## Follow-up filed separately

`fix --json` output is corrupted whenever a fixer prints an advisory to stdout — `console.log` inside `run()` isn't gated on `silent`. Pre-existing across ~21 sites in the JS and Swift modules; the Python hooks fixer follows the same pattern rather than diverging in one place. Filed on its own.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -6,7 +6,7 @@ Orientation for coding agents working with `@rtorcato/repo-tooling`. Human-reada
 
 A one-package JavaScript / TypeScript tooling distribution. Ships every preset (TypeScript, Biome, ESLint, Prettier, Vitest, Jest, Commitlint, semantic-release, tsup, esbuild, Vite, Playwright) plus a CLI to scaffold and audit projects. Consumers get one install.
 
-**Swift** repos (detected via `Package.swift`) are covered end to end: `setup --preset swift-library` scaffolds a SwiftPM package, and `doctor`/`fix` run the language-agnostic checks plus SwiftLint / Periphery / `.gitignore` / `Package.swift`. Python and Perl are audit-only for now. See `src/languages/` — one directory per language module, `src/base/` for what's shared.
+**Swift** repos (detected via `Package.swift`) are covered end to end: `setup --preset swift-library` scaffolds a SwiftPM package, and `doctor`/`fix` run the language-agnostic checks plus SwiftLint / Periphery / `.gitignore` / `Package.swift`. **Python** repos (detected via `pyproject.toml` / `setup.py`) get `doctor`/`fix` — Ruff / mypy / pytest / `.gitignore` / CI / git hooks — but no `setup` preset yet. Perl is audit-only for now. See `src/languages/` — one directory per language module, `src/base/` for what's shared.
 
 ## CLI surface (agent-friendly)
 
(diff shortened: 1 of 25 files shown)


rtorcato:
Rebase note: `main` moved under this branch (#354/#355 + the 3.5.0 release), and #355 added a required `language` field on `BaseCheckOptions` plus a `Monorepo` base check — the first CI run failed to compile because of it. Brought up to date with a merge rather than a force-push; `69aef87` threads `language` through the Python dispatch and accounts for `Monorepo` in the fixer-coverage test.

The `fix --json` stdout bug mentioned above is filed as #357.

All checks green.

PR #361, opened 2026-08-09
fix(cli): finish routing fixer advisories off stdout

### LINKED ISSUE #357: fix --json output is corrupted by fixer advisories written to stdout
`fix --json` promises parseable JSON on stdout (AGENTS.md documents it as the agent-friendly contract, and `--json` implies `--yes` precisely so prompts can't corrupt the stream). But a fixer's own `console.log` isn't gated on `silent`, so any fixer that prints an advisory writes it into the middle of the JSON payload.

Reproduced while adding the Python module (#290) — piping `fix --yes --json` into a parser dies on:

```
   run `git config core.hooksPath .githooks` once per clone — it's local git config, not a committed file
   ^
SyntaxError: Unexpected token 'r', "   run `git c"... is not valid JSON
```

`applyFixer` in `src/cli/commands/fix.ts` gates *its* logging on `silent`, but the `run()` bodies write to stdout directly. Roughly 21 sites:

- `src/languages/js/fixers.ts` — ~18 (`no package.json found — skipping`, `reminder: run pnpm install`, etc.)
- `src/languages/swift/fixers.ts` — 2 (`swift-git-hooks`, `docc`)
- `src/languages/python/fixers.ts` — 1 (`python-git-hooks`)

Only bites when the advisory branch actually fires, which is why it survived: the common paths stay quiet.

**Likely fix:** these are human-facing notes, not results — `console.error` is the right stream for them regardless, and stdout stays clean for the JSON. One-word change per site. Worth checking nothing asserts on stdout for those strings first, and that interactive output still reads sensibly with the notes on stderr.

Alternative if the interleaving turns out to matter: thread `silent` onto `FixerContext` so `run()` can suppress them. More plumbing, but keeps every human-facing line on one stream.

Not fixed inline in #356 — it predates that PR across two other modules, and half-fixing it in the new module only would leave the three inconsistent for no gain.

rtorcato:
Closes #357 (reopened — #358's fix was incomplete).

#358 routed the fixer advisories to stderr, but two reachable sites kept writing to stdout, so `fix --json` is still unparseable on those paths.

## What was missed, and why

**`src/languages/python/fixers.ts`** — #358 was authored while the Python module (#290) was still in review, and said so explicitly: *"17 sites, not the ~21 the issue estimated: `src/languages/python/fixers.ts` doesn't exist on `main` yet."* Accurate when written. #290 then merged at 14:50:56 and #358 at 14:51:12 — sixteen seconds later — so the new module landed carrying the exact bug the rest of the CLI had just lost.

**`src/base/github-settings.ts`** (5 sites) — `applyGithubSettings` is called from the `github-settings` base fixer's `run()`, so its advisories land in the payload like any fixer's. Out of #358's scope because that grep covered the fixer *files*, and this isn't one; it's just called by one.

Both are the same class of miss — an enumeration kept in sync by hand.

## The guard

So the invariant is stated over the **trees** rather than a file list:

```
nothing under src/base or src/languages writes to stdout
```

A new language module is covered the day it lands, with nothing to remember to update. `src/cli/commands` is deliberately outside it — printing is its job, and it already gates on `silent`.

Advisories still print. They go to stderr, which is where diagnostics belong whether or not `--json` is set — same reasoning #358 gave for preferring this over threading `silent` through `FixerContext`.

## Verification

Mutation-checked the scan rather than trusting it: reverting one site to `console.log` fails it naming the offending file, and it passes again once restored. Two `*/`-inside-a-block-comment typos of mine also got caught along the way — the first made the scan file itself parse as zero tests, which is a good reminder that a green run and a *ran* test aren't the same thing.

Also adds the behavioural test on the Python path, mirroring #358's: `fix python-git-hooks --json` must emit exactly one stdout write that parses.

`check` + `typecheck` + 706 tests + `build-cli` green (pre-push ran the full gate).

### DIFF
--- a/src/base/github-settings.ts
+++ b/src/base/github-settings.ts
@@ -479,17 +479,22 @@ export function buildGhApplyCommands(state: GhApplyState): GhCommand[] {
  * on skip (no gh/auth/remote, or already compliant), logging the reason. Mirrors
  * the "no package.json found — skipping" fixer pattern. Read-only unless a delta
  * exists, so re-runs (walk-all hits it up to 3×) are safe.
+ *
+ * Reached from the `github-settings` fixer, so its advisories go to
+ * `console.error` for the same reason the fixers' do (#357): stdout carries the
+ * `--json` payload. #358 fixed the fixer files but not this one — it isn't a
+ * fixer file, it's just called by one.
  */
 export async function applyGithubSettings(dir: string, exec?: GhExec): Promise<string[]> {
 	if (!(await fs.pathExists(path.join(dir, '.git')))) {
-		console.log(chalk.gray('   skipped — not a git repository'))
+		console.error(chalk.gray('   skipped — not a git repository'))
 		return []
 	}
 	// Bind gh's cwd to the target dir so its repo resolution honors `-d` (#218).
 	const gh: GhExec = exec ?? ((args, stdin) => realGhExec(args, stdin, dir))
 	const probe = await probeRepo(gh)
 	if ('skip' in probe) {
-		console.log(chalk.gray(`   skipped — ${probe.skip}`))
+		console.error(chalk.gray(`   skipped — ${probe.skip}`))
 		return []
 	}
 	const { info } = probe
@@ -511,7 +516,9 @@ export async function applyGithubSettings(dir: string, exec?: GhExec): Promise<s
 		const r = await gh(cmd.args, cmd.stdin)
 		if (r.ok) applied.push(cmd.label)
 		else
-			console.log(chalk.yellow(`   could not apply ${cmd.label}: ${r.stderr.trim() || 'gh error'}`))
+			console.error(
+				chalk.yellow(`   could not apply ${cmd.label}: ${r.stderr.trim() || 'gh error'}`)
+			)
 	}
 
 	// Code-scanning ruleset (#269): POST only when CodeQL is on and no active gate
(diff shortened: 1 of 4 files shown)


rtorcato:
Rebased onto `main` (now 3.7.0, with the Perl module #359 landed). Scope has narrowed:

**#360 fixed the Python site independently** while this was open, so that hunk is main's now. What's left here is the part nothing else covers — `src/base/github-settings.ts`, still **5 stdout writes on `main`**. `applyGithubSettings` runs from the `github-settings` base fixer, so `fix github-settings --json` and any walk-all that reaches it still emit unparseable output today.

**On the duplicated guard.** #360 added its own source scan over `src/base/fixers.ts` plus each module's fixers file. That one passes on `main` right now *while github-settings.ts is still broken* — it can't see the file, because the file isn't a fixer. I dropped it in favour of the tree-wide invariant in `tests/base/stdout-discipline.test.ts` (all of `src/base` + `src/languages`), which is a strict superset. Two overlapping guards where the narrower one reports green on a broken tree is worse than one that actually holds.

Kept the Python runtime test — #360 added no behavioural coverage for that path, only the scan.

Re-ran the mutation check against latest main: reverting the `already configured — nothing to apply` line to `console.log` fails the guard naming `src/base/github-settings.ts`, and passes once restored. 775 tests green.

Third pass at #357, and each miss was the same shape: a hand-maintained file list. That's the thing this replaces.

Issue #364, opened 2026-08-09
Generated CI workflow fails out of the box: references unscaffolded scripts and no packageManager for pnpm/action-setup


rtorcato:
## Summary

`repo-tooling fix github-actions` generates `.github/workflows/ci.yml` that cannot pass on a repo scaffolded with the same tool. It failed three times in a row, each for a different reason, before going green.

## 1. `pnpm/action-setup` has no version to resolve

The workflow uses:

```yaml
- name: 📦 Setup pnpm
  uses: pnpm/action-setup@v6
```

with no `version:` input. `fix github-actions` doesn't add a `packageManager` field to `package.json`, and nothing else in the toolchain does either, so the action has no source of truth:

```
Error: No pnpm version is specified.
Please specify it by one of the following ways:
  - in the GitHub Action config with the key "version"
  - in the package.json with the key "packageManager"
  - in the package.json with the key "devEngines.packageManager"
```

Every job dies at setup. Fixed locally by adding `"packageManager": "pnpm@11.20.0"`.

**Suggestion:** have the scaffolder write `packageManager` from the detected pnpm version, the same way `fix engines` writes `engines.node`.

## 2. The workflow calls scripts no `fix` target creates

The generated jobs run:

```yaml
- run: pnpm check     # line 89
- run: pnpm knip      # line 92
- run: pnpm coverage  # line 143
```

On a project where `fix github-actions` was applied without `fix knip` or a coverage-producing target, these don't exist:

```
[ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL] Command "coverage" not found
```

`fix biome` doesn't add a `check` script either, so `pnpm check` is missing too unless the user writes it.

**Suggestion:** either generate the CI steps conditionally from the scripts that actually exist in `package.json`, or have `fix github-actions` declare its prerequisites and add the missing scripts (the way `fix verify` composes a chain from detected tools). Failing that, `doctor` could flag "CI references scripts that don't exist".

Related: `fix verify` currently skips with *"not enough tools enabled to compose a verify chain — skipping (need 2+ of typecheck/lint/tests)"* even on a repo that has both `typecheck` and `test` scripts plus a scaffolded `biome.json` — it appears to want a `lint`/`check` script that no target provides, so the two gaps compound.

## 3. Install step fails on undecided pnpm build approvals

`pnpm install --frozen-lockfile` exits non-zero when build scripts are neither approved nor declined:

```
[...]

PR #374, opened 2026-08-10
fix(js): add the scripts that run eslint and prettier

### LINKED ISSUE #371: fix eslint and fix prettier scaffold configs with no script to run them
## Summary

`fix biome` was fixed in #364 to add the scripts that run Biome (`lint`, `format`, `check`, `check:fix`) alongside the config. `fix eslint` and `fix prettier` still have the original gap — they write a config file and nothing else.

`src/languages/js/fixers.ts`:

```ts
target: 'eslint',
outputs: ['eslint.config.mjs'],
async run({ targetDir, pkg }) {
  await generateESLintConfig(inferProjectConfig(pkg), targetDir)
  return { filesWritten: ['eslint.config.mjs'] }   // no scripts
},
```

`prettier` is the same shape.

## Why it bites

A repo that runs `fix eslint` gets a linter it cannot invoke — there is no `pnpm lint`. Three things then quietly go wrong:

1. **The CI lint step vanishes rather than failing.** Since #364, `githubJobs` only emits a step when the script exists. The eslint lint command is `pnpm lint` (`lintScript()` in `src/languages/js/ci.ts`), so with no `lint` script the step is dropped — and if `knip` is absent too, the whole `lint` job is dropped. The pipeline goes green having linted nothing, which is worse than the old loud failure.
2. **`fix verify` can't count the lint half.** `composeVerifyScriptFromPkg` looks for `scripts.check` then `scripts.lint` (`src/cli/generators/package-json.ts`); neither exists, so a repo with typecheck + tests + eslint composes a 2-command chain that silently omits linting.
3. **`setup` disagrees with `fix`.** `getScripts()` already writes `lint` / `lint:fix` / `format` for eslint and `format` for prettier, so a scaffolded project has them and a `fix`-ed one doesn't.

## Suggested fix

Mirror what the `biome` target already does — `ensureScripts` is already exported from `src/cli/generators/package-json.ts` and only fills in names that are missing, so a repo's own commands are never overwritten:

```ts
const ESLINT_SCRIPTS = { lint: 'eslint .', 'lint:fix': 'eslint . --fix' }
const PRETTIER_SCRIPTS = { format: 'prettier --write .' }
```

Keep these in step with the eslint/prettier branches of `getScripts()` in `package-json.ts` — that's the source they should match. Update each fixer's `outputs` to mention `package.json (scripts)` and set `riskLevel: 'safe-merge'`, as the `biome` target does.

## Verification

- A test that `fix eslint` adds `lint` without clobbering an existing one (mirror `fix biome scripts` in `tests/cli/commands/fix.test.ts`).
- A test that a repo with eslint + typecheck + vitest now composes a verify chain including `pnpm lint`.

rtorcato:
`fix eslint` and `fix prettier` wrote a config and nothing else, so a repo got a linter it could not invoke. Since #364 the generated CI only emits a step when the script exists, so the lint step vanished instead of failing, and `fix verify` composed a chain with no linting in it.

Both fixers now call `ensureScripts`, which only fills in names that are missing, so a repo's own commands survive. The commands match the eslint branch of `getScripts()` in `src/cli/generators/package-json.ts`.

Left `riskLevel` alone. The issue suggested `safe-merge`, but these fixers still overwrite the config file itself, so that prompt wording ("existing fields preserved") would be wrong and would default a drift overwrite to Yes. `fix biome` does not set it either.

Tests: three cases in `tests/cli/commands/fix.test.ts` — eslint scripts added without clobbering, prettier `format` added, and a repo with eslint + typecheck + vitest composing a verify chain that includes `pnpm lint`.

Closes #371

### DIFF
--- a/src/languages/js/fixers.ts
+++ b/src/languages/js/fixers.ts
@@ -37,6 +37,17 @@ const BIOME_SCRIPTS: Record<string, string> = {
 	check: 'biome check .',
 	'check:fix': 'biome check --fix .',
 }
+
+/** Kept in step with the eslint branch of getScripts() in package-json.ts. */
+const ESLINT_SCRIPTS: Record<string, string> = {
+	lint: 'eslint .',
+	'lint:fix': 'eslint . --fix',
+}
+
+/** Kept in step with the eslint branch of getScripts() in package-json.ts. */
+const PRETTIER_SCRIPTS: Record<string, string> = {
+	format: 'prettier --write .',
+}
 import {
 	WORKSPACE_FILE,
 	dependsOnEsbuild,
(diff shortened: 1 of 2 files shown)


rtorcato (commented):
Security review: no issues found.

- Adds static, hardcoded script strings (`ESLINT_SCRIPTS`, `PRETTIER_SCRIPTS`) written via the existing `ensureScripts` helper — no user-controlled input reaches these values, no shell/SQL construction, no injection surface.
- No secrets introduced or logged.
- No new/changed dependencies (supply chain unaffected).
- Deliberately leaving `riskLevel` unset instead of `safe-merge` (per PR description) is a sound call — these fixers still overwrite the config file itself, so `safe-merge` wording would be misleading.
- Test coverage (3 new cases) matches the fix.

No blocking findings.

rtorcato (commented):
Verified against issue #371 and the current \`fixers.ts\`.

**Correctness:** Both fixers now call \`ensureScripts\` with script maps matching the eslint/prettier branches of \`getScripts()\`, mirroring the pattern \`fix biome\` already uses (verified in the current \`src/languages/js/fixers.ts\`, lines 154-169) — \`ensureScripts\` only fills in missing script names (confirmed in \`package-json.ts\`), so existing repo scripts survive, matching the first test case.

**Deviation from the issue is justified:** the issue suggested \`riskLevel: 'safe-merge'\`, but the PR leaves it unset. Confirmed that the \`biome\` fixer — the pattern being mirrored — also leaves \`riskLevel\` unset, so this is consistent with existing precedent, not an inconsistency. The stated reasoning (the fixer still overwrites the config file itself, so "existing fields preserved" prompt wording would mislead a drift-overwrite prompt into defaulting to Yes) is sound.

**Tests:** the three new cases in \`tests/cli/commands/fix.test.ts\` directly cover the issue's two verification bullets (scripts added without clobbering, and a verify chain that now includes \`pnpm lint\`), plus a prettier case.

No bugs found. Minor nit only: the two new \`const\` script maps are inserted before the file's \`import\` statement — harmless since import declarations hoist, and matches the existing \`BIOME_SCRIPTS\` placement already in the file, so not worth blocking on.

PR #375, opened 2026-08-10
fix(doctor): flag transitive pnpm build approvals

### LINKED ISSUE #373: pnpm build approvals check misses transitive dependencies
## Summary

The `pnpm build approvals` check added in #364 only inspects **direct** dependencies. This is a documented limitation in `src/languages/js/checks.ts`:

> Only direct dependencies are inspected: reaching transitive ones means walking node_modules/.pnpm, and the packages that trip this in practice (better-sqlite3, @prisma/client, esbuild) are ones the repo asked for by name. A transitive offender is a false negative, never a false positive.

The reasoning holds, but the false negative is real: pnpm's `ERR_PNPM_IGNORED_BUILDS` names transitive packages too, and when it does, CI fails while `doctor` reports the repo clean.

## Reproduction

A repo depending on something that pulls a native module in transitively — e.g. a package depending on `sharp` or `esbuild` without listing it directly. `pnpm install --frozen-lockfile` fails; `doctor` says `every dependency with a build script has a decision`.

## Suggested fix

Walk `node_modules/.pnpm/*/node_modules/<name>/package.json` rather than only `node_modules/<name>/package.json`. Each `.pnpm` entry is `<name>@<version>` (scoped packages use `+` for `/`, e.g. `@prisma+client@5.22.0`), so the package name has to be parsed back out of the directory name — or read `name` from the `package.json` itself, which is simpler and exact.

Watch for:
- **Deduplication.** The same package appears once per version; report a name once, not per copy.
- **Cost.** `.pnpm` can hold thousands of entries. Read `package.json` only, never recurse into the trees, and skip the whole pass when `.pnpm` is absent (npm/yarn repos, or a repo that hasn't installed).
- **Noise.** Transitive build scripts are common and mostly harmless. Consider reporting direct ones as `drift` (actionable — the repo chose them) and transitive ones separately, so the check stays worth reading. `optional-missing` may fit better for the transitive set.

Keep the existing guard that returns early when `node_modules` is absent — the manifest alone can't say which packages ship an install script.

## Verification

Extend the `checkBuildApprovals` tests in `tests/cli/commands/doctor.test.ts`: a transitive package with a `postinstall` and no decision is flagged; one already listed under `allowBuilds` is not; the same package at two versions is reported once.

rtorcato:
The `pnpm build approvals` check only inspected direct dependencies, so a repo pulling a native module in transitively reported clean while `pnpm install --frozen-lockfile` failed CI with `ERR_PNPM_IGNORED_BUILDS`.

Adds a second pass over `node_modules/.pnpm/*/node_modules/<name>/package.json`:

- One `package.json` read per unique package name — never a recursive walk into the store trees.
- The whole pass is skipped when `.pnpm` is absent (npm/yarn repos, or nothing installed).
- A package present at two versions is reported once.
- The existing early return when `node_modules` is absent is unchanged.

Store directory names are decoded back to package names (`@prisma+client@5.22.0` → `@prisma/client`), splitting on the *first* `@` after index 0 so peer-suffixed entries like `@babel+core@7.0.0_supports-color@8.0.0` decode correctly. The name is then taken from the manifest itself, with the decoded name as fallback.

Direct and transitive offenders are graded apart: direct stays `drift` (the repo asked for the package by name), transitive is `optional-missing`, since transitive build scripts are common enough that grading them as drift would drown the report.

## Verification

Six new `checkBuildApprovals` cases plus a `pnpmStoreDirToName` unit test. `pnpm run check`, `tsc --noEmit --project src/cli/tsconfig.json` and `vitest run` (787 passed, 18 skipped) all pass.

Closes #373

### DIFF
--- a/src/languages/js/checks.ts
+++ b/src/languages/js/checks.ts
@@ -1008,9 +1008,18 @@ async function undecidedTransitiveBuilds(
 		// A package present at two versions has two store dirs; report it once.
 		if (!name || seen.has(name) || skip(name)) continue
 		seen.add(name)
+		// Store directory names come from registry metadata, and only the *first*
+		// `+` is decoded — the rest of the entry passes through literally, so a
+		// hostile name (`..+..@1.0.0` → `../..`) would climb out of the package
+		// directory and have `fs.readJson` read an unrelated file, whose `name`
+		// then lands in the report. Confirm the resolved path stays under this
+		// entry's `node_modules` before reading it.
+		const root = path.resolve(store, entry, 'node_modules')
+		const target = path.join(root, name, 'package.json')
+		if (!target.startsWith(root + path.sep)) continue
 		let depPkg: { name?: string; scripts?: Record<string, string> }
 		try {
-			depPkg = await fs.readJson(path.join(store, entry, 'node_modules', name, 'package.json'))
+			depPkg = await fs.readJson(target)
 		} catch {
 			continue
 		}
--- a/src/languages/js/checks.ts
+++ b/src/languages/js/checks.ts
@@ -974,17 +974,65 @@ export function decidedBuilds(yaml: string): Set<string> {
 }
 
 /**
- * Direct dependencies whose install would run a build script (#364).
+ * Package name encoded in a `node_modules/.pnpm` directory name (#373).
+ *
+ * Entries read `<name>@<version>` with `/` written as `+`, and the version may
+ * carry a peer suffix that contains further `@`s
+ * (`@babel+core@7.0.0_supports-color@8.0.0`), so the separator is the *first*
+ * `@` after index 0 — never the last.
+ */
+export function pnpmStoreDirToName(entry: string): string | null {
+	const at = entry.indexOf('@', 1)
+	if (at < 1) return null
+	return entry.slice(0, at).replace('+', '/')
+}
+
+/**
+ * Transitive packages that ship a build script and have no decision (#373).
+ *
+ * One `package.json` read per unique name — never a recursive walk — and the
+ * whole pass is skipped when `.pnpm` is absent (npm/yarn repos, or nothing
+ * installed). `skip` drops the names the direct pass already owns.
+ */
+async function undecidedTransitiveBuilds(
+	modulesDir: string,
+	skip: (name: string) => boolean
+): Promise<string[]> {
+	const store = path.join(modulesDir, '.pnpm')
+	if (!(await fs.pathExists(store))) return []
+
+	const seen = new Set<string>()
+	const undecided = new Set<string>()
+	for (const entry of await fs.readdir(store)) {
+		const name = pnpmStoreDirToName(entry)
+		// A package present at two versions has two store dirs; report it once.
+		if (!name || seen.has(name) || skip(name)) continue
+		seen.add(name)
+		let depPkg: { name?: string; scripts?: Record<string, string> }
+		try {
+			depPkg = await fs.readJson(path.join(store, entry, 'node_modules', name, 'package.json'))
+		} catch {
+			continue
+		}
+		const scripts = depPkg.scripts ?? {}
+		// `name` from the manifest is exact; the decoded dir name is the fallback.
+		if (BUILD_LIFECYCLE_SCRIPTS.some((s) => scripts[s])) undecided.add(depPkg.name ?? name)
+	}
+	return [...undecided].filter((n) => !skip(n)).sort()
+}
+
+/**
+ * Dependencies whose install would run a build script (#364, #373).
  *
  * pnpm 11 turned undecided build scripts into a hard error, so
  * `pnpm install --frozen-lockfile` exits non-zero with ERR_PNPM_IGNORED_BUILDS
  * — a guaranteed CI failure. Locally the same message reads as advisory, which
  * is exactly why it gets missed until the pipeline goes red.
  *
- * Only direct dependencies are inspected: reaching transitive ones means
- * walking node_modules/.pnpm, and the packages that trip this in practice
- * (better-sqlite3, @prisma/client, esbuild) are ones the repo asked for by
- * name. A transitive offender is a false negative, never a false positive.
+ * Direct and transitive offenders are graded apart. A direct one is `drift`:
+ * the repo asked for the package by name and can answer for it. A transitive
+ * one is `optional-missing` — still worth recording, but common enough that
+ * grading it as drift would drown the report in noise.
  */
 export async function checkBuildApprovals(dir: string, pkg: Pkg | null): Promise<CheckResult> {
 	const check = 'pnpm build approvals'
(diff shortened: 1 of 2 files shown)


rtorcato (commented):
Reviewed diff against #373's constraints.

**Correctness**
- `undecidedTransitiveBuilds` reads exactly one `package.json` per unique top-level `.pnpm` entry (`store/<entry>/node_modules/<name>/package.json`) — no recursive walk into the store trees, matching the issue's cost constraint.
- Pass is skipped up front via `fs.pathExists(store)` when `.pnpm` is absent (npm/yarn repos, nothing installed).
- Dedup: `seen` set keyed on decoded name skips repeat reads for a package present at multiple store versions (verified by the "reports a package present at two versions once" test, asserting exactly one occurrence in the detail string).
- `pnpmStoreDirToName` correctly splits on the first `@` after index 0, so peer-suffixed entries like `@babel+core@7.0.0_supports-color@8.0.0` decode to `@babel/core` rather than truncating at the peer's `@`. Non-conforming entries (e.g. `node_modules`) return `null` and are safely skipped — good defensive handling of stray `.pnpm` top-level entries.
- Direct vs. transitive grading matches the issue's suggestion: direct undecided → `drift`, transitive-only → `optional-missing`, combined detail when both are present. Confirmed by the "grades a direct offender as drift even when transitive ones exist" test.
- A direct dependency that also happens to sit in the store isn't double-reported — `skip` excludes anything already in `deps` or `decided` before it's even read, which also avoids a wasted read.
- The existing early return for absent `node_modules` is untouched by this diff, so that guard is preserved as required.

**Tests**
Six `checkBuildApprovals` cases plus a dedicated `pnpmStoreDirToName` unit test cover the transitive-flagging, scoped-name decoding, allowBuilds suppression, dedup-at-two-versions, direct+transitive combined grading, and no-double-report-for-direct-in-store scenarios called out in the issue.

No bugs found. This is clean, closely follows the issue's stated approach, and the perf/dedup/skip constraints are all correctly implemented.

rtorcato (commented):
**Security review**

Scope: `pnpmStoreDirToName` + `undecidedTransitiveBuilds` in `src/languages/js/checks.ts`.

**Medium — unvalidated path traversal via decoded store-dir name**
`pnpmStoreDirToName` splits `entry` on the first `@` and does `.replace('+', '/')` (single, non-global replacement). The result (`name`) is then used unchecked in `path.join(store, entry, 'node_modules', name, 'package.json')` before `fs.readJson`. There's no check that the decoded name stays within `store`.

Under normal pnpm output this is safe (one `+` per scoped name), but `.pnpm` directory names ultimately derive from package names/versions pulled from third-party registry metadata — exactly the kind of input this scan is meant to treat as untrusted. An entry with a crafted name (extra `+`/`..` segments, only the *first* `+` gets decoded, the rest passes through literally) can make `name` resolve outside `store` via `path.join`'s normalization, causing `fs.readJson` to read an arbitrary JSON file elsewhere on disk. Since `depPkg.name ?? name` feeds directly into the report `detail` string, that file's `name` field would leak into `doctor`'s output — a real (if narrow) path traversal / info-disclosure primitive, not RCE.

Suggested fix: after building the path (or after decoding `name`), verify it resolves inside `store` before reading, e.g.
```ts
const target = path.join(store, entry, 'node_modules', name, 'package.json')
if (!target.startsWith(path.resolve(store) + path.sep)) continue
```
or simpler, reject any decoded `name` containing `..`.

**Handled correctly**
- Malformed/non-JSON `package.json` under a store entry: `fs.readJson` is wrapped in `try/catch` and the entry is skipped — no crash on hostile JSON.
- No recursive walk into `.pnpm`: one `readdir` + one `readJson` per unique package name, matching the stated cost constraint; pass is skipped entirely when `.pnpm` is absent.
- No secrets, no shell/SQL construction, no new/changed dependencies in this diff.

No other findings.

rtorcato (commented):
**Re-review (security): fix verified.**

The path-traversal fix in \`f3ab578\` holds up.

- \`pnpmStoreDirToName\` is unchanged (still a single, non-global \`.replace('+', '/')\`), so a hostile store entry like \`..+..@1.0.0\` still decodes to \`../..\`. What's new is the containment guard around the read: \`root = path.resolve(store, entry, 'node_modules')\`, \`target = path.join(root, name, 'package.json')\`, then \`if (!target.startsWith(root + path.sep)) continue\`.
- This is the correct pattern: \`path.resolve\`/\`path.join\` normalize \`..\` segments before the comparison (not a raw string check), and the trailing \`path.sep\` in the prefix comparison avoids the classic \`/store/pkg\` vs \`/store/pkg-evil\` false-positive-pass bug. It's applied before the single \`fs.readJson\` call site in \`undecidedTransitiveBuilds\` — there's no other read path to miss.
- Traced the max escape distance: since only the *first* \`+\` is ever decoded to \`/\`, a decoded name can contain at most one separator, so the worst case is two \`..\` segments (\`../..\`), which lands at or above the store root — and that's exactly what the guard blocks. Walked a couple of adjacent variants (single \`..\` segment landing one level up, etc.) and the same containment check catches them all.
- The new regression test (\`refuses a store entry whose decoded name escapes its package directory\`) is genuinely hostile, not happy-path: it creates the \`..+..@1.0.0\` store dir plus a decoy \`package.json\` with \`name: 'pwned'\` at the traversal target, and asserts both \`status === 'ok'\` and that \`'pwned'\` never appears in the output. That's the right assertion shape — it checks the guard actually prevents disclosure, not just that it doesn't throw.
- No new traversal or info-disclosure vector from the guard itself — it silently \`continue\`s on rejection, no path/error data leaks into the report.

No other findings in the diff (no injection, secrets, or dependency changes).

rtorcato (commented):
Re-review of the path-traversal fix in f3ab578 (pnpmStoreDirToName / undecidedTransitiveBuilds).

Traced the guard by hand against the PoC and the legitimate decode cases:

- Malicious entry `..+..@1.0.0` decodes to name `../..`. `root` is built from the literal (single-segment) `entry`, so it never itself contains `..`. `target = path.join(root, name, 'package.json')` normalizes to `<store>/package.json`, which does not start with `root + path.sep` — correctly `continue`s before the read. Traced the same for a double-`+` entry (`..+..+..@1.0.0`); since `.replace('+', '/')` only substitutes the first `+`, the second segment stays a literal `..+..` (not `..`), so at most one traversal level is possible, and the prefix check still rejects it.
- Legitimate scoped entry `@prisma+client@5.22.0` → `@prisma/client`: target lands under `root + sep` as expected, passes the guard.
- Peer-suffixed entry `@babel+core@7.0.0_supports-color@8.0.0` → `@babel/core` (first `@` wins over the one in the peer suffix): decodes and passes the guard correctly.
- The `root + path.sep` check (not a bare `startsWith(root)`) avoids the classic sibling-directory prefix bug (e.g. `node_modules_evil` matching a `node_modules` prefix).

Constraints from #373 all still hold in this version: one `package.json` read per unique name (Set-based `seen`), no recursive walk, skip when `.pnpm` is absent, dedup across versions, direct/transitive graded apart (`drift` vs `optional-missing`), and the existing early-return when `node_modules` is absent is untouched.

No bugs found. The new guard is correct and doesn't regress the legitimate decode cases.

PR #376, opened 2026-08-10
feat(doctor): report and fix a missing packageManager pin

### LINKED ISSUE #372: packageManager is only written by fix github-actions, so most repos never get it
## Summary

#364 added `ensurePackageManager` so the generated workflow has a pnpm version to resolve. It is wired into exactly one place — the `github-actions` fixer (`src/languages/js/fixers.ts:375`) — and there is no doctor check for it.

```
$ grep -rn ensurePackageManager src/ | grep -v generators/misc.ts
src/languages/js/fixers.ts:24:  ensurePackageManager,
src/languages/js/fixers.ts:375:    if ((await ensurePackageManager(targetDir)) === 'added') {
```

## Why it bites

`fix github-actions` only rewrites `ci.yml` when doctor reports the workflow itself as drifted (the `overwrite` guard from #349). So a repo whose `ci.yml` already matches the preset never runs that fixer, and never gets `packageManager` — even though its workflow uses `pnpm/action-setup` with no `version:` input and will fail the moment it runs on a machine where pnpm isn't already present.

The fix is reachable only by a repo that was already broken in a *different* way. Nothing in `doctor` reports the missing field, so there is no signal at all.

## Suggested fix

Two halves, either or both:

1. **A doctor check** — `packageManager` missing on a pnpm repo (detected the way `checkPnpmWorkspace` does it: a `pnpm-workspace.yaml`, a `pnpm-lock.yaml`, or a `packageManager` already starting with `pnpm`). Report `drift`, hint at the fix target.
2. **Wire it into `fix engines`** — that target already writes `engines.node`, and this is the same concern: declaring the toolchain the repo expects in `package.json`. `ensureEnginesNode` and `ensurePackageManager` are already neighbours in `src/cli/generators/misc.ts` and share the `EnsureEnginesResult` return type. If you take this route, widen the target's description and `outputs`.

Note `ensurePackageManager` already detects the version from the running pnpm and falls back to `PNPM_FALLBACK_VERSION` — don't hardcode a version at the call site.

## Verification

- A doctor test: a pnpm repo without `packageManager` reports drift; one with it reports ok; a non-pnpm repo reports ok (not applicable).
- A fix test asserting the field lands via whichever target you wire it to, and that an existing value is left alone.

rtorcato:
## Approach

Both halves of the issue — they compose: the check names the problem, `fix engines` is the target its hint points at.

1. **Doctor check** `packageManager` in `src/languages/js/checks.ts`. pnpm repo with no pin → `drift`; pinned → `ok`; non-pnpm repo → `ok` ("not a pnpm repo"). Mapped to the `engines` fix target.
2. **`fix engines`** now calls `ensurePackageManager` alongside `ensureEnginesNode` — same concern, and they were already neighbours in `generators/misc.ts`. Description, `appliesTo` and `outputs` widened.

Two notes:

- The call site is guarded on pnpm detection. `ensurePackageManager` writes a `pnpm@` range, and `fix engines` is a general target, so an npm or yarn repo must not get one. The guard is at the call site, not inside the generator — the `github-actions` fixer wants it unconditionally, since the workflow it emits is pnpm-based even in a fresh repo with no lockfile yet.
- The version still comes from the running pnpm, falling back to `PNPM_FALLBACK_VERSION`. Nothing hardcoded at the call site.

`checkPnpmWorkspace` had the pnpm-detection triple inline; extracted as `usesPnpm` so both checks agree on the definition.

## Verification

`pnpm run check`, `tsc --noEmit --project src/cli/tsconfig.json`, `vitest run` — 786 passed, 18 skipped.

New tests: three doctor cases (drift / ok / not applicable) in `tests/cli/commands/doctor.test.ts`, three fix cases (pins on pnpm, leaves an existing value alone, skips a non-pnpm repo) in `tests/cli/commands/fix.test.ts`.

Closes #372

### DIFF
--- a/src/cli/commands/doctor.ts
+++ b/src/cli/commands/doctor.ts
@@ -48,6 +48,7 @@ import {
 	checkNodeVersionConsistency,
 	checkNodeVersionPin,
 	checkPackageJson,
+	checkPackageManager,
 	checkPublint,
 	checkSemanticRelease,
 	checkSizeLimit,
@@ -368,6 +369,7 @@ export async function runDoctor(dir: string): Promise<CheckResult[]> {
 	results.push(evaluateNodeVersion(process.version))
 	results.push(checkPackageJson(pkg))
 	results.push(checkEnginesNode(pkg))
+	results.push(await checkPackageManager(targetDir, pkg))
 	results.push(await checkConfigSchemaVersions(targetDir, pkg))
 	results.push(checkGitDependencies(pkg))
 	results.push(await checkVscodeExtensions(targetDir))
--- a/src/cli/commands/fix-targets.ts
+++ b/src/cli/commands/fix-targets.ts
@@ -4,6 +4,7 @@ import type { ProjectConfig } from './setup.js'
 export const FIX_TARGETS: Record<string, string> = {
 	'package.json': 'package-json',
 	'engines.node': 'engines',
+	packageManager: 'engines',
 	EditorConfig: 'editorconfig',
 	'VS Code extensions': 'vscode-extensions',
 	'Node version pin': 'nvmrc',
--- a/src/languages/js/checks.ts
+++ b/src/languages/js/checks.ts
@@ -177,6 +177,42 @@ export function checkPackageJson(pkg: Pkg | null): CheckResult {
 	}
 }
 
+/**
+ * Whether the repo installs with pnpm. Three independent signals, any of which
+ * is enough: the workspace file, the lockfile, or a `packageManager` that
+ * already names pnpm. Shared so `packageManager` and `pnpm settings` agree on
+ * what "a pnpm repo" means.
+ */
+export async function usesPnpm(dir: string, pkg: Pkg | null): Promise<boolean> {
+	return (
+		(await fs.pathExists(path.join(dir, WORKSPACE_FILE))) ||
+		(await fs.pathExists(path.join(dir, 'pnpm-lock.yaml'))) ||
+		((pkg?.packageManager as string | undefined) ?? '').startsWith('pnpm')
+	)
+}
+
+/**
+ * `packageManager` pins the pnpm that `pnpm/action-setup` resolves (#364). The
+ * generated workflow passes no `version:` input, so a pnpm repo without this
+ * field fails CI at setup — before a single check runs (#372).
+ */
+export async function checkPackageManager(dir: string, pkg: Pkg | null): Promise<CheckResult> {
+	const check = 'packageManager'
+	if (!(await usesPnpm(dir, pkg))) {
+		return { check, status: 'ok', detail: 'not a pnpm repo' }
+	}
+	const declared = (pkg?.packageManager as string | undefined) ?? ''
+	if (declared.trim() !== '') {
+		return { check, status: 'ok', detail: `packageManager = ${declared}` }
+	}
+	return {
+		check,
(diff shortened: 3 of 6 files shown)


rtorcato (commented):
**Security review**

No criticals/highs. This diff is wiring, not the sensitive logic:

- `checkPackageManager` (checks.ts) is read-only — it inspects `pkg.packageManager` and reports drift/ok, no writes, no shell-out.
- The write path (`ensurePackageManager`, which shells out to detect the installed pnpm version and falls back to `PNPM_FALLBACK_VERSION`) already existed from #364 and isn't touched by this diff — only two new call sites are added (the doctor check, and the `engines` fix target). I can't re-verify that function's subprocess/sanitization behavior from this diff alone since its body isn't part of the change, but nothing here bypasses it or adds new string interpolation into a shell command.
- The new `fix engines` call site correctly guards on `usesPnpm(targetDir, pkg)` before calling `ensurePackageManager`, so npm/yarn repos won't get a bogus `pnpm@` pin — matches the concern called out in #372.
- An existing non-empty `packageManager` value is left alone (`declared.trim() !== ''` → `ok`), so the fixer won't clobber a maintainer-set pin.
- No secrets, no new dependencies, no SQL/command construction introduced.

One non-blocking note: since the actual version-detection/write logic lives outside this diff, a follow-up worth confirming at some point is that `ensurePackageManager` validates/sanitizes the pnpm-reported version string before it's written into `packageManager` (corepack will fetch whatever that field names). Not something this PR introduces or needs to fix.

rtorcato (commented):
Reviewed against issue #372. Diff is scoped, tested, and both flagged design points check out.

**1. Guard split (call site vs generator)** — correct, and at the right layer. `fix engines` is a general-purpose target that can run on any repo, so guarding `ensurePackageManager` on `usesPnpm(targetDir, pkg)` at the call site (fixers.ts) is right — the generator itself stays reusable/unopinionated. The `github-actions` fixer's unguarded call is pre-existing code from #364 (not touched by this diff) and is consistent: that fixer's emitted `ci.yml` is always pnpm-based regardless of the target repo's actual package manager, so it always needs the pin. No inconsistency introduced.

**2. `usesPnpm` extraction** — verified against the pre-change `checkPnpmWorkspace` body: same three checks (workspace file, `pnpm-lock.yaml`, `packageManager` starting with `pnpm`), same OR/short-circuit order, and the `exists`/`file` locals `checkPnpmWorkspace` still needs downstream (for reading the workspace YAML) are untouched by the extraction. Behavior is preserved exactly.

Other things checked: `ensurePackageManager`'s version still comes from `detectPnpmVersion()` / `PNPM_FALLBACK_VERSION` (no hardcoded/org-specific value baked into emitted files), `FIX_TARGETS['packageManager'] = 'engines'` and the doctor hint text line up with the actual fix target name, and the `'added'`/`'already-set'` sentinel used to compute `filesWritten` in the `engines` fixer is a legitimate shared `EnsureEnginesResult` value, not an invented string.

Nit (non-blocking): `checkPackageManager` treats any non-empty `packageManager` as `ok` once `usesPnpm` is true — a repo with a `pnpm-lock.yaml` but `packageManager: "yarn@..."` would report `ok`. This mirrors the pre-existing `checkPnpmWorkspace` detection semantics, so not a regression, just a pre-existing edge case outside this issue's scope.

Verdict: no blocking defects.

PR #379, opened 2026-08-10
fix(doctor): accept an inlined biome preset, not just an extends pointer

### LINKED ISSUE #378: copy biome leaves the repo in permanent doctor drift
## Summary

`copy biome` writes the full preset inline (`tooling/biome/biome.json` → `biome.json`). That file contains no `extends`, so it carries no `@rtorcato/repo-tooling/biome` string.

Doctor's `Biome` check matches on exactly that string (`src/languages/js/checks.ts`):

```ts
check: 'Biome',
candidates: ['biome.json', 'biome.jsonc'],
expected: `extends "${PACKAGE}/biome"`,
matcher: /@rtorcato\/(?:js|repo)-tooling\/biome/,
```

So a repo that runs the documented `copy biome` command is immediately reported as drifted, and **nothing it does with `copy` can ever clear it** — re-running writes the same non-matching file.

## Reproduction

```bash
npx @rtorcato/repo-tooling copy biome
npx @rtorcato/repo-tooling doctor      # Biome: drift
npx @rtorcato/repo-tooling copy biome  # same file again; still drift
```

`copy biome` is advertised in the README (`copy biome  # → biome.json`) and in the `list` output, so this is a supported path, not a misuse.

## Background

#365 (merged as PR #366) made `fix biome` write a thin `extends` pointer instead of copying the preset, which fixed the `fix` path. The doctor hint was repointed from `copy biome` to `fix biome` at the same time — but `copy biome` itself was left as-is, so the underlying inconsistency is still there for anyone who uses it.

## Suggested fix

**Preferred:** teach the `Biome` check to accept an inlined preset as well as a pointer. A repo that has deliberately vendored the config is not drifted — it has just chosen to own the file, which is exactly what `copy` is for. Match either the `extends` string **or** a preset-shaped config (e.g. a `linter.rules.preset` key plus the biomejs `$schema`), and word `expected` to cover both.

**Alternative:** make `copy biome` emit the same thin pointer `fix biome` does. Simpler, but it changes what `copy` means — every other preset in `PRESETS` copies a real file, and a pointer is not "the config file" the command promises.

Whichever you pick, say so in the PR body. Do **not** change both paths.

## Verification

- A doctor test: a `biome.json` that is a verbatim copy of the shipped preset reports `ok`, not `drift`.
- The existing test that a thin `extends` pointer reports `ok` must keep passing.
- A genuinely unrelated `biome.json` (e.g. `{"linter":{"enabled":false}}`) must still report drift — the check has to stay capable of failing.

rtorcato:
## Approach

Took the **preferred** option from the issue: the `Biome` doctor check now
accepts an inlined preset as well as an `extends` pointer. `copy biome` is
untouched — it still copies the real preset file, like every other entry in
`PRESETS`. The alternative (making `copy biome` emit a thin pointer) was not
implemented; only one path changed.

## What changed

`src/languages/js/checks.ts` — the `Biome` check's matcher is now
`BIOME_CONFIG`, which matches either:

- `@rtorcato/(js|repo)-tooling/biome` — the pointer `fix biome` writes; or
- the biomejs `$schema` **and** a `"preset":` key — the shape `copy biome`
  drops inline.

The second alternative uses two lookaheads rather than a sequence, so a
reformatted or key-reordered config still matches. `expected` now reads
`extends "@rtorcato/repo-tooling/biome" or inlines the preset`, and the stale
comment explaining why the hint says `fix biome` instead of `copy biome` is
gone — both are valid now.

## The check can still fail

Three cases, all covered by tests in `tests/cli/commands/doctor.test.ts`:

| `biome.json` | status |
|---|---|
| verbatim `copy biome` output (via `copyPreset('biome', dir)`) | `ok` (new test) |
| thin `extends` pointer | `ok` (existing test, still passing) |
| `{"linter":{"enabled":false}}` | `drift` (new test) |

The first test copies the real shipped preset rather than a fixture, so it
keeps tracking `tooling/biome/biome.json` if that file changes.

## Verification

`pnpm run check`, `tsc --noEmit --project src/cli/tsconfig.json`, and
`vitest run` (799 passed, 18 skipped) all pass.

## Noted, not fixed

`copy tsconfig` writes `tooling/typescript/tsconfig.base.json`, which likewise
carries no `@rtorcato/.../typescript/` string, so the `TypeScript` check has
the same shape of problem. Left alone — out of scope for this issue.

Closes #378

### DIFF
--- a/src/base/checks.ts
+++ b/src/base/checks.ts
@@ -40,9 +40,18 @@ export interface FileCheck {
 	candidates: string[]
 	/** Phrased to read after the filename: `.swiftlint.yml ${expected}`. */
 	expected: string
-	matcher: RegExp
+	/**
+	 * A regex over the raw text, or — where presence of a marker isn't enough to
+	 * conclude the config is still ours — a predicate that parses it (#378).
+	 */
+	matcher: RegExp | ((contents: string) => boolean)
 	optional?: boolean
 	hint?: string
+}
+
+/** Applies a `FileCheck` matcher of either form to a file's contents. */
+export function matches(matcher: FileCheck['matcher'], contents: string): boolean {
+	return typeof matcher === 'function' ? matcher(contents) : matcher.test(contents)
 }
 
 export async function checkFile(dir: string, spec: FileCheck): Promise<CheckResult> {
@@ -51,7 +60,7 @@ export async function checkFile(dir: string, spec: FileCheck): Promise<CheckResu
 		if (!(await fs.pathExists(filepath))) continue
 
 		const contents = await fs.readFile(filepath, 'utf-8')
-		if (spec.matcher.test(contents)) {
+		if (matches(spec.matcher, contents)) {
 			return {
 				check: spec.check,
 				status: 'ok',
--- a/src/languages/js/checks.ts
+++ b/src/languages/js/checks.ts
@@ -58,17 +58,58 @@ export function evaluateNodeVersion(version: string): CheckResult {
 }
 
 /**
+ * JSONC → object, or null when it genuinely won't parse. `biome.jsonc` is a
+ * declared candidate and the format allows comments and trailing commas, so
+ * bare `JSON.parse` would reject configs Biome itself accepts.
+ *
+ * The first alternative consumes whole string literals, so a `//` or `/*`
+ * inside one (a `$schema` URL, most obviously) is never mistaken for a comment.
+ */
+function parseJsonc(text: string): Record<string, any> | null {
+	const withoutComments = text.replace(
+		/("(?:\\.|[^"\\])*")|\/\/[^\n]*|\/\*[\s\S]*?\*\//g,
+		(_, str: string | undefined) => str ?? ''
+	)
+	try {
+		return JSON.parse(withoutComments.replace(/,(\s*[}\]])/g, '$1')) as Record<string, any>
+	} catch {
+		return null
+	}
+}
+
+const BIOME_PRESET_REF = /@rtorcato\/(?:js|repo)-tooling\/biome/
+
+/**
  * Both shapes of a `biome.json` this package produces (#378): the thin
  * `extends` pointer `fix biome` writes, and the whole preset inlined, which is
  * what `copy biome` drops — it carries no `extends` at all, so matching only
  * the first left every `copy biome` repo permanently drifted with no way out.
  *
- * The inline form is recognised by the biomejs `$schema` plus a `preset` key,
- * via lookaheads so key order doesn't matter. A config that is merely valid
- * JSON (`{"linter":{"enabled":false}}`) has neither and still reports drift.
+ * Read structurally, not as text. Accepting the inline form on the presence of
+ * a `$schema` URL and a `preset` key alone would pass a config that keeps both
+ * markers and turns the linter off — the very drift this check exists to catch
+ * — and the markers could even sit in a `biome.jsonc` comment. So the config is
+ * parsed, `linter.enabled: false` disqualifies either shape, and anything
+ * unparseable counts as drift rather than being waved through.
  */
-const BIOME_CONFIG =
-	/@rtorcato\/(?:js|repo)-tooling\/biome|^(?=[\s\S]*biomejs\.dev\/schemas\/)(?=[\s\S]*"preset"\s*:)/
+function matchesBiomeConfig(contents: string): boolean {
+	const config = parseJsonc(contents)
+	if (!config) return false
+
+	const linter = config.linter as { enabled?: unknown; rules?: Record<string, unknown> } | undefined
+	if (linter?.enabled === false) return false
+
+	if (BIOME_PRESET_REF.test(JSON.stringify(config.extends ?? ''))) return true
+
+	// The inlined preset: biome's own `$schema` plus the `linter.rules.preset`
+	// key the shipped `tooling/biome/biome.json` sets.
+	const schema = config.$schema
+	return (
+		typeof schema === 'string' &&
+		schema.includes('biomejs.dev/schemas/') &&
+		linter?.rules?.preset !== undefined
+	)
+}
 
 export const FILE_CHECKS: FileCheck[] = [
 	{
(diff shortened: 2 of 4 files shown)


rtorcato (commented):

Verified the three required behaviours directly by testing the `BIOME_CONFIG` regex in isolation:

- `extends: ["@rtorcato/repo-tooling/biome"]` → matches (first alternative).
- `{"$schema": ".../biomejs.dev/schemas/...", "preset": "...", ...}` (with keys reordered) → matches (second alternative, both lookaheads satisfied regardless of key order).
- `{"linter":{"enabled":false}}` → does not match, correctly stays `drift`.

The fix does not weaken the check to accept arbitrary JSON — it still requires *both* the biomejs `$schema` URL and a `"preset"` key to be present before treating a file as an inlined preset, so an unrelated config still fails as intended. The three test cases the issue asked for are present in `tests/cli/commands/doctor.test.ts`: verbatim `copy biome` output (via the real `copyPreset('biome', dir)`, not a fixture, so it tracks the shipped preset), the existing thin-pointer case, and a new unrelated-config case — all asserting the correct `ok`/`drift` status.

Only one of the issue's two suggested approaches was implemented (relaxing the doctor matcher); `copy biome` itself is untouched, and the PR body states this explicitly, matching the issue's "do not change both paths" requirement.

Diff is scoped to the one check plus its tests, comment/`expected` string updated to match the new accepted shapes, and the stale comment explaining the old `fix`/`copy` asymmetry was correctly removed since both are now valid. No correctness, security, or convention issues found.

rtorcato (commented):

**Summary:** 1 medium (drift-detection bypass), 0 high/critical, no secrets/injection/dependency issues.

## Medium: the relaxed matcher can be satisfied by a config with the linter disabled

`BIOME_CONFIG` (`src/languages/js/checks.ts`) now accepts either the `extends` pointer *or* raw text containing both `biomejs.dev/schemas/` and `"preset":` anywhere in the file. Both checks are plain substring lookaheads over the raw file text — they don't parse JSON structure or verify no overriding keys are present.

Concretely, this file would report `Biome: ok`:

```json
{
  "$schema": "https://biomejs.dev/schemas/1.9.4/schema.json",
  "preset": "whatever",
  "linter": { "enabled": false }
}
```

This is exactly the scenario the check exists to catch: a copied preset that's since been edited to disable linting. Someone who runs `copy biome` and later flips `linter.enabled` to `false` (or a repo that has silently drifted that way) keeps `$schema`/`preset` intact and the doctor check goes right on reporting `ok` — a disabled linter is reported healthy. Because `biome.jsonc` is also a candidate and allows comments, the same two substrings could even appear inside a comment rather than live config and still satisfy the matcher.

The new tests don't cover this gap: the "drift" test case (`{"linter":{"enabled":false}}`) has neither `$schema` nor `"preset"`, so it doesn't exercise the realistic case of a modified-but-marker-intact preset.

**Fix:** parse the file as JSON and check structurally — e.g., confirm `linter?.enabled !== false` (or the field is absent) in addition to the pointer/preset marker check — rather than relying on raw-text substring presence for a security/drift-relevant field.

**Not an issue:** the regex itself isn't vulnerable to catastrophic backtracking — the `[\s\S]*` lookaheads are simple linear scans with no nested/ambiguous quantifiers, and inputs are local config files, not untrusted network input. No secrets, no shell/SQL construction, no new dependencies in this diff.

rtorcato (commented):

**Verdict: the medium finding from the prior round is resolved.** No blocking security issues — 0 critical, 0 high, 0 medium, 2 info/nit.

## Re-verification of the prior bypass

The fix (`src/languages/js/checks.ts`, `matchesBiomeConfig` + `parseJsonc`) genuinely parses the config rather than doing a better substring heuristic:

- **Structural, not textual.** `linter?.enabled === false` is read from the parsed object and disqualifies *before* the `extends`/`$schema`+`preset` check runs — this isn't special-cased to the reported bypass, it's a general precondition. Confirmed against the new test `reports drift for a preset-shaped biome.json with the linter disabled`, which reproduces the exact prior bypass shape (`$schema` + `linter.rules.preset` present, `enabled: false`) and now correctly reports `drift`.
- **JSONC still parses.** `parseJsonc` strips `//` and `/* */` comments via a regex that matches string literals *first* in the alternation, so a `$schema` URL's `//` or a comment token embedded in a string is never mistaken for a comment start. Trailing commas before `}`/`]` are stripped separately. The added test with a `// shared preset` leading comment confirms this still resolves to `ok`.
- **Fails safe on corrupt/hostile input.** `parseJsonc` wraps `JSON.parse` in try/catch and returns `null` on any failure (including a `RangeError` from pathologically deep nesting, which Node's native JSON.parse throws as a catchable exception rather than crashing the process); `matchesBiomeConfig` treats `null` as non-matching → `drift`, not a thrown error out of `doctor`. Confirmed by the new "too corrupt to parse" test. No new dependency was introduced (no JSONC parser package pulled in — just a small in-house regex + native `JSON.parse`), so there's no new supply-chain surface to scrutinize.
- Property access after parsing (`config.linter`, `config.extends`, `linter?.rules?.preset`) is all optional-chained / nullish-coalesced, so a config where those fields are unexpected types (string, null, array) degrades to non-matching rather than throwing.

I don't see a way to reconstruct the original substring-matching bypass against this version.

## Info / nit (non-blocking)

1. `linter?.rules?.preset !== undefined` treats an explicit `preset: null` as "present" — cosmetic edge case, not exploitable (doesn't affect the `enabled: false` disqualifier, which is checked independently).
2. The `$schema` check is still `.includes('biomejs.dev/schemas/')`, i.e. a substring check on one field — fine here since it's just an identifying marker and the security-relevant gate (`linter.enabled`) doesn't depend on it.

No injection, secrets, shell/SQL construction, or dependency changes in this diff.

rtorcato (commented):

Re-reviewed commit `49c4e95` against the security round's finding. The relaxed matcher now parses (JSONC-aware: string-literal-safe comment stripping + trailing-comma removal) instead of doing raw substring/lookahead matching, and explicitly checks `linter.enabled === false` before accepting either the `extends` pointer or the inlined-preset shape — this directly closes the hole where a preset-shaped config with the linter turned off was previously waved through.

All five required scenarios are covered in `tests/cli/commands/doctor.test.ts`:
- verbatim shipped preset via the real `copyPreset('biome', dir)` → `ok`
- thin `extends` pointer → `ok` (existing test, now also exercises a `biome.jsonc` with a genuine `//` comment, confirming JSONC is still accepted)
- `{"linter":{"enabled":false}}` → `drift`
- preset-shaped config (`$schema` + `linter.rules.preset`) with `linter.enabled: false` → `drift` (the exact bypass the security review flagged)
- unparseable config → `parseJsonc` catches and returns `null`, `matchesBiomeConfig` returns `false`, falls through to `drift` — no crash

`copy biome` / `copy-preset.ts` is untouched. The `matches()` helper cleanly widens `FileCheck['matcher']` to `RegExp | predicate` and is applied consistently in both `base/checks.ts` and `languages/python/checks.ts`, so the type change doesn't leave a caller behind. Diff stays scoped to the fix and tests.

No defects found.

PR #380, opened 2026-08-10
fix(js): add the scripts that run knip, vitest and tsc

### LINKED ISSUE #377: fix knip, vitest and tsconfig still scaffold configs with no script to run them
## Summary

#371 (merged as #374) fixed this for `biome`, `eslint` and `prettier`. The same gap remains on three targets whose scripts the **generated CI actually calls**:

| target | scaffolds | script the CI calls | added by `fix`? |
|---|---|---|---|
| `knip` | `knip.json` | `pnpm knip` | no |
| `vitest` | `vitest.config.ts` | `pnpm coverage` (or `pnpm test`) | no |
| `tsconfig` | `tsconfig.json` | `pnpm typecheck` | no |

None of them calls `ensureScripts`, and none lists `package.json (scripts)` in `outputs`.

`getScripts()` in `src/cli/generators/package-json.ts` already writes all of these on the `setup` path, so a scaffolded project has them and a `fix`-ed one doesn't — the same divergence #371 closed.

## Why it bites

Since #364, `githubJobs` emits a CI step **only when the script exists**. So the failure is silent, not loud:

- No `knip` script → the "Check for unused files, exports, and dependencies" step disappears. If `check`/`lint` is also absent, the entire `lint` job is dropped.
- No `typecheck` script → the `typecheck` job disappears.
- No `coverage`/`test` script → the `test` job disappears, taking the Codecov upload with it.

A repo can end up with a green pipeline that type-checks nothing and tests nothing. Before #364 this failed loudly with `ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL`; now it just quietly does less.

`fix verify` is affected too — `composeVerifyScriptFromPkg` needs `scripts.typecheck` (or the `typescript` dep) and a test signal to compose a chain at all.

## Suggested fix

Follow the merged precedent in the `biome` / `eslint` / `prettier` targets exactly — `ensureScripts` is already exported from `src/cli/generators/package-json.ts` and only fills in names that are missing, so a repo's own commands are never overwritten:

- `knip` → `{ knip: 'knip' }`
- `tsconfig` → `{ typecheck: 'tsc --noEmit' }`
- `vitest` → `{ test: 'vitest', 'test:watch': 'vitest --watch', coverage: 'vitest run --coverage' }`

Keep these in step with `getScripts()` — that's the source they should match. Update each target's `outputs` to mention `package.json (scripts)` and set `riskLevel: 'safe-merge'`.

Note `tsconfig` is shared with non-JS paths — only add the script when there's a `package.json` to add it to (`ensureScripts` already returns false when there isn't).

## Verification

Mirror the `fix biome scripts` test in `tests/cli/commands/fix.test.ts` for each target: the script lands, and an existing one of the same name is left untouched.

rtorcato:
`fix knip`, `fix vitest` and `fix tsconfig` wrote a config and nothing else, so a repo got a tool it could not invoke. Since #364 the generated CI only emits a step when the script exists, so the failure is silent rather than loud:

- no `knip` script → the unused-code step disappears, and with no `check`/`lint` either the whole `lint` job goes with it
- no `typecheck` script → the `typecheck` job disappears
- no `coverage`/`test` script → the `test` job disappears, Codecov upload included

A repo could end up with a green pipeline that type-checks nothing and tests nothing.

All three targets now call `ensureScripts`, which only fills in names that are missing, so a repo's own commands survive. The commands match `getScripts()` in `src/cli/generators/package-json.ts`.

Two deliberate deviations from the issue text:

- `test:ui` is left out of the vitest set. `getScripts()` emits it, but only the browser/both environments install `@vitest/ui`, so a fixed-up repo would get a script it cannot run.
- `riskLevel` is left alone, as #374 did for biome/eslint/prettier. The issue suggested `safe-merge`, but these fixers still overwrite the config file itself, so that prompt wording ("existing fields preserved") would be wrong and would default a drift overwrite to Yes.

`tsconfig` is shared with non-JS paths. `ensureScripts` returns false when there is no `package.json`, so the target still reports only the `tsconfig.json` it wrote — covered by a test.

## Verification

Four cases in `tests/cli/commands/fix.test.ts`: the `knip`, `typecheck` and vitest scripts land, an existing script of the same name is left untouched, and `fix tsconfig` still writes `tsconfig.json` in a directory with no `package.json`.

`pnpm run check`, `pnpm run typecheck` and `vitest run` (801 passed, 18 skipped) all pass locally.

Closes #377

### DIFF
--- a/src/languages/js/fixers.ts
+++ b/src/languages/js/fixers.ts
@@ -49,6 +49,25 @@ const ESLINT_SCRIPTS: Record<string, string> = {
 const PRETTIER_SCRIPTS: Record<string, string> = {
 	format: 'prettier --write .',
 }
+
+/** Kept in step with the knip line of getScripts() in package-json.ts. */
+const KNIP_SCRIPTS: Record<string, string> = {
+	knip: 'knip',
+}
+
+/** Kept in step with the typescript branch of getScripts() in package-json.ts. */
+const TSCONFIG_SCRIPTS: Record<string, string> = {
+	typecheck: 'tsc --noEmit',
+}
+
+// getScripts() also emits `test:ui`, but only the browser/both environments
+// install @vitest/ui, so a fixed-up repo would get a script it can't run.
+/** Kept in step with the vitest branch of getScripts() in package-json.ts. */
+const VITEST_SCRIPTS: Record<string, string> = {
+	test: 'vitest',
+	'test:watch': 'vitest --watch',
+	coverage: 'vitest run --coverage',
+}
 import {
 	WORKSPACE_FILE,
 	dependsOnEsbuild,
@@ -181,13 +200,19 @@ export const FIXERS: Fixer[] = [
 	},
 	{
 		target: 'tsconfig',
-		description: 'Scaffold tsconfig.json extending the @rtorcato/repo-tooling preset',
+		description:
+			'Scaffold tsconfig.json extending the @rtorcato/repo-tooling preset, plus the `typecheck` script',
 		appliesTo: ['TypeScript'],
-		outputs: ['tsconfig.json'],
+		outputs: ['tsconfig.json', 'package.json (scripts)'],
 		canFixDrift: true,
 		async run({ targetDir }) {
 			const result = await copyPreset('tsconfig', targetDir)
-			return { filesWritten: [result.target] }
+			const filesWritten = [result.target]
+			// Without `typecheck` the generated CI drops its typecheck job entirely
+			// (#377). Shared with non-JS paths, so this is a no-op when there's no
+			// package.json to add it to.
+			if (await ensureScripts(targetDir, TSCONFIG_SCRIPTS)) filesWritten.push('package.json')
+			return { filesWritten }
 		},
 	},
 	{
@@ -221,9 +246,10 @@ export const FIXERS: Fixer[] = [
 	},
 	{
 		target: 'vitest',
-		description: 'Scaffold vitest.config.ts (preserves vitest.setup.ts if present)',
+		description:
+			'Scaffold vitest.config.ts (preserves vitest.setup.ts if present), plus the scripts that run it',
 		appliesTo: ['Vitest'],
-		outputs: ['vitest.config.ts'],
+		outputs: ['vitest.config.ts', 'package.json (scripts)'],
 		canFixDrift: true,
 		async run({ targetDir, pkg }) {
 			const setupPath = path.join(targetDir, 'vitest.setup.ts')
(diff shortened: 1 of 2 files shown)


rtorcato (commented):

No security defects found.

- **Injection**: The three new script maps (`KNIP_SCRIPTS`, `TSCONFIG_SCRIPTS`, `VITEST_SCRIPTS`) are static object literals (`knip`, `tsc --noEmit`, `vitest`, `vitest --watch`, `vitest run --coverage`) — no user- or repo-derived input is interpolated into these command strings, so there's no injection surface.
- **Overwrite risk**: The specific concern flagged (whether an existing user script in a consumer's `package.json` could be silently clobbered) is addressed — `ensureScripts` only fills in missing keys, matching the merged precedent from #374 for biome/eslint/prettier. The new tests in `tests/cli/commands/fix.test.ts` explicitly verify this: seeding `scripts.knip: 'my own knip'` (and similarly for `typecheck`, `test`) before rerunning `fix` asserts the user's value survives untouched.
- **Secrets**: None present in the diff.
- **Dependencies/supply chain**: None touched — no new packages, no version bumps.
- **Consistency**: The literal values match `getScripts()` in `src/cli/generators/package-json.ts` (the stated source of truth per the linked issue #377), so scaffolded and fixed repos converge rather than diverge.

One non-security nit for the code reviewer: the issue's suggested fix also calls for setting `riskLevel: 'safe-merge'` on these three fixer entries, but that isn't visible in the diff hunks — worth confirming it was added (or already present) since it's not a security concern either way.

rtorcato (commented):

Reviewed against issue #377 and the diff only.

**Correctness**
- Script values match the issue's stated source of truth exactly: `knip: 'knip'`, `typecheck: 'tsc --noEmit'`, and the three vitest scripts (`test`, `test:watch`, `coverage`). No divergence from `getScripts()` as described.
- `ensureScripts` guard is used correctly on all three targets (`knip`, `tsconfig`, `vitest`) and the accompanying tests explicitly verify an existing same-named script is left untouched (`fix.test.ts:1299+`) — satisfies the "don't overwrite" requirement.
- `tsconfig` correctly guards the no-`package.json` case: `ensureScripts` returns false and `filesWritten` simply omits `'package.json'`, with a dedicated test confirming `tsconfig.json` still gets written when there's no `package.json` in the target dir.
- `outputs` widened to include `'package.json (scripts)'` on all three targets, matching the #374 precedent.

**Convention**
- New `KNIP_SCRIPTS`/`TSCONFIG_SCRIPTS`/`VITEST_SCRIPTS` consts are inserted before the `import { WORKSPACE_FILE, ... }` statement — unusual, but this mirrors the file's existing pre-#380 layout (that import already sat below `ESLINT_SCRIPTS`/`PRETTIER_SCRIPTS`), so it's not a new issue introduced here.
- Couldn't confirm from the diff alone whether `riskLevel: 'safe-merge'` is set on these three fixers (no `riskLevel` lines appear in any of the three hunks) — plausible it was already `'safe-merge'` pre-PR and thus outside the diff context, but flagging since the issue calls it out explicitly as part of the #374 precedent to follow.

**Tests**
- New `describe('fix knip/tsconfig/vitest scripts', ...)` block mirrors the `fix biome scripts` precedent per the issue's verification checklist: script lands, existing script of the same name is preserved, and the no-`package.json` tsconfig case is covered.

No blocking defects found.

PR #383, opened 2026-08-10
fix(cli): route fix tsconfig through generateTSConfig

### LINKED ISSUE #381: fix tsconfig writes a config doctor immediately flags, and discards project-type settings
## Summary

`fix tsconfig` copies the preset inline instead of writing an `extends` pointer:

```ts
target: 'tsconfig',
description: 'Scaffold tsconfig.json extending the @rtorcato/repo-tooling preset',
canFixDrift: true,
async run({ targetDir }) {
  const result = await copyPreset('tsconfig', targetDir)   // tooling/typescript/tsconfig.base.json
  return { filesWritten: [result.target] }
},
```

`tooling/typescript/tsconfig.base.json` contains no `@rtorcato/repo-tooling/typescript/` string, but doctor's `TypeScript` check matches on exactly that:

```ts
check: 'TypeScript',
expected: `extends "${PACKAGE}/typescript/*"`,
matcher: /@rtorcato\/(?:js|repo)-tooling\/typescript\//,
```

So the target's own description ("extending the … preset") is false, and since `canFixDrift: true`, `fix` offers it as the remedy for the very check it then fails. Re-running writes the same non-matching file — **permanent drift nothing can clear.**

This is the same defect #365 fixed for Biome (PR #366), one layer worse: the `TypeScript` check is not `optional`, so it reports at higher severity than Biome's did.

## Second, worse half: it discards project-type settings

`setup` uses `generateTSConfig` (`src/cli/generators/tsconfig.ts`), which writes a thin pointer **plus** settings derived from the project:

- `extends` targets the right preset per project type (`base` / `react` / `next` / `node`), or `bun` when Bun is enabled
- libraries get `compilerOptions.outDir` / `rootDir`
- `web-app` on the `base` preset gets `lib: ['ES2022','DOM','DOM.Iterable']` — without which `document`/`window` don't typecheck
- `nextjs-app` gets `next-env.d.ts` + app/pages/components in `include`, and `.next` excluded

`fix tsconfig` throws all of that away and drops a generic base config. On a Next.js app it silently removes the Next wiring.

## Reproduction

```bash
npx @rtorcato/repo-tooling doctor      # TypeScript: drift
npx @rtorcato/repo-tooling fix tsconfig
npx @rtorcato/repo-tooling doctor      # TypeScript: still drift
```

## Suggested fix

Route the `tsconfig` target through `generateTSConfig` with `inferProjectConfig(pkg)`, exactly as PR #366 routed `fix biome` through `generateBiomeConfig`. `inferProjectConfig` already derives `projectType` and `typescript.config` from the package.json, so the fix path produces the same shape as `setup`.

Preserving a user's existing `compilerOptions` would be a bonus, but is not required — match `setup`'s output first.

Leave `copy tsconfig` alone; #378 covers the `copy`-path question separately.

## Verification

- `fix tsconfig` on a repo with no tsconfig writes one whose `extends` matches doctor's `TypeScript` matcher, and `doctor` reports `ok` immediately after.
- On a repo with `next` in dependencies, the emitted config carries the Next includes and `.next` exclusion.
- On a library, `outDir`/`rootDir` are present.

rtorcato:
`fix tsconfig` copied `tooling/typescript/tsconfig.base.json` inline. That file carries no `@rtorcato/repo-tooling/typescript/` string, which is exactly what doctor's `TypeScript` matcher looks for — so the fix could never clear the check it was offered for, and it discarded every project-type setting `setup` emits.

The target now builds a `ProjectConfig` with `inferProjectConfig(pkg)` and hands it to `generateTSConfig`, the same routing PR #366 used for `fix biome`. `copy tsconfig` is untouched (#378).

## Two gaps in `inferProjectConfig`, both closed here

- **`typescript.config`** — it mapped `nextjs-app` to `next` and everything else to `base`, so a React app landed on the base preset, whose `lib` is ES2022. `document`/`window` would not typecheck. It now maps `react-app` to `react`, matching the react-app setup preset. `generateTSConfig` is the only consumer of that field.
- **`bun`** — it does not derive it, and cannot: Bun leaves no package.json trace. The fixer reads a `bunfig.toml` in the target dir as the signal, so `fix tsconfig` on a Bun repo keeps the Bun-typed preset instead of stripping `types: ['bun']` off it.

`web-app` never comes out of `inferProjectConfig` (nothing in a package.json distinguishes it from a library), so the `web-app`-on-`base` DOM-lib branch stays unreachable from the fix path. It is reachable from `setup`, where the user picks the type.

The doctor hint for `TypeScript` now points at `fix tsconfig` rather than telling the user to hand-edit `extends`, mirroring what #366 did to the Biome hint.

## Verification

Five tests in `tests/cli/commands/fix.test.ts`, covering the issue's list: the emitted `extends` pointer plus `runDoctor` reporting `TypeScript: ok` straight after; library `outDir`/`rootDir`; Next includes and `.next` exclusion on a repo with `next` in dependencies; the react preset on a repo with `react-dom`; the bun preset when a `bunfig.toml` is present.

`pnpm run check`, `pnpm run typecheck`, and `vitest run` (806 passed, 18 skipped) are green.

Closes #381

### DIFF
--- a/src/languages/js/checks.ts
+++ b/src/languages/js/checks.ts
@@ -63,7 +63,9 @@ export const FILE_CHECKS: FileCheck[] = [
 		candidates: ['tsconfig.json'],
 		expected: `extends "${PACKAGE}/typescript/*"`,
 		matcher: /@rtorcato\/(?:js|repo)-tooling\/typescript\//,
-		hint: 'Set `"extends": "@rtorcato/repo-tooling/typescript/base"` in tsconfig.json',
+		// `fix tsconfig`, not `copy tsconfig`: copy drops the whole preset inline,
+		// which carries no `extends` for the matcher above to find (#381).
+		hint: 'Run `npx @rtorcato/repo-tooling fix tsconfig` to scaffold',
 	},
 	{
 		check: 'Biome',
--- a/src/languages/js/fixers.ts
+++ b/src/languages/js/fixers.ts
@@ -77,6 +77,7 @@ import {
 import { generatePostcss } from '../../cli/generators/postcss.js'
 import { generateCypressConfig, generateVitestConfig } from '../../cli/generators/testing.js'
 import { generateTreeshakeCheck, inferSubpathsFromExports } from '../../cli/generators/treeshake.js'
+import { generateTSConfig } from '../../cli/generators/tsconfig.js'
 import { generateTailwind } from '../../cli/generators/tailwind.js'
 import { generateTurborepo } from '../../cli/generators/turborepo.js'
 import { generateNx } from '../../cli/generators/nx.js'
@@ -104,7 +105,13 @@ export function inferProjectConfig(pkg: Pkg): ProjectConfig {
 	return {
 		projectName: (pkg?.name as string) ?? 'project',
 		projectType,
-		typescript: { enabled: true, config: projectType === 'nextjs-app' ? 'next' : 'base' },
+		typescript: {
+			enabled: true,
+			// Same mapping the react-app/nextjs-app setup presets use — a react-app on
+			// the base preset has no DOM libs, so `document`/`window` don't typecheck.
+			config:
+				projectType === 'nextjs-app' ? 'next' : projectType === 'react-app' ? 'react' : 'base',
+		},
 		linting: {
 			tool: 'biome',
 			eslintConfig: projectType === 'nextjs-app' ? 'nextjs' : 'base',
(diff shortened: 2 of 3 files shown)


rtorcato (commented):

**Verdict: clean — no criticals, no highs.**

Reviewed for injection, leaked secrets, unsafe shell/SQL, and supply-chain risk in the switch from `copyPreset('tsconfig', ...)` to `generateTSConfig(inferProjectConfig(pkg), targetDir)`.

**`extends` value is not attacker-controlled pass-through.** The new `typescript.config` field is chosen from a fixed enum (`'next' | 'react' | 'base'`, plus `bun` gated on the *existence* of a `bunfig.toml`, not its contents) — it is not built from an arbitrary string in the consumer's `package.json`. No path/value from untrusted package.json content flows unvalidated into the emitted `extends` string.

**`pkg.name` → `projectName`** is passed through as-is (`pkg?.name ?? 'project'`), but this generator files aren't in this diff, so I can't confirm where `projectName` lands in the emitted output. Given the trust boundary here is a developer running this tool against their own repo's own `package.json` (not a dependency's, not remote input), this isn't an actionable finding — info-level at most.

**Existing hand-tuned `tsconfig.json` is still fully overwritten** — `run()` never reads the current file before calling `generateTSConfig`. This *is* a real loss-of-settings concern (e.g. a hand-set `strict`/`noEmit` could vanish), but it's pre-existing behavior (the prior `copyPreset` also fully overwrote the file) and is explicitly called out and deferred in issue #381 ("Preserving a user's existing compilerOptions would be a bonus, but is not required"). Not a regression introduced by this PR, and arguably an improvement since the emitted config is now project-type-aware instead of a static base copy. Not blocking.

No new dependencies, no secrets, no shell/SQL construction in this diff.

rtorcato (commented):

Reviewed diff + issue #381 only (no wider repo exploration).

Both halves of the issue are delivered:
1. `fix tsconfig` now routes through `inferProjectConfig(pkg)` + `generateTSConfig`, so the emitted `extends` matches doctor's `TypeScript` regex — the doctor hint was also updated to point at `fix tsconfig` instead of a manual edit, mirroring #366's Biome hint change.
2. Project-type settings survive: library `outDir`/`rootDir`, Next `include`/`exclude` + `.next` exclusion, and a new `react-app` → `react` preset mapping (previously `react-app` fell through to `base`, which has no DOM lib). Bun is handled by checking for `bunfig.toml` in `targetDir` since `inferProjectConfig` has no package.json signal for it.

The PR body is upfront about the one real gap: `inferProjectConfig` can't distinguish `web-app` from a library (no package.json marker), so the `web-app`-on-`base` DOM-lib branch stays unreachable from the fix path and only reachable via interactive `setup`. That's disclosed rather than papered over, and it's outside the issue's stated verification checklist (extends+doctor-ok, library outDir/rootDir, Next wiring), so it's not a blocking gap.

`copy tsconfig` is untouched, consistent with #378 owning that path. Five new tests in `fix.test.ts` directly exercise the issue's acceptance criteria, and the PR states `pnpm run check`/`typecheck`/`vitest run` are green.

No bugs or convention violations found.

PR #384, opened 2026-08-10
fix(js): give the size-limit budget a script and a CLI to run it

### LINKED ISSUE #382: fix size-limit scaffolds a budget with no script to run it
## Summary

`fix size-limit` writes `.size-limit.cjs` or `.size-limit.json` and stops there. No `size-limit` script is added, and unlike the other tools, `getScripts()` in `src/cli/generators/package-json.ts` doesn't write one on the `setup` path either — so **no path in the tool produces a way to run it.**

The scaffolded budget is inert: it is never executed locally, never wired into `verify`, and never called by the generated CI. A size budget nothing runs is worse than no budget, because the file's presence implies it is being enforced.

## Suggested fix

Follow the merged precedent from PR #374 / #380 — `ensureScripts` from `src/cli/generators/package-json.ts`, which only fills in names that are missing:

```ts
{ 'size-limit': 'size-limit' }
```

Update the target's `outputs` to mention `package.json (scripts)` and set `riskLevel: 'safe-merge'`.

Two follow-on decisions worth making in the same PR:

1. **Install the dependency.** `size-limit` is not in `getDependencies()`, so the script would fail with "command not found" on a repo that never added it. The `attw` and `publint` targets both add their own devDependency — mirror that.
2. **Add it to `getScripts()` too**, so `setup` and `fix` agree. The divergence between those two paths is the root cause behind #371 and #377; don't reintroduce it here.

Wiring it into the `verify` chain is **out of scope** — a size budget failing a local verify run is a bigger behavioural change than this issue is asking for, and `composeVerifyScriptFromPkg` has its own rules about what earns a slot in the chain.

## Verification

- `fix size-limit` adds the script and the devDependency, and leaves an existing `size-limit` script untouched.
- A test asserting `getScripts()` and the fixer emit the same command, so the two paths can't drift again.

rtorcato:
`fix size-limit` wrote `.size-limit.cjs`/`.size-limit.json` and stopped, and `getScripts()` never emitted a script on the `setup` path either — so no path in the tool produced a way to run the budget. A budget nothing runs is worse than no budget: the file's presence implies an enforcement that isn't there.

## What changed

- **Fixer** (`src/languages/js/fixers.ts`): after scaffolding the budget it installs `size-limit` and fills in a missing `size-limit` script via `ensureScripts` — an existing script or pin is left alone. `outputs` now mentions `package.json (devDependencies + scripts)` and `riskLevel` is `safe-merge`.
- **Setup** (`getScripts()` / `getDependencies()`): libraries get the same script and devDependency, gated exactly like `attw`.
- **Setup, cont.**: `generateConfigs` now writes the budget for libraries, and `computeFileList` lists `.size-limit.json`. Without this the script `getScripts()` emits would have nothing to run — the same half-wired state on the other side of the fence.
- The command string and the version range are defined once (`SIZE_LIMIT_SCRIPTS`, `SIZE_LIMIT_VERSION` in `src/cli/generators/package-json.ts`) and shared by both paths, so they can't drift the way #371 and #377 did.

Wiring `size-limit` into the `verify` chain is deliberately **not** done — out of scope per the issue.

## Tests

- `fix size-limit` adds the script + devDependency alongside the budget.
- An existing `size-limit` script and pin survive a re-run.
- A parity test asserting `getScripts()` and the fixer emit the same command and the same range.
- `computeFileList` and `generateConfigs` agree on the budget, for libraries only.

`pnpm run check`, `tsc --noEmit -p src/cli/tsconfig.json` and `vitest run` (825 passed) are green.

Closes #382

### DIFF
--- a/apps/docs/docs/guides/cli.md
+++ b/apps/docs/docs/guides/cli.md
@@ -204,7 +204,7 @@ Implementation note: the preview is computed by shadow-running the fixer in a te
 | `bun` | `bunfig.toml` + a Bun-typed `tsconfig.json` for Bun runtime/test users |
 | `rollup` | `rollup.config.mjs` re-exporting the shared library preset |
 | `rolldown` | `rolldown.config.mjs` re-exporting the shared library preset |
-| `size-limit` | a size-limit budget (`.size-limit.cjs`/`.json`) |
+| `size-limit` | a size-limit budget (`.size-limit.cjs`/`.json`), plus the `size-limit` script and devDependency that run it |
 | `treeshake-check` | `apps/treeshake-check` — esbuild + metafile bundle assertion |
 | `pnpm-workspace` | merges the family-wide pnpm settings into `pnpm-workspace.yaml` (`verifyDepsBeforeRun`, `minimumReleaseAgeExclude`, esbuild's build approval) — never rewrites the file |
 | `turborepo` | `turbo.json` task pipeline (pnpm-workspace monorepos) |
--- a/src/cli/commands/setup-presets.ts
+++ b/src/cli/commands/setup-presets.ts
@@ -231,6 +231,9 @@ export function computeFileList(config: ProjectConfig): string[] {
 	if (config.typescript.enabled) {
 		files.push('tsconfig.json', 'reset.d.ts')
 	}
+	// Libraries get a bundle-size budget plus the `size-limit` script that runs it
+	// (#382). The scaffold's exports are root-only, so it's always the static form.
+	if (config.projectType === 'library') files.push('.size-limit.json')
 	if (config.linting.tool === 'biome' || config.linting.tool === 'both') {
 		files.push(BIOME_CONFIG)
 	}
--- a/src/cli/generators/index.ts
+++ b/src/cli/generators/index.ts
@@ -9,7 +9,12 @@ import { ensurePnpmSettings, familyGlob } from './pnpm-workspace.js'
 import { generateGitConfigs } from './git.js'
 import { generateGitHubActions } from './github-actions.js'
 import { generateLintingConfigs } from './linting.js'
-import { generateKnipConfig, generateMiscBaseline, generateVscodeExtensions } from './misc.js'
+import {
+	generateKnipConfig,
+	generateMiscBaseline,
+	generateSizeLimitConfig,
+	generateVscodeExtensions,
+} from './misc.js'
 import { generatePackageJson } from './package-json.js'
 import { generateReadme } from './readme.js'
 import { generateSecurityConfigs } from './security.js'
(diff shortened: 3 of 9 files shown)


rtorcato (commented):

**Supply-chain check on the new `size-limit` devDependency — no issues found.**

- Package name is spelled correctly: `size-limit` is the real, correctly-named CLI (npmjs.com/package/size-limit), not a typosquat.
- Version is pinned as a sensible caret range, `^11.2.0`, which is the actual current latest release — not `*`/`latest`/unbounded. Matches the pinning style used for `@arethetypeswrong/cli` (`^0.18.2`) elsewhere in the same file.
- The single source of truth (`SIZE_LIMIT_SCRIPTS`, `SIZE_LIMIT_VERSION` in `src/cli/generators/package-json.ts`) is shared by both the `setup` and `fix` paths, so the two can't diverge in version or command.
- The script value (`{ 'size-limit': 'size-limit' }`) is a static literal, not interpolated from any untrusted package.json content — no injection surface.
- `ensureSizeLimitDependency` in `src/languages/js/fixers.ts` checks both `dependencies` and `devDependencies` for an existing `size-limit` key before writing, and `ensureScripts` only fills in a missing script name — an existing user-declared pin or script survives untouched (covered by a dedicated test, "leaves an existing script and pin alone").
- No secrets, no shell/exec construction, no org-specific values baked into emitted files — the version/command are generic across all consumers.

No blocking defects found.

rtorcato (commented):

**Scope expansion (generateConfigs + computeFileList) — correct, not creep.** Before this PR, `generateSizeLimitConfig` was only ever invoked by the `fix size-limit` fixer; `generators/index.ts` never called it, so a fresh `setup` run on a library never wrote `.size-limit.json`/`.cjs` at all. Adding `size-limit` to `getScripts()` without also teaching `generateConfigs` to scaffold the budget would have shipped a `size-limit` script pointing at a file that setup never creates — the exact defect #382 describes, just relocated to the setup path instead of the fixer path. The gate is applied consistently in all four places touched: `getScripts()`, `getDependencies()`, `generateConfigs`, and `computeFileList` all check `config.projectType === 'library'`, matching the `attw` precedent. This closes the loop the issue's own reasoning demands, even though the issue's "Suggested fix" section didn't spell it out.

**Snapshot update is legitimate.** The diff to `snapshot.test.ts.snap` inserts `"size-limit": "size-limit"` immediately after `"attw": "attw --pack"` in scripts, and `"size-limit": "^11.2.0"` immediately after `"@arethetypeswrong/cli"` in devDependencies — which is exactly the insertion order the code produces (`Object.assign(scripts, SIZE_LIMIT_SCRIPTS)` right after the `attw` assignment; `deps['size-limit'] = ...` right after `deps['@arethetypeswrong/cli']`). This reflects genuinely new output, not a blessed regression.

**Library-only gating is consistent.** `getScripts()`, `getDependencies()`, `generateConfigs`, and `computeFileList` all gate on `projectType === 'library'`. The fixer's own `run()` (`ensureSizeLimitDependency` + `ensureScripts`) applies unconditionally when explicitly invoked, matching how `fix` commands already behave for other targets (fix runs on explicit user intent regardless of project type) — not a new setup/fix divergence.

**Verify chain left alone**, confirmed: `composeVerifyScriptFromPkg` has no changes in this diff, matching the issue's explicit out-of-scope call.

`SIZE_LIMIT_SCRIPTS`/`SIZE_LIMIT_VERSION` are defined once and imported by both the setup and fixer paths, which is the actual fix for the #371/#377 divergence pattern.

No defects found.

PR #388, opened 2026-08-10
fix(doctor): accept an inlined TypeScript preset

### LINKED ISSUE #385: The TypeScript doctor check still rejects an inlined preset, so copy tsconfig drifts forever
## Summary

#378 (merged as PR #379) taught doctor's `Biome` check to accept an inlined preset as well as an `extends` pointer, so a repo that ran `copy biome` is no longer reported as permanently drifted. The `TypeScript` check has the identical problem and was not given the same treatment:

```ts
check: 'TypeScript',
candidates: ['tsconfig.json'],
expected: `extends "${PACKAGE}/typescript/*"`,
matcher: /@rtorcato\/(?:js|repo)-tooling\/typescript\//,
hint: 'Run `npx @rtorcato/repo-tooling fix tsconfig` to scaffold',
```

`tooling/typescript/tsconfig.base.json` contains no `@rtorcato/repo-tooling/typescript/` string — it *is* the preset, it doesn't extend it. So `copy tsconfig` produces a config the check can never match, and re-running `copy` writes the same file again.

## Why it matters more here than it did for Biome

The `Biome` check is `optional: true`. The `TypeScript` check is not, so it reports at a higher severity — a repo that vendored its tsconfig is flagged as a real problem rather than an optional one.

`copy tsconfig` is advertised in the README (`copy tsconfig  # → tsconfig.json`) and in `list` output, so this is a supported path.

## Suggested fix

Apply the #379 treatment to the `TypeScript` check: accept either an `extends` pointing at our preset **or** a config that *is* the preset inlined. #379 parses the config (JSONC-aware, string-safe) rather than substring-matching — reuse that machinery rather than writing a second, weaker version of it.

**The check must stay capable of failing.** #379's lesson was that the tempting fix is to loosen the matcher until everything passes. Prove by test:
- `extends: "@rtorcato/repo-tooling/typescript/base"` → `ok`
- verbatim copy of the shipped preset (via the real `copyPreset('tsconfig')`, not a fixture) → `ok`
- an unrelated `tsconfig.json` such as `{"compilerOptions":{"strict":false}}` → **still drift**
- a config with the preset's shape but `compilerOptions.strict: false` → **drift** (this is the exact bypass the security review caught on #379; do not reintroduce it)
- unparseable / commented JSONC → does not crash doctor

Leave `fix tsconfig` alone — #381 (PR #383) already routed it through `generateTSConfig`, and it emits a proper `extends`.

## Verification

Extend the existing `TypeScript` check tests alongside the `Biome` ones added by #379.

rtorcato:
`copy tsconfig` writes `tooling/typescript/tsconfig.base.json` verbatim. That file *is* the preset, so it carries no `@rtorcato/repo-tooling/typescript/` string for the old text matcher to find, and every repo that ran `copy tsconfig` was reported as drifted with no way out — the check is not optional, so it read as a real problem rather than an optional one.

The config is now parsed the way #379 taught the Biome check to parse (JSONC-aware, string-safe, via the shared `parseJsonc`), and either shape is accepted: an `extends` pointing at the preset, or the preset inlined (a `${configDir}`-anchored `rootDir` plus the preset's strict block).

Structural rather than textual, so the check stays capable of failing:

- `extends: "@rtorcato/repo-tooling/typescript/base"` → ok
- the real `copyPreset('tsconfig')` output → ok
- `{"compilerOptions":{"strict":false}}` → drift
- the copied preset with only `strict` flipped to false → drift (the bypass the #379 security review caught, one preset over)
- `extends` the preset but `strict: false` → drift
- commented JSONC → ok; unparseable → drift, no crash

`fix tsconfig` is untouched: #383 already routes it through `generateTSConfig`, which emits an `extends`.

Closes #385

### DIFF
--- a/src/languages/js/checks.ts
+++ b/src/languages/js/checks.ts
@@ -111,14 +111,60 @@ function matchesBiomeConfig(contents: string): boolean {
 	)
 }
 
+const TS_PRESET_REF = /@rtorcato\/(?:js|repo)-tooling\/typescript\//
+
+/**
+ * Options `tooling/typescript/tsconfig.base.json` sets that a tsconfig written
+ * by hand is unlikely to carry all of. `strict` alone is far too common to
+ * conclude anything from.
+ */
+const TS_PRESET_STRICTNESS = [
+	'strict',
+	'noUncheckedIndexedAccess',
+	'noImplicitOverride',
+	'noPropertyAccessFromIndexSignature',
+] as const
+
+/**
+ * The same two shapes as `matchesBiomeConfig`, for the same reason (#385): the
+ * `extends` pointer `fix tsconfig` writes, and the whole preset inlined, which
+ * is what `copy tsconfig` drops — `tooling/typescript/tsconfig.base.json` *is*
+ * the preset, so it names it nowhere, and every `copy tsconfig` repo was
+ * reported as drifted with no way out. This check isn't optional, so that read
+ * as a real problem rather than an optional one.
+ *
+ * Structural rather than textual, and for the same reason: `strict: false` is
+ * drift whichever shape the config takes, a marker sitting in a comment proves
+ * nothing, and unparseable counts as drift rather than being waved through.
+ */
+function matchesTsConfig(contents: string): boolean {
+	const config = parseJsonc(contents)
+	if (!config) return false
+
+	const compilerOptions = (config.compilerOptions ?? {}) as Record<string, unknown>
+	// Naming the preset and then switching strictness off is precisely the drift
+	// this check exists to catch, so it disqualifies either shape.
+	if (compilerOptions.strict === false) return false
+
+	if (TS_PRESET_REF.test(JSON.stringify(config.extends ?? ''))) return true
+
+	// The inlined preset: a `${configDir}`-anchored rootDir — how a preset meant
+	// to be extended resolves against the consuming repo — plus the strict block
+	// the shipped preset sets.
+	const rootDir = compilerOptions.rootDir
+	return (
+		typeof rootDir === 'string' &&
+		rootDir.includes('${configDir}') &&
+		TS_PRESET_STRICTNESS.every((key) => compilerOptions[key] === true)
+	)
+}
+
 export const FILE_CHECKS: FileCheck[] = [
 	{
 		check: 'TypeScript',
 		candidates: ['tsconfig.json'],
-		expected: `extends "${PACKAGE}/typescript/*"`,
-		matcher: /@rtorcato\/(?:js|repo)-tooling\/typescript\//,
-		// `fix tsconfig`, not `copy tsconfig`: copy drops the whole preset inline,
-		// which carries no `extends` for the matcher above to find (#381).
+		expected: `extends "${PACKAGE}/typescript/*" or inlines the preset, with strict on`,
+		matcher: matchesTsConfig,
 		hint: 'Run `npx @rtorcato/repo-tooling fix tsconfig` to scaffold',
 	},
 	{
--- a/src/languages/js/checks.ts
+++ b/src/languages/js/checks.ts
@@ -126,6 +126,25 @@ const TS_PRESET_STRICTNESS = [
 ] as const
 
 /**
+ * `strict` and every flag it implies. TypeScript lets any one of these be
+ * switched off individually while `strict: true` is still literally present —
+ * the explicit override wins — so a config can keep every marker this check
+ * looks for and still ship with type safety off. Any of them explicitly
+ * `false` is drift.
+ */
+const TS_STRICT_FLAGS = [
+	'strict',
+	'noImplicitAny',
+	'strictNullChecks',
+	'strictFunctionTypes',
+	'strictBindCallApply',
+	'strictPropertyInitialization',
+	'noImplicitThis',
+	'useUnknownInCatchVariables',
+	'alwaysStrict',
+] as const
+
+/**
  * The same two shapes as `matchesBiomeConfig`, for the same reason (#385): the
  * `extends` pointer `fix tsconfig` writes, and the whole preset inlined, which
  * is what `copy tsconfig` drops — `tooling/typescript/tsconfig.base.json` *is*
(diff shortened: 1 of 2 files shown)


rtorcato (commented):

Verified all five required cases against the diff and confirmed the real preset file matches the inline-shape branch:

- `extends: "@rtorcato/repo-tooling/typescript/base"` → `ok` (TS_PRESET_REF matches, strict guard passes)
- `copyPreset('tsconfig')` output (real util, not a fixture) → `ok` — checked `tooling/typescript/tsconfig.base.json` directly: `rootDir` is `${configDir}/src` and all four `TS_PRESET_STRICTNESS` keys (`strict`, `noUncheckedIndexedAccess`, `noImplicitOverride`, `noPropertyAccessFromIndexSignature`) are `true`, so this actually matches the inline branch rather than coincidentally passing
- `{"compilerOptions":{"strict":false}}` → still `drift` (no preset markers, `strict` guard also fires)
- preset-shaped config with `strict: false` → `drift`, and the same for `extends` the preset with `strict: false` — the check disqualifies `strict === false` before evaluating either shape, so the #379-style bypass (markers present, protection disabled, still passes) is closed on this check too
- commented JSONC → `ok`; unparseable → `drift`, no crash — `parseJsonc` returns `null` and `matchesTsConfig` returns `false` cleanly

`matchesTsConfig` reuses the existing `parseJsonc` helper rather than a second parser (same one `matchesBiomeConfig` uses). `fix tsconfig`/`generateTSConfig` are untouched — only the `checks.ts` matcher and its tests changed. Structure and naming (`TS_PRESET_REF`, `TS_PRESET_STRICTNESS`) mirror the #379 Biome pattern. No correctness or security issues found.

rtorcato (commented):

**1 high — the exact bypass class from #379 still applies, one level down.**

`matchesTsConfig` (src/languages/js/checks.ts) only disqualifies a config via `compilerOptions.strict === false`. TypeScript lets any of `strict`'s constituent flags (`noImplicitAny`, `strictNullChecks`, `strictPropertyInitialization`, `useUnknownInCatchVariables`, etc.) be overridden individually even while `strict: true` is still literally present — the override always wins over the implied default. So a config can keep every marker this check looks for and still ship with meaningful type safety turned off, and `doctor` will report `ok`.

Concrete bypass, extends shape:
```json
{
  "extends": "@rtorcato/repo-tooling/typescript/base",
  "compilerOptions": { "noImplicitAny": false }
}
```
`compilerOptions.strict` is `undefined` here (not `=== false`), the guard passes, `extends` matches `TS_PRESET_REF`, and the function returns `true` — `ok` — despite `noImplicitAny` being explicitly disabled.

Concrete bypass, inlined shape (copy the real preset, add one line):
```json
{
  "compilerOptions": {
    "rootDir": "${configDir}/",
    "strict": true,
    "noUncheckedIndexedAccess": true,
    "noImplicitOverride": true,
    "noPropertyAccessFromIndexSignature": true,
    "noImplicitAny": false
  }
}
```
All four `TS_PRESET_STRICTNESS` keys are still `true`, `rootDir` still matches, `strict` is still `true` (not `=== false`) — `matchesTsConfig` returns `true`. `noImplicitAny` is off.

This is the same shape of bug the #379 review blocked on the Biome check (marker present, protection silently disabled, check reports healthy) — just at the level of an individual `strict` sub-flag instead of the top-level switch. Given the `TypeScript` check is *not* optional (unlike `Biome`), a false "ok" here is a stronger signal to a maintainer than the Biome equivalent.

Suggested fix: also disqualify when any of TypeScript's `strict`-implied flags (at minimum `noImplicitAny`, `strictNullChecks`, `strictPropertyInitialization`, `useUnknownInCatchVariables`) is explicitly `false` in `compilerOptions`, regardless of which shape/branch matched.

**Everything else checked clean:**
- `parseJsonc` fails safe — malformed/hostile JSON returns `null`, `matchesTsConfig` returns `false` (drift), doesn't throw out of `doctor`. Confirmed by the new "does not crash on a corrupt one" test.
- Both regexes (`TS_PRESET_REF`, the comment/string-stripping regex in `parseJsonc`) are linear, no nested quantifiers — no ReDoS concern, and unchanged from the already-reviewed #379 machinery.
- No secrets, no shell/SQL construction, no dependency or supply-chain changes — this diff is pure JSON parsing and structural comparison.
- Test coverage is good and does directly assert the `strict: false` bypass is still caught, it just doesn't reach the sub-flag case above.

rtorcato (commented):

Fix verified. `TS_STRICT_FLAGS` (covers all 8 requested flags plus `strict` itself) is checked once via an early return (`if (TS_STRICT_FLAGS.some((key) => compilerOptions[key] === false)) return false`) *before* branching into the `extends` vs. inlined-preset shape — so both branches are guarded by construction, not by duplicated conditionals. It's a single named list, so adding a flag later is one line.

The check uses strict `=== false`, so an absent flag (`undefined === false` → `false`) does not disqualify — only an explicit `false` does. Confirmed no false-positive drift on ordinary tsconfigs that simply don't mention a given flag.

Regression tests cover both bypass shapes from the security review: `extends` + `noImplicitAny: false`, and the inlined preset with `strictNullChecks: false` added (test: "reports drift when a strict-implied flag is switched off on either shape"), plus the original `strict: false` cases for both shapes and the `copy tsconfig` / commented / corrupt-JSONC regressions from the first round. All previously-passing cases are retained, not replaced.

No issues found.

rtorcato (commented):

Re-verified against the prior finding. Both bypasses are closed and covered by new tests:

- `extends` the preset + `strict: false` locally → drift (`TS_STRICT_FLAGS.some(... === false)` runs before the `extends` short-circuit returns `true`).
- Inlined preset with one strict-implied flag (e.g. `noImplicitAny`) flipped false while `strict: true` and all four `TS_PRESET_STRICTNESS` markers remain → drift, for both the `extends` and inlined shapes (`TS_STRICT_FLAGS` now lists `strict` plus all 8 flags it implies).
- `compilerOptions` omitted entirely → correctly falls through to drift (not silently `ok`).
- `parseJsonc` is untouched (reused verbatim from #379) — same comment-stripping regex, same fail-safe `try/catch → null` on unparseable input, same trailing-comma handling. No new ReDoS surface, no dependency changes.

Two real but non-blocking gaps for the next rung:

1. The disqualifier is `=== false` (strict equality). A truthy non-boolean override — `"strict": "false"` (string) or `"strict": 0` — slips past the check. In practice tsc itself rejects/ignores a wrong-typed boolean option rather than silently applying it, so this isn't a working exploit against real type safety, just a gap in the checker's own logic. Info-level; not worth blocking on.
2. Multi-file / extends-chain setups aren't resolved: if `tsconfig.json` extends a local intermediate file that in turn extends the preset and weakens it there, the immediate-`extends` regex won't find the preset reference and reports drift (fails safe, not a false "ok"). Conversely, a *separate* config file (e.g. `tsconfig.build.json`) that actually drives the build but isn't `tsconfig.json` is outside this checker's reach entirely — a known, disclosed structural limit of a single-file presence check, not a regression introduced here.

No criticals or highs. Recommend merge.

PR #389, opened 2026-08-10
fix(gitlab-ci): only emit jobs whose scripts the repo has

### LINKED ISSUE #386: Generated .gitlab-ci.yml calls scripts that may not exist — #364 was only fixed for GitHub Actions
## Summary

#364 (merged as PR #370) made the GitHub Actions generator emit a step **only when the script it calls exists**, via `JobOptions.scripts` / `scriptsOf(pkg)` threaded into `githubJobs`. `gitlabSpec` in the same file never got that treatment:

```ts
export function gitlabSpec(config: ProjectConfig): GitLabSpec {
  const hasTypeScript = config.typescript.enabled
  const hasTests = config.testing.framework !== 'none'
  const hasLint = config.linting.tool !== 'none'
  const hasBuild = config.bundler !== 'none'

  const jobs = [
    ...(hasLint       ? [{ id: 'lint',      script: [lintCommand(config)] }] : []),
    ...(hasTypeScript ? [{ id: 'typecheck', script: ['pnpm typecheck'] }]    : []),
    ...(hasTests      ? [{ id: 'test',      script: [gitlabTestCommand(config)] }] : []),
    ...(hasBuild      ? [{ id: 'build',     script: ['pnpm build'] }]        : []),
  ]
```

It takes only a `ProjectConfig` — there is no way to tell it what the repo actually has.

## Why every generated pipeline is affected

`fix gitlab-ci` builds its config from `inferProjectConfig(pkg)`, which hardcodes `typescript.enabled: true`, `linting.tool: 'biome'`, `testing.framework: 'vitest'` and `bundler: 'tsup'` regardless of the repo. So all four flags are always true, and the emitted `.gitlab-ci.yml` **always** calls `pnpm check`, `pnpm typecheck`, `pnpm exec vitest run` and `pnpm build`.

On a repo missing any of them the pipeline fails at that job — the exact `ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL` failure #364 fixed for GitHub, still live on the GitLab path.

## Suggested fix

Thread the same `scripts` gate through `gitlabSpec` that `githubJobs` already uses. `scriptsOf(pkg)` is exported from `src/languages/js/ci.ts` — reuse it rather than writing a second predicate. Drop a job entirely when the script it would run is absent, matching the GitHub behaviour, and update `generateGitLabCI` and the `gitlab-ci` fixer to pass it through.

Mind the two asymmetries with the GitHub path:
- `gitlabTestCommand` emits `pnpm exec vitest run` (a direct binary invocation, not a script) for vitest, but `pnpm test:e2e` for playwright/cypress. Only the latter needs a script to exist — gate on the right thing per branch rather than assuming every command maps to a script name.
- `build` carries an `artifacts.paths: dist/` block. If the build job is dropped, nothing else should still reference that artifact.

Undefined `scripts` must keep meaning "assume the setup shape", exactly as in `githubJobs` — `setup` writes the scripts itself in the same scaffold, so gating there would strip a working pipeline.

## Verification

Mirror the `generateGitHubActions script gating` tests from `tests/cli/generators/github-actions.test.ts` for `.gitlab-ci.yml`: steps omitted when scripts are missing, kept when present, and the full setup shape when no scripts are supplied. Also assert `stages:` stays valid when jobs are dropped — `renderGitLabCI` derives stages from the jobs, and a pipeline with no jobs still needs a well-formed `stages:` key.

rtorcato:
fix(gitlab-ci): only emit jobs whose scripts the repo has

#364 gated GitHub Actions steps on the scripts a repo defines, but
`gitlabSpec` never got the same treatment. `inferProjectConfig`
hardcodes typescript/biome/vitest/tsup, so every generated
`.gitlab-ci.yml` called `pnpm check`, `pnpm typecheck` and `pnpm
build` regardless of the target and died with
ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL on repos missing them.

Thread the existing `JobOptions.scripts` / `scriptsOf(pkg)` through
`gitlabSpec` and `generateGitLabCI`, and pass it from the `gitlab-ci`
fixer. Undefined `scripts` still means "assume the setup shape".

Two GitLab-only asymmetries:

- vitest runs as `pnpm exec vitest run`, a direct binary call that
  needs no script, while playwright/cypress run `pnpm test:e2e`,
  which does. `gitlabTest` now pairs the command with the script it
  needs, or null.
- dropping the build job takes its `artifacts.paths: dist/` block
  with it; no other job references that artifact.

`renderGitLabCI` already falls back to a single `test` stage when the
job list is empty, so a fully gated-off pipeline stays well-formed -
now covered by a test.

Closes #386

### DIFF
--- a/src/cli/generators/gitlab-ci.ts
+++ b/src/cli/generators/gitlab-ci.ts
@@ -4,9 +4,18 @@ import { renderGitLabCI } from '../../base/ci.js'
 import { gitlabSpec } from '../../languages/js/ci.js'
 import type { ProjectConfig } from '../commands/setup.js'
 
-export async function generateGitLabCI(config: ProjectConfig, targetDir: string) {
+/**
+ * @param scripts The target's package.json scripts, so the pipeline only calls
+ * commands that exist (#386). Omit on the `setup` path, which writes the
+ * scripts itself as part of the same scaffold.
+ */
+export async function generateGitLabCI(
+	config: ProjectConfig,
+	targetDir: string,
+	{ scripts }: { scripts?: Record<string, string> } = {}
+) {
 	const yamlPath = path.join(targetDir, '.gitlab-ci.yml')
 	// ponytail: direct JS import until a second language module ships CI (#287).
-	await fs.writeFile(yamlPath, renderGitLabCI(gitlabSpec(config)))
+	await fs.writeFile(yamlPath, renderGitLabCI(gitlabSpec(config, { scripts })))
 	return '.gitlab-ci.yml'
 }
--- a/src/languages/js/ci.ts
+++ b/src/languages/js/ci.ts
@@ -106,9 +106,14 @@ export function scriptsOf(pkg: Record<string, unknown> | null): Record<string, s
 	return (pkg.scripts as Record<string, string> | undefined) ?? {}
 }
 
+/** The repo defines this script — or we have no scripts to check it against. */
+function hasScript(opts: JobOptions, script: string): boolean {
+	return !opts.scripts || script in opts.scripts
+}
+
 /** Emit a step only when the script it runs exists (or we can't know yet). */
 function stepFor(opts: JobOptions, script: string, step: string): string | null {
-	return !opts.scripts || script in opts.scripts ? step : null
+	return hasScript(opts, script) ? step : null
 }
 
 /** Join the steps that survived gating onto the shared setup preamble. */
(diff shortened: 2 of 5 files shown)


rtorcato (commented):

Reviewed against issue #386's two callouts:

- **vitest asymmetry**: `gitlabTest` pairs `pnpm exec vitest run` with `script: null`, and the gate is `test.script === null || hasScript(opts, test.script)` — so the vitest job is never dropped for a missing script, while playwright/cypress (`test:e2e`) are correctly gated on the script existing. Test `'keeps the vitest job without a test script'` confirms this even with an explicit empty `scripts: {}`.
- **build artifacts**: the `artifacts.paths: dist/` block lives inside the same array literal as the `build` job entry, so dropping the job (`hasBuild && hasScript(opts, 'build')`) drops the artifacts block with it — nothing else references `dist/`. Covered by the "no build script → no build job... artifacts block goes with it" assertions.

Other checks:
- Undefined `scripts` still resolves to "assume the setup shape" via `hasScript`'s `!opts.scripts ||` short-circuit, matching the `setup`-path contract documented in the new JSDoc.
- `scriptsOf`/`hasScript` are reused (refactored out of the existing `stepFor`), not duplicated — no second predicate was written.
- `stages:` staying well-formed when all jobs are dropped relies on pre-existing `renderGitLabCI` fallback behavior (untouched by this diff) and is now exercised by a dedicated test.
- The `test.script: string | null` narrowing in `test.script === null || hasScript(opts, test.script)` type-checks fine under TS control-flow analysis across `||`.

No bugs found. Diff is a faithful mirror of #370's script-gating for the GitLab path.

rtorcato (commented):

No injection risk: `hasScript`/`stepFor` use script names only as `in` lookups against `opts.scripts` (existence checks), never interpolated into the emitted YAML. Every job body remains a fixed literal string (`pnpm typecheck`, `pnpm build`, `pnpm test:e2e`, `pnpm exec vitest run`, `lintCommand(config)` driven by `config.linting.tool`). A malicious script name or value in a repo's `package.json` cannot break out of the YAML structure or reach the runner's shell — it can only flip a job on/off.

Coverage-reduction risk is deliberate and correctly scoped, not a silent regression: `scriptsOf(pkg)` is only passed by the `gitlab-ci` fixer (post-scaffold, reflects the real repo), while `generateGitLabCI`'s `setup` call site omits `scripts`, so `undefined` still means "assume the full setup shape" — matching the precedent set by #364/PR #370 for GitHub Actions. Dropping `build` correctly also drops the now-orphaned `artifacts.paths: dist/` block. Tests cover the empty-pipeline fallback (`stages:` stays well-formed) and the "keeps everything when scripts are present" case, so this isn't gating more than intended.

No secrets, no shell/SQL construction, no dependency changes. Clean.

PR #390, opened 2026-08-10
fix(vitest): emit a config that imports the shipped preset

### LINKED ISSUE #387: fix vitest emits a config the Vitest doctor check can never match
## Summary

Doctor's `Vitest` check expects the config to import our preset:

```ts
check: 'Vitest',
candidates: ['vitest.config.ts', 'vitest.config.js', 'vitest.config.mjs'],
expected: `imports "${PACKAGE}/vitest/config"`,
matcher: /@rtorcato\/(?:js|repo)-tooling\/vitest\/config/,
optional: true,
```

`generateVitestConfig` (`src/cli/generators/testing.ts`) writes a fully hand-rolled config instead:

```ts
const vitestConfig = `import { defineConfig } from 'vitest/config'

export default defineConfig({
  test: {
    globals: true,
    environment: '...',
    setupFiles: ['./vitest.setup.ts'],
    coverage: { provider: 'v8', reporter: ['text', 'lcov'] }
  }
})
`
```

No reference to `@rtorcato/repo-tooling/vitest/config` anywhere. So `fix vitest` — the target doctor's own hint points at — writes a file the check cannot match, and `canFixDrift: true` means `fix` offers it as the remedy for the check it then fails. Re-running produces the same file.

This is the same defect as #365 (biome) and #381 (tsconfig), on the third of the three config-scaffolding targets.

## Why it is less severe than the other two

The `Vitest` check is `optional: true`, so it reports as optional-missing rather than a hard failure. It is still a permanent, unclearable state and a false statement by the tool: the target claims to scaffold the preset and doesn't.

Note the package **does** ship `./vitest/config` (see `exports` in package.json), so the preset the check looks for genuinely exists — this isn't a check pointing at something imaginary.

## Suggested fix

Pick one and say which in the PR body:

**Preferred — make the emitted config import the preset**, the way `fix prettier` re-exports and `fix biome` extends. The generated config's per-project bits (jsdom vs node environment, the JSX/oxc override for React) would then layer on top of the shared preset rather than restating it. Check what `tooling/vitest/vitest.config.mjs` actually exports before assuming it composes cleanly — if the preset can't express the per-project variation, say so rather than forcing it.

**Alternative — relax the check** to accept a well-formed standalone vitest config, as #379 did for Biome. Weaker: it gives up on the preset reaching consumers at all, and preset improvements would never propagate.

Whichever you choose, the check must stay capable of failing — an unrelated or empty `vitest.config.ts` must still be reported.

## Verification

- `fix vitest` output satisfies the `Vitest` matcher, and `doctor` reports `ok` immediately after.
- The existing behaviour that `vitest.setup.ts` is preserved across a re-run (already tested) must keep passing.
- Per-project variation still lands: `environment: 'jsdom'` for a browser project, the JSX override for React.

rtorcato:
## Approach

Took the **preferred** route from the issue: the emitted `vitest.config.ts` now imports the shipped preset. **Not** the alternative (relaxing the check) — the check is unchanged.

The preset composes cleanly, so nothing had to be given up. `tooling/vitest/vitest.config.mjs` exports a plain `UserConfig` object, and `mergeConfig` layers over it — exactly what the already-shipped `vitest/react` preset does. `fix vitest` now writes:

```ts
import base from '@rtorcato/repo-tooling/vitest/config'
import { defineConfig, mergeConfig } from 'vitest/config'

export default mergeConfig(
  base,
  defineConfig({
    test: {
      environment: 'node',
      setupFiles: ['./vitest.setup.ts']
    }
  })
)
```

Per-project variation survives and is still emitted explicitly, on top of the preset rather than restating it:

- `environment: 'jsdom'` vs `'node'` — a scalar, so the override wins over the preset's `'node'`.
- The Next.js `oxc: { jsx: { runtime: 'automatic' } }` override, unchanged apart from indentation.

## Three preset fixes this exposed

Making consumers actually import the preset surfaced defects that were dormant while nobody did:

- **`resolve.alias` pointed inside `node_modules`.** Both presets built `@`/`~` aliases from their own `__dirname`, i.e. `<install dir>/tooling/vitest/src` — a directory that exists in no consumer and isn't in the published tarball either (not in `files`). Dead today; it would have silently broken `@/…` imports in every repo that took the new generated config. Removed from both presets.
- **The preset omitted `lcov`.** The CI we generate uploads coverage to Codecov, and the old hand-rolled config was the only thing asking for lcov. Adding it to the preset means the generated config doesn't restate the coverage block at all — which also avoids `mergeConfig` concatenating two reporter arrays into a list with a duplicate `text`.
- **`vitest.config.d.mts` imported a type that doesn't exist.** vitest 4 exports vite's `UserConfig` as `ViteUserConfig`; `import type { UserConfig } from 'vitest/config'` is an error. Invisible while no TypeScript file imported the preset — it would now land in every consumer's `tsc --noEmit`.

## The check keeps its teeth

The `Vitest` check and its matcher are untouched. Verified against a real fixture:

| state | doctor |
|---|---|
| after `fix vitest` | `ok` — `vitest.config.ts imports "@rtorcato/repo-tooling/vitest/config"` |
| `export default {}` | `drift` — `found but does not imports "…/vitest/config"` |

`tests/cli/generators/testing.test.ts` asserts both directions against the real `FILE_CHECKS` matcher rather than a copy of the regex, so the two can't drift apart.

## One correction to the issue

The issue lists "`vitest.setup.ts` is preserved across a re-run (already tested)" as a guardrail. **It wasn't tested** — the fixer does save and restore the file around `generateVitestConfig`, but nothing asserted it, so this change could have taken the restore with it unnoticed. Second commit adds that test.

## Verification

- `pnpm run check`, `tsc --noEmit --project src/cli/tsconfig.json` — clean.
- `vitest run` — 839 passed / 55 files. The repo's own `vitest.config.mjs` merges the same preset, so the alias and reporter changes are exercised by every test in the suite.
- Separately typechecked the `mergeConfig(base, defineConfig({…}))` shape under `--strict` to confirm the composition is valid TypeScript for consumers, not just valid at runtime.

Closes #387

### DIFF
--- a/src/cli/generators/testing.ts
+++ b/src/cli/generators/testing.ts
@@ -24,26 +24,28 @@ export async function generateVitestConfig(config: ProjectConfig, targetDir: str
 	// and Vite 8 silently ignores the old `esbuild` key.
 	const jsxOverride =
 		config.projectType === 'nextjs-app'
-			? `  // tsconfig sets jsx: "preserve" for Next's compiler; vitest needs a
-  // transform of its own or .tsx tests won't parse.
-  oxc: { jsx: { runtime: 'automatic' } },
+			? `    // tsconfig sets jsx: "preserve" for Next's compiler; vitest needs a
+    // transform of its own or .tsx tests won't parse.
+    oxc: { jsx: { runtime: 'automatic' } },
 `
 			: ''
 
-	const vitestConfig = `import { defineConfig } from 'vitest/config'
+	// Layered on the shipped preset rather than hand-rolled: the preset owns
+	// globals, the shared excludes and the v8/lcov coverage reporters, so preset
+	// improvements reach consumers — and doctor's `Vitest` check (which looks for
+	// this import) can actually pass after `fix vitest` (#387).
+	const vitestConfig = `import base from '@rtorcato/repo-tooling/vitest/config'
+import { defineConfig, mergeConfig } from 'vitest/config'
 
-export default defineConfig({
-${jsxOverride}  test: {
-    globals: true,
-    environment: '${config.testing.environment === 'browser' ? 'jsdom' : 'node'}',
-    setupFiles: ['./vitest.setup.ts'],
-    coverage: {
-      provider: 'v8',
-      // lcov feeds Codecov; text is the local console summary.
-      reporter: ['text', 'lcov']
+export default mergeConfig(
+  base,
+  defineConfig({
+${jsxOverride}    test: {
+      environment: '${config.testing.environment === 'browser' ? 'jsdom' : 'node'}',
+      setupFiles: ['./vitest.setup.ts']
     }
-  }
-})
+  })
+)
 `
 
 	await fs.writeFile(vitestConfigPath, vitestConfig)
--- a/tests/cli/commands/fix.test.ts
+++ b/tests/cli/commands/fix.test.ts
@@ -1033,6 +1033,16 @@ describe('fix + lockfile', () => {
 		expect(lock.config.linting.tool).toBe('biome')
 	})
 
+	it('fix vitest --yes regenerates the config but keeps an existing vitest.setup.ts', async () => {
+		const dir = newTmpDir()
+		await seedPackageJson(dir)
+		await fs.writeFile(join(dir, 'vitest.setup.ts'), '// my real setup\n')
+		await fixCommand('vitest', { directory: dir, yes: true })
+		expect(await fs.readFile(join(dir, 'vitest.setup.ts'), 'utf-8')).toBe('// my real setup\n')
+		const config = await fs.readFile(join(dir, 'vitest.config.ts'), 'utf-8')
+		expect(config).toContain("from '@rtorcato/repo-tooling/vitest/config'")
+	})
+
 	it('fix vitest --yes on a jest-locked project auto-resyncs the lockfile', async () => {
 		const dir = newTmpDir()
 		await seedPackageJson(dir)
--- a/tests/cli/generators/__snapshots__/snapshot.test.ts.snap
+++ b/tests/cli/generators/__snapshots__/snapshot.test.ts.snap
@@ -324,19 +324,17 @@ exports[`generator output snapshots > package.json (library) 1`] = `
 `;
 
 exports[`generator output snapshots > vitest.config.ts (node library) 1`] = `
-"import { defineConfig } from 'vitest/config'
+"import base from '@rtorcato/repo-tooling/vitest/config'
+import { defineConfig, mergeConfig } from 'vitest/config'
 
-export default defineConfig({
-  test: {
-    globals: true,
-    environment: 'node',
-    setupFiles: ['./vitest.setup.ts'],
-    coverage: {
-      provider: 'v8',
-      // lcov feeds Codecov; text is the local console summary.
-      reporter: ['text', 'lcov']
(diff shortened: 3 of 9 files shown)


rtorcato (commented):

Verified against the diff and issue #387.

- **Per-project variation survives.** `environment: 'jsdom'|'node'` is a scalar in the second `defineConfig` arg to `mergeConfig`, so it correctly overrides the preset's `'node'` default. The Next.js `oxc.jsx.runtime: 'automatic'` override is still emitted at the same nesting level as `test`, just re-indented under `mergeConfig(base, defineConfig({...}))` — syntactically consistent with the pre-PR version, and confirmed by the updated snapshot test and the new "keeps the Next.js jsx transform override on top of the preset" test.
- **Preset edits are safe.** `resolve.alias` built from the preset's own `__dirname` pointed at `<install dir>/tooling/vitest/src`, which is genuinely absent from `package.json#files` — confirmed dead code in any published consumer, safe to remove. Adding `lcov` to the preset's reporter array is additive; since the generated config no longer restates `test.coverage` at all, there's no `mergeConfig` array-concat duplicate-`text` risk. The `UserConfig` → `ViteUserConfig` rename in the `.d.mts` files is type-only and scoped to those two files.
- **`vitest.setup.ts` preservation is now actually tested** (`tests/cli/commands/fix.test.ts`), correcting a real gap the issue had assumed was covered.
- **The `Vitest` check keeps its teeth.** Matcher/check definitions are untouched; the new test in `tests/cli/generators/testing.test.ts` asserts both the positive match against the generated config and a negative match against `export default {}` using the real `FILE_CHECKS` matcher, so the two can't drift apart.

No bugs found. Nit: the PR body itself flags and fixes the issue's incorrect claim about setup-file preservation being pre-tested — good catch, already handled in the diff.

rtorcato (commented):

No injection, secrets, or supply-chain issues in this diff — it's config-generation and shipped-preset changes only, no new dependencies, no shell/SQL construction, no untrusted input handling.

Specific to the preset propagation risk:
- `exclude` in `tooling/vitest/vitest.config.mjs` is untouched (`[...configDefaults.exclude, '.claude/**']`); coverage `reporter` gains `lcov` and keeps the existing entries — strictly additive, not narrowed. No path by which a consumer's suite would silently run fewer tests or lose coverage gating on upgrade.
- The removed `resolve.alias` block built its path from each preset file's own `__dirname` (i.e. inside `node_modules`, and not included in the published `files`), so it never resolved to anything in a real consumer — removing it deletes dead/broken code, not working behavior.
- `setupFiles: ['./vitest.setup.ts']` is still written as a literal relative path into the generated config in the consumer's own target directory, so it resolves against the consuming repo as before — confirmed unchanged in the diff.
- The `vitest.config.d.mts` type fix (`UserConfig` → `ViteUserConfig`) is a correctness-only change with no security impact.

One thing worth a maintainer's eye, not blocking: the generated config now hardcodes the literal string `@rtorcato/repo-tooling/vitest/config` in `src/cli/generators/testing.ts` rather than deriving it from a shared `PACKAGE`-style constant (the pre-existing doctor check's matcher already hardcodes `@rtorcato`, so this isn't a new pattern, and it's a self-reference to the tool's own published export rather than an org-private value like a family glob or pinned version — but a literal copy in two places can drift if the package is ever renamed again).

PR #394, opened 2026-08-11
fix(doctor): flag agent docs that name the dead js-tooling bin

### LINKED ISSUE #393: fix ai reports 'already configured' for docs that reference the dead js-tooling bin
## Problem

`fix ai` short-circuits on "the files exist" rather than "their contents are
current", so any repo scaffolded before the `js-tooling` → `repo-tooling`
rename keeps agent instructions that point at a **dead package and bin**, forever.

Reproduced in `db-x`, `cf-common` and `supabase-common` while migrating them
off the unpublished `@rtorcato/js-tooling` (#327 fallout):

```console
$ repo-tooling fix ai --yes
✅ AI setup is already configured

$ cat CLAUDE.md
<!-- js-tooling:start -->
See @AGENTS.md for project and agent guidance (kept in sync by `js-tooling fix ai`).
<!-- js-tooling:end -->
```

`js-tooling fix ai` is not a runnable command — the bin is `repo-tooling`, and
`@rtorcato/js-tooling` is 404 on the registry. Doctor actively reports this as
healthy:

```
✅ AI setup — ok
   AGENTS.md has the js-tooling block
```

## Why it matters

This is the one check whose entire audience is agents, and it is handing them
a command that cannot work. It is also self-concealing: `doctor` says ok, and
`fix ai` says already configured, so nothing surfaces it.

Note the individual targets *do* work — `fix claude-skill`, `fix cursor-rules`,
`fix copilot-instructions` and `fix agents-md` each rewrote correctly and even
removed the legacy `js-tooling` filenames via `legacyTarget`. Only the umbrella
`ai` target and the doctor check are wrong, which is what makes it easy to miss.

## Suggested fix

Have the AI-setup check treat a managed block containing the legacy package or
bin name as **drift** rather than ok, and let `fix ai` rewrite in that case. The
detection string already exists as `LEGACY_LOCKFILE_NAME`'s sibling concept in
`src/cli/utils/lockfile.ts`.

Worth deciding separately: `src/cli/generators/agent-rules.ts` still emits
`<!-- js-tooling:start -->` markers. Keeping the old marker text is defensible
for back-compat, but then the check must not use "has the js-tooling block" as
its success message, since that is what makes the stale state look intentional.

## Acceptance

- A repo whose `CLAUDE.md` / `AGENTS.md` mention `js-tooling` reports drift.
- `fix ai` refreshes them in place.
- A test covers the legacy-content case, not just the missing-file case.

rtorcato:
## What

`fix ai` short-circuited on "the files exist" rather than "their contents are current", so any repo scaffolded before the `js-tooling` → `repo-tooling` rename kept agent instructions pointing at a dead package and bin — `js-tooling fix ai` is not runnable and `@rtorcato/js-tooling` is 404 on the registry. Self-concealing: `doctor` said ok and `fix ai` said "already configured".

`checkAiSetup` now inspects the managed block in `AGENTS.md`, `CLAUDE.md` and `.github/copilot-instructions.md` and returns `drift` when the block body still names the legacy package/bin. The existing `ai` fixer is already `canFixDrift: true` + `safe-merge`, so `fix ai --yes` refreshes the blocks in place without touching surrounding content.

## The marker subtlety

`agent-rules.ts` deliberately keeps emitting `<!-- js-tooling:start -->` / `<!-- js-tooling:end -->` — changing the marker text would orphan every block already written and append a second one on the next fix. So detection matches the **block body only** (`hasStaleAgentBlock` slices between the markers). Flagging any occurrence of `js-tooling` in the file would flag the markers the tool itself writes and leave every repo permanently dirty.

The `ok` detail also no longer reads "AGENTS.md has the js-tooling block" — as the issue notes, that phrasing is what made the stale state look intentional.

## Changes

- `src/cli/utils/lockfile.ts` — `LEGACY_TOOL_NAME = 'js-tooling'`, with `LEGACY_LOCKFILE_NAME` derived from it
- `src/cli/generators/agent-rules.ts` — export the markers plus `hasStaleAgentBlock()`
- `src/base/checks.ts` — `checkAiSetup` scans the three block-managed files, reports drift, and drops the misleading ok detail

## Tests

- `doctor.test.ts` — a `CLAUDE.md` whose block names the legacy bin reports `drift` (the legacy-content case, not just the missing-file case)
- `fix.test.ts` — `fix ai --yes` rewrites stale `CLAUDE.md` + `AGENTS.md` blocks in place, preserving the user's own content

`pnpm run verify` passes: biome, typecheck, 836 tests, build.

Closes #393

### DIFF
--- a/src/base/checks.ts
+++ b/src/base/checks.ts
@@ -1,5 +1,6 @@
 import path from 'node:path'
 import fs from 'fs-extra'
+import { BLOCK_START, hasStaleAgentBlock } from '../cli/generators/agent-rules.js'
 import { BADGE_START, hasPublicOnlyBadges } from '../cli/generators/badges.js'
 import { type DetectedLanguage, detectNestedLanguages } from '../cli/utils/detect-language.js'
 import type { CheckResult } from './types.js'
@@ -378,21 +379,35 @@ export async function checkCoverageUpload(dir: string): Promise<CheckResult> {
 }
 
 export async function checkAiSetup(dir: string): Promise<CheckResult> {
-	// Consider AI setup present if AGENTS.md carries the js-tooling block or the
-	// Claude skill is installed — the two primary markers `fix ai` writes.
-	const agentsPath = path.join(dir, 'AGENTS.md')
-	const hasAgentsBlock =
-		(await fs.pathExists(agentsPath)) &&
-		(await fs.readFile(agentsPath, 'utf8')).includes('<!-- js-tooling:start -->')
+	// Consider AI setup present if one of the block-managed files carries the
+	// agent block or the Claude skill is installed — the markers `fix ai` writes.
+	let blockFile: string | null = null
+	const stale: string[] = []
+	for (const rel of ['AGENTS.md', 'CLAUDE.md', path.join('.github', 'copilot-instructions.md')]) {
+		const filepath = path.join(dir, rel)
+		if (!(await fs.pathExists(filepath))) continue
+		const content = await fs.readFile(filepath, 'utf8')
+		if (!content.includes(BLOCK_START)) continue
+		blockFile ??= rel
+		if (hasStaleAgentBlock(content)) stale.push(rel)
+	}
 	const hasSkill =
 		(await fs.pathExists(path.join(dir, '.claude', 'skills', 'repo-tooling.md'))) ||
 		// Pre-rename name — still counts as present until `fix` migrates it.
 		(await fs.pathExists(path.join(dir, '.claude', 'skills', 'js-tooling.md')))
-	if (hasAgentsBlock || hasSkill) {
+	if (stale.length > 0) {
+		return {
+			check: 'AI setup',
+			status: 'drift',
+			detail: `${stale.join(', ')} still point agents at the dead \`js-tooling\` package/bin`,
+			hint: 'Run `npx @rtorcato/repo-tooling fix ai` to refresh the agent instructions in place',
+		}
+	}
+	if (blockFile || hasSkill) {
 		return {
 			check: 'AI setup',
 			status: 'ok',
-			detail: hasAgentsBlock ? 'AGENTS.md has the js-tooling block' : '.claude skill installed',
+			detail: blockFile ? `${blockFile} has the agent block` : '.claude skill installed',
 		}
 	}
 	return {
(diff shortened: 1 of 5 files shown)


rtorcato (commented):

No security concerns. This diff is detection-only: `hasStaleAgentBlock` does a plain string `.includes()` check for a hardcoded constant (`LEGACY_TOOL_NAME = 'js-tooling'`) scoped to the block body between fixed markers — no regex built from external input, no shell/SQL construction. The three files it inspects (`AGENTS.md`, `CLAUDE.md`, `.github/copilot-instructions.md`) are fixed relative paths, not user-controlled, and this PR touches no write/overwrite path (the existing `fix ai` writer is unchanged here). No secrets, no dependency or supply-chain changes.

rtorcato (commented):

Reviewed against issue #393's acceptance criteria and the diff only.

**Marker/body distinction (the key risk area):** `hasStaleAgentBlock` in `src/cli/generators/agent-rules.ts` correctly slices between `BLOCK_START`/`BLOCK_END` and only checks the body for `LEGACY_TOOL_NAME` ('js-tooling') — it never matches the markers themselves, which are deliberately kept as `<!-- js-tooling:start/end -->` for back-compat. Confirmed no false-positive risk: the source template (`tooling/claude/repo-tooling.md`) that `fix ai` regenerates from contains no `js-tooling` string in its body, so a freshly-fixed block can never re-trip the drift check — no infinite-drift loop.

**Reaches a clean state:** `fix.test.ts`'s new test exercises `fix ai --yes` on stale `CLAUDE.md` + `AGENTS.md` blocks and asserts both are rewritten (`repo-tooling fix ai` present, `js-tooling fix ai` absent), including preservation of surrounding user content in `AGENTS.md`. This is exactly the acceptance bar in #393.

**checkAiSetup logic:** scans all three block-managed files (`AGENTS.md`, `CLAUDE.md`, `.github/copilot-instructions.md`), returns `drift` if any is stale, otherwise `ok` with a corrected (non-misleading) detail string per the issue's specific request. `LEGACY_LOCKFILE_NAME` is now derived from the new `LEGACY_TOOL_NAME` constant with identical resulting value (`.js-tooling.json`), so no behavior change there.

No bugs or convention violations found. No repo-root `CLAUDE.md` exists to check against for additional rules.

PR #398, opened 2026-08-11
feat(doctor): audit milestone hygiene, close completed milestones

### LINKED ISSUE #397: doctor: milestone hygiene — flag 100%-complete milestones left open, and empty ones
`doctor` already audits GitHub-side state (branch protection, merge settings, workflow permissions, code-scanning gate), so milestone hygiene belongs there too.

## The problem, measured

A survey of 10 family repos found **13 open milestones that are 100% complete and were never closed**:

| repo | milestone | closed issues |
|---|---|---|
| infra-x | v0.1 — Usable for the author | 29 |
| supabase-common | Helpers v1.x | 6 |
| repo-tooling | Language module: Swift | 6 |
| supabase-common | Release & CI hardening | 5 |
| cf-common | Beta | 4 |
| db-common | v1.0 — Stable API | 3 |
| supabase-common | Post-1.0 | 3 |
| repo-tooling | Consumer-repo extraction | 3 |
| repo-tooling | Language modules: Perl & Python | 2 |
| repo-tooling | Verification debt | 2 |
| react-common | Backlog | 2 |
| db-common | Backlog | 1 |
| browser-common | Tooling & release | 1 |

The consequence: **"open" stops meaning "in flight."** Finished work is indistinguishable from active work, so the milestone list carries no signal — you cannot glance at it and learn what is being worked on.

Separately, **6 milestones have zero issues**, which renders a permanent 0% (GitHub computes the bar as `closed / total`, so an empty milestone can never move):

- js-common #1 v2.x — Shipped modules [closed], #2 Beta — npm preview, #3 v1.0 — Stable API *(now populated — see rtorcato/js-common#173–#178)*
- db-common #1 v0.2 — First helpers [closed], #2 Beta — npm preview
- react-common #1 v0.1 — Scaffold [closed]

## Ask

Add a `milestones` check to `doctor`:

- [ ] **100%-complete but open** → drift. `open_issues == 0 && closed_issues > 0` on an open milestone. Hint: close it.
- [ ] **Empty milestone** → drift. `open_issues == 0 && closed_issues == 0`. Hint: file its issues or delete it; it shows a permanent 0%.
- [ ] **Catch-all milestone** → warn on titles matching `backlog` / `post-\d` / `someday` (case-insensitive). A milestone with no completion criterion can never close, so it is structurally a label — suggest converting it. Present in db-common, react-common, supabase-common.
- [ ] Skip the whole check when the repo has no milestones at all (swift-common) — that is a legitimate choice, not drift.
- [ ] `fix milestones` should **only** close the unambiguous 100%-complete case. Never delete a milestone and never create one — both destroy or invent planning intent.

Read-only via `gh api repos/{owner}/{repo}/milestones?state=all`; no extra auth beyond what the existing GitHub checks already use.

## Policy this encodes

For the docs: a milestone is a **release gate**, not a chore bucket. Milestone an issue only if leaving it undone would block declaring that stage complete — refactors, CI, dependency bumps and docs typos get no milestone. Otherwise the percentage inverts: a v1.0 milestone reads 80% because it is full of chores while the three items that actually define v1.0 stay open.

Corollary worth stating: `ai-ready` and milestone-worthy are near-mutually-exclusive. `ai-ready` means mechanical and bounded; milestone-worthy means it shapes the public API. An agent-sized task that redefines public exports is precisely what should not run unattended.

rtorcato:
Adds a `Milestones` check. An open milestone at 100% (open_issues 0,
closed_issues > 0) and an empty one both report drift — GitHub renders
the bar as closed/total, so an empty milestone is a permanent 0%.

A catch-all title (backlog / post-N / someday) rides in the detail
rather than as its own finding: converting one to a label is a
planning decision `fix` must not make, and a permanently-unfixable
drift row would nag forever. A repo with no milestones is skipped;
that is a legitimate choice.

`fix milestones` closes only the unambiguous 100%-complete case. It
never deletes or creates a milestone: both destroy or invent
planning intent.

Closes #397

### DIFF
--- a/apps/docs/docs/guides/cli.md
+++ b/apps/docs/docs/guides/cli.md
@@ -193,6 +193,7 @@ Implementation note: the preview is computed by shadow-running the fixer in a te
 | `renovate` | `renovate.json` (alternative to Dependabot) |
 | `codeql` | `.github/workflows/codeql.yml` (security scanning) |
 | `github-settings` | branch protection + auto-merge + workflow permissions + a code-scanning branch ruleset (when CodeQL is on) via `gh api` (mutates the remote repo) |
+| `milestones` | closes 100%-complete open milestones via `gh api` (mutates the remote repo). Never deletes or creates one — see [milestone hygiene](./git-flow.md#milestones) |
 | `codeowners` | `.github/CODEOWNERS` with commented examples |
 | `community-health` | `CONTRIBUTING.md`, `SECURITY.md`, PR + issue templates |
 | `lockfile` | records current tool choices in the repo-tooling lockfile |
--- a/apps/docs/docs/guides/git-flow.md
+++ b/apps/docs/docs/guides/git-flow.md
@@ -83,6 +83,30 @@ under an admin's PAT / GitHub App token and leave "include administrators" off,
 so that identity bypasses.) This bypass is for the release bot only — humans
 always go through a PR.
 
+## Milestones
+
+A milestone is a **release gate, not a chore bucket**. Milestone an issue only if
+leaving it undone would block declaring that stage complete; refactors, CI,
+dependency bumps and docs typos get no milestone. Otherwise the percentage
+inverts — a `v1.0` milestone reads 80% because it is full of chores while the
+three items that actually define v1.0 stay open.
+
+Corollary: `ai-ready` and milestone-worthy are near-mutually-exclusive.
+`ai-ready` means mechanical and bounded; milestone-worthy means it shapes the
+public API, which is precisely what should not run unattended.
+
+`doctor` audits this as the `Milestones` check:
+
+| Finding | Why it matters |
+| --- | --- |
+| 100% complete but still open | "Open" stops meaning "in flight", so the milestone list carries no signal |
+| No issues at all | GitHub renders the bar as `closed / total`, so an empty milestone is a permanent 0% |
+| Titled `backlog` / `post-N` / `someday` | No completion criterion means it can never close — that is a label |
+
+`fix milestones` closes only the first case. It never deletes a milestone and
+never creates one, because both destroy or invent planning intent. A repo with
+no milestones at all is skipped — that is a legitimate choice.
+
 ## Single source of truth
 
 - This standard, `CONTRIBUTING.md` (branch prefixes, the ≤67-char PR-title
--- a/src/base/fixers.ts
+++ b/src/base/fixers.ts
@@ -26,6 +26,7 @@ import { detectLanguage } from '../cli/utils/detect-language.js'
 import type { Lockfile } from '../cli/utils/lockfile.js'
 import { resolveLanguageModule } from '../languages/registry.js'
 import { applyGithubSettings } from './github-settings.js'
+import { closeCompletedMilestones } from './milestones.js'
 import type { CheckResult } from './types.js'
 
 /** A parsed package.json, or null when the repo has none (any non-JS repo). */
@@ -159,6 +160,20 @@ export const BASE_FIXERS: Fixer[] = [
 		canFixDrift: true,
 		async run({ targetDir }) {
 			return { filesWritten: await applyGithubSettings(targetDir) }
+		},
+	},
+	{
+		target: 'milestones',
+		description:
+			'Close 100%-complete open milestones on GitHub via gh api (mutates the remote repo, not files). Never deletes or creates one',
+		appliesTo: ['Milestones'],
+		outputs: ['GitHub milestones (remote, via gh api)'],
+		// safe-add for the same reason github-settings is: it exempts this fixer
+		// from the `--diff` shadow-run, which executes run() for a mere preview.
+		riskLevel: 'safe-add',
+		canFixDrift: true,
+		async run({ targetDir }) {
+			return { filesWritten: await closeCompletedMilestones(targetDir) }
 		},
 	},
 	{
--- a/src/base/milestones.ts
+++ b/src/base/milestones.ts
@@ -1,0 +1,159 @@
+import path from 'node:path'
+import chalk from 'chalk'
+import fs from 'fs-extra'
+import { type GhExec, realGhExec } from './github-settings.js'
+import type { CheckResult } from './types.js'
+
+/**
+ * Milestone hygiene (#397). `doctor` already audits GitHub-side state, and an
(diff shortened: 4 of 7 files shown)


rtorcato (commented):

**Verdict: clean — no criticals, no highs.**

Reviewed the diff for injection, secrets, unsafe shell construction, and dependency risk.

- **Injection boundary is handled correctly.** The only remote mutation (`PATCH repos/{owner}/{repo}/milestones/${m.number}`) interpolates `m.number`, which is guarded by `Number.isInteger(m.number)` in `classify()` before it ever reaches the template string (`src/base/milestones.ts`). `gh` is invoked with an args array (same `GhExec` seam as `github-settings.ts`), not a shell string, so there's no shell-injection vector even without that guard — the integer check is solid defense-in-depth on top of that.
- **No secrets.** No new credentials, tokens, or env vars introduced; reuses the existing authenticated `gh` seam.
- **Mutation scope matches the stated contract.** `closeCompletedMilestones` only ever issues `state=closed` PATCHes for the unambiguous 100%-complete case. It never deletes or creates milestones — confirmed both by code path (empty milestones only get a console notice, no `gh` call) and by the test `never deletes an empty milestone`.
- **Untrusted text (milestone titles) is only used for regex matching and CLI-output strings** (`detail`/`hint`), never passed to a shell, HTML sink, or file path — no injection surface there.
- **No dependency changes.** Only reuses already-installed packages (`chalk`, `fs-extra`, `path`) and internal modules.
- **Self-skip logic is safe**: no `.git` → no `gh` spawn at all, keeping tmp-dir doctor runs offline (verified by the "never spawns gh outside a git repo" test).

No blocking issues.

rtorcato (commented):

**Bug** — `readMilestones()` in `src/base/milestones.ts` will never actually list milestones.

```ts
const r = await gh([
	'api',
	'repos/{owner}/{repo}/milestones',
	'-f',
	'state=all',
	'-F',
	'per_page=100',
])
```

Per `gh api --help`: "adding request parameters will automatically switch the request method to `POST`. To send the parameters as a `GET` query string instead, use `--method GET`." Since `-f`/`-F` are present here with no explicit `-X`/`--method`, `gh` sends this as `POST repos/{owner}/{repo}/milestones` — the *create milestone* endpoint — not the intended `GET` list. Without a `title` field, GitHub will reject it (422), so `r.ok` is `false`, `readMilestones` returns `null`, and `checkMilestones`/`closeCompletedMilestones` will always self-skip with "could not read milestones" against a real repo. The whole feature silently no-ops in production.

The comment above the call ("`-f`/`-F` on a GET land in the query string...") describes the intended behavior but the code doesn't request `GET` explicitly, so it doesn't hold. Every other `-f`/`-F` usage in `github-settings.ts` (the file this explicitly says it shares a seam with) pairs the fields with an explicit `-X PATCH`/`-X PUT` — this call needs the same treatment, e.g. add `'-X', 'GET',` (or `'--method', 'GET'`) before the field flags.

This wasn't caught by the test suite because `fakeGh` mocks the exec function directly and never exercises `gh`'s real flag-to-method inference.

Everything else checked out: matches the issue's spec (100%-complete, empty, catch-all-as-note-not-drift, skip-when-no-milestones, fixer only closes/never deletes/creates), `Number.isInteger` guard before interpolating into the PATCH path is a nice touch, docs and `fix-targets.ts` mapping are consistent, and test coverage of the logic (once the `gh` call is fixed) is thorough.

rtorcato (commented):

Round 2: the previously reported defect is fixed. `readMilestones()` in `src/base/milestones.ts` now sends `['api', '-X', 'GET', 'repos/{owner}/{repo}/milestones', '-f', 'state=all', '-F', 'per_page=100']` — the explicit `-X GET` stops `gh` from inferring POST from the presence of `-f`/`-F`.

The test gap is also closed: `fakeGh` in `tests/base/milestones.test.ts` now has a `ghMethod()` helper that mirrors gh's real flag-to-method inference (explicit `-X` wins; otherwise `-f`/`-F` present → POST), and unrecognized methods return `ok: false`. Verified the regression would actually be caught: the `lists milestones with an explicit GET, not gh's inferred POST` test asserts the exact args array passed to `exec`, so reverting the fix fails that assertion outright regardless of the mock's response logic; separately, `flags a 100%-complete milestone left open` and the other drift-detection tests would flip from `drift` to `ok` (self-skip) if `readMilestones` started returning `null` again. This is a real fix, not a mock that happens to agree with itself.

Rest of the diff is unchanged in substance from round 1 and still checks out: `Number.isInteger(m.number)` guard before interpolating into the PATCH path, empty/catch-all/no-git-repo handling matches the issue spec, `fix-targets.ts` and docs stay consistent, and the fixer never deletes or creates a milestone (confirmed by the `never deletes an empty milestone` test).

No blocking issues.

rtorcato (commented):

Re-reviewed after the round-2 fix. No criticals, highs, or mediums — clean.

**Mutation scope confirmed limited to the unambiguous case.** `closeCompletedMilestones` issues exactly one mutating call path: `PATCH repos/{owner}/{repo}/milestones/${m.number}` with `state=closed`, gated by `state === 'open' && open_issues === 0 && closed_issues > 0 && Number.isInteger(m.number)`. No DELETE, no create path — matches the issue's explicit "never delete, never create" requirement, and `tests/base/milestones.test.ts` asserts no DELETE call is ever made.

**The `-X GET` fix looks correct and closes the likely round-1 defect.** `gh api` infers POST when `-f`/`-F` flags are present without an explicit `-X`, which would have sent the read-only milestone list request to the *create*-milestone endpoint — a read silently becoming a mutation. `readMilestones` now passes `-X GET` explicitly, is covered by a dedicated test (`lists milestones with an explicit GET, not gh's inferred POST`), and the PATCH call correctly keeps its own explicit `-X PATCH`.

**Injection boundary is guarded.** `m.number` is interpolated into the PATCH path but is constrained to `Number.isInteger(m.number)` before any milestone reaches the mutating loop, and `gh(args: string[])` is invoked with an argument array (not a shell string), so there's no command-injection surface even if a title contained shell metacharacters. Titles themselves are never used in a path or shell context, only in output text.

**No secrets, no new dependencies, no supply-chain changes.** Docs additions accurately describe the implemented behavior.

Nit (non-blocking): milestone titles are echoed into `console.error`/detail strings unsanitized — pre-existing pattern from other `doctor` checks, not a new risk, not blocking.

PR #399, opened 2026-08-11
feat(ai): scaffold .claude/settings.json worktree symlinks

### LINKED ISSUE #396: Scaffold .claude/settings.json with worktree node_modules symlinking in the AI setup fixer
Claude Code worktrees start with no `node_modules`, so every agent that runs in one pays a full `pnpm install` before it can typecheck, lint or test. Claude Code can symlink the directory from the main checkout instead:

```json
{
  "worktree": {
    "symlinkDirectories": ["node_modules"]
  }
}
```

`repo-tooling`'s own `ai-issue-loop` and `_ai-workflow` both run agents in per-issue worktrees, so this is squarely a repo-tooling concern — and it's the difference between an agent starting work immediately and burning a minute-plus install first, on every issue.

## Current state across the family

Only three repos have a tracked `.claude/settings.json` at all, and none carry the setting:

| repo | `.claude/settings.json` | has the setting |
|---|---|---|
| js-common | tracked | ✓ (rtorcato/js-common#171) |
| browser-common | tracked | – |
| react-common | tracked | – |
| swift-common | tracked | n/a — no node_modules |
| api-common, cf-common, db-common, db-x, infra-x, repo-tooling, supabase-common, hedgehog, shared-docs, create-agentic-app | none | – |

Hand-editing twelve repos is the wrong shape; the fixer should own it.

## Ask

- [ ] Extend the AI-setup fixer (the one that already emits `.claude/skills/<name>.md`) to also scaffold or upsert `.claude/settings.json` with the `worktree.symlinkDirectories` entry.
- [ ] **Upsert, don't overwrite** — repos already have hooks and permissions blocks in that file (js-common has a `hooks` block). Merge the `worktree` key and leave everything else untouched, the way `copilot-instructions` and `agents-md` handle their delimited blocks.
- [ ] Skip it for the `swift-library` preset — no `node_modules` there, so the key would be noise.
- [ ] `doctor` check reporting a repo whose `.claude/settings.json` lacks the entry, demotable to "intentionally declined" via the lockfile like other optional checks.

## Reference

rtorcato/js-common#171 adopts it by hand, and is also worth a look for a second bug it fixes: a pre-commit hook that pipes staged files to `biome format` aborts the commit when every staged file is biome-ignored (which `.claude/**` is). If the husky/lint-staged preset here emits that shape, it wants `--no-errors-on-unmatched` too.

rtorcato:
Closes #396

A Claude Code worktree starts with no `node_modules`, so every agent working one pays a full `pnpm install` before it can typecheck, lint or test. `fix ai` now upserts the `worktree.symlinkDirectories` entry into `.claude/settings.json` so Claude symlinks the directory from the main checkout instead:

```json
{
  "worktree": {
    "symlinkDirectories": ["node_modules"]
  }
}
```

## What changed

- `installClaudeSettings` in `src/cli/generators/agent-rules.ts`, called from the `installAiSetup` umbrella — so `fix ai`, `setup` and the preset file list all get it.
- **Upsert, not overwrite.** Only `worktree.symlinkDirectories` is merged; existing `hooks` / `permissions` blocks survive, and entries already listed are kept (union, no duplicates). Re-running is idempotent.
- **Skipped when there's no `package.json`** — swift-library, Python and Perl repos have no `node_modules` to symlink, so the key would be noise. One predicate covers all three rather than a per-preset flag.
- **Never clobbers a broken file.** A `settings.json` that doesn't parse is left alone and `doctor` reports it as drift for a human to repair.
- New JS-only doctor check `Claude worktree settings` (`ok` / `optional-missing` / `drift`), mapped to the `ai` fix target and demotable to "intentionally declined" through the existing `aiSetup` flag in `.repo-tooling.json` — no new lockfile field.
- Dogfooded: this repo's own `.claude/settings.json` is the generator's output.
- Docs: a new section in `apps/docs/docs/guides/ai-setup.md`.

## The second half of the referenced report

The lint-staged preset here already passes `--no-errors-on-unmatched` to both biome invocations (`src/cli/generators/git.ts`), so the aborted-commit bug from rtorcato/js-common#171 doesn't reproduce from this scaffolder. No change needed.

## Verification

`pnpm verify` (biome check, typecheck, 846 tests, build) passes. New tests cover the upsert-preserves-other-keys case, the no-`package.json` skip, the unparseable-file refusal, and all four doctor statuses including the lockfile demotion.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,0 +1,7 @@
+{
+  "worktree": {
+    "symlinkDirectories": [
+      "node_modules"
+    ]
+  }
+}
--- a/apps/docs/docs/guides/ai-setup.md
+++ b/apps/docs/docs/guides/ai-setup.md
@@ -20,6 +20,7 @@ All of it derives from **one source of truth** — the shipped Claude skill
 | `.github/copilot-instructions.md` | GitHub Copilot | Merge-safe delimited block |
 | `.claude/skills/repo-tooling.md` | Claude Code skill | Copied verbatim |
 | `.mcp.json.example` | Model Context Protocol | Commented template (see below) |
+| `.claude/settings.json` | Claude Code worktrees | Merge-safe key upsert (see below), JS repos only |
 | `README.md` | Your repo's own skills | Merge-safe block, only if this repo ships `skills/<name>/SKILL.md` |
 
 Every file is either a merge-safe **delimited block** (`<!-- js-tooling:start -->`
@@ -69,6 +70,30 @@ auto-writes this same install section into your `README.md` — one
 `npx skills add` command per skill, with the GitHub URL derived from
 `package.json`'s `repository`. It's a merge-safe delimited block, so your own
 README content is never touched, and repos without a `skills/` dir get nothing.
+
+## Worktrees: symlink `node_modules` instead of reinstalling it
+
+A Claude Code worktree starts empty, so an agent working one pays a full
+`pnpm install` before it can typecheck, lint or test — every time. `fix ai`
+upserts this into `.claude/settings.json` so Claude symlinks the directory from
+the main checkout instead:
+
+```json
+{
+  "worktree": {
+    "symlinkDirectories": ["node_modules"]
+  }
+}
+```
+
+Only the `worktree.symlinkDirectories` key is touched — your `hooks`,
+`permissions` and anything else in that file survive, and entries you added
+yourself are kept. A file that doesn't parse is left alone rather than
+clobbered; `doctor` reports it as drift for you to repair by hand.
+
+Skipped for Swift, Python and Perl repos — no `node_modules` to symlink, so the
+key would be noise. `doctor` reports `Claude worktree settings`, and it honours
+the same `aiSetup` opt-out in `.repo-tooling.json` as the rest of this feature.
 
 ## CLAUDE.md is a pointer, not a copy
 
--- a/src/base/fixers.ts
+++ b/src/base/fixers.ts
@@ -195,7 +195,7 @@ export const BASE_FIXERS: Fixer[] = [
 		target: 'ai',
 		description:
 			'Install all AI agent files at once (AGENTS.md, CLAUDE.md, Cursor, Copilot, Claude skill, MCP example)',
-		appliesTo: ['AI setup'],
+		appliesTo: ['AI setup', 'Claude worktree settings'],
 		outputs: [
 			'AGENTS.md',
 			'CLAUDE.md',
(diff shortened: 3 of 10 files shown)


rtorcato (commented):

**Verdict: clean — no criticals, highs, or mediums.**

Reviewed for injection, secrets, unsafe shell/SQL, dependency/supply-chain risk, and org-specific data leaking into scaffolded output.

- No shell/`child_process`/`eval` usage introduced; no SQL; no new dependencies.
- No secrets or credentials in the diff.
- The scaffolded `.claude/settings.json` payload is purely generic — `{"worktree":{"symlinkDirectories":["node_modules"]}}` — nothing `@rtorcato`-specific or otherwise org-bound gets baked into consuming repos.
- Scope matches the issue exactly: only the `worktree.symlinkDirectories` key is upserted (union, dedup on `node_modules`); `hooks`/`permissions` and any other existing keys are left untouched. No broader permissions are granted than #396 asked for.
- Malformed `settings.json` is detected via try/catch around `fs.readJson` and left alone rather than clobbered (`installClaudeSettings` returns `null`) — avoids destroying hand-written hooks/permissions on a parse failure.
- `package.json`-presence gate uses a literal joined path, no traversal/injection surface; `targetDir` is the same locally-trusted CLI target every other generator already operates on.
- Doctor check and lockfile demotion (`aiSetup: false`) are read-only reporting, no side effects.

No findings to block on.

rtorcato (commented):

Reviewed the diff against issue #396. Implementation matches the ask and is internally consistent:

- **Upsert logic** (`installClaudeSettings` in `src/cli/generators/agent-rules.ts`) only merges `worktree.symlinkDirectories`, preserves other `worktree` keys (e.g. `copyFiles`) and top-level keys (`hooks`), and is idempotent (dedupes via `existing.includes`).
- **Malformed-file refusal**: when `.claude/settings.json` exists but doesn't parse, `readClaudeSettings` returns `null`, so `installClaudeSettings` returns `null` and leaves the file untouched — matches "never clobbers a broken file."
- **package.json gating**: generalizing the swift/python/perl skip to "no package.json" is a reasonable and correct broadening of the issue's swift-library-only ask — verified `swiftFileList` (untouched by this PR) never lists `.claude/settings.json`, and `computeFileList` returns early for `config.language === 'swift'` before reaching the new push, so the JS-only addition there is safe.
- **Doctor check placement**: `checkClaudeWorktreeSettings` is called only in the JS-suite branch of `runDoctor` (after the swift/python/perl/unsupported branches all early-return), so it's correctly JS-only despite having no internal package.json check of its own.
- **Lockfile demotion**: `Claude worktree settings` → `ai` in `FIX_TARGETS` and the added case in `declinedInLock` follow the exact existing pattern used for `AI setup`.
- Tests cover the upsert-preserves-other-keys case, no-package.json skip, unparseable-file refusal, and all doctor statuses including lockfile demotion — assertions match the implementation.

Nit only: `tests/cli/commands/doctor.test.ts` has a stray extra blank line between the new `Claude worktree settings: drift...` test and the following `it('detects lint-staged...')` test. Cosmetic, not blocking.

No bugs or convention violations found.

PR #408, opened 2026-08-12
fix(ai): symlink nested workspace node_modules into agent worktrees

### LINKED ISSUE #406: WORKTREE_SYMLINK_DIRS misses nested workspace node_modules, so agent worktrees can't run workspace builds
_(AI-authored issue.)_

`WORKTREE_SYMLINK_DIRS` in `src/cli/generators/agent-rules.ts` is:

```ts
export const WORKTREE_SYMLINK_DIRS = ['node_modules'] as const
```

Only the **root** `node_modules` is symlinked into a fresh agent worktree. In a
workspace repo, every nested `node_modules` is absent, so any workspace-scoped
command fails in a worktree:

```
$ pnpm --filter @rtorcato/js-common-docs build
docusaurus: command not found
```

## Why this bites harder than a missing binary

On `rtorcato/js-common` the docs build is **the only check that catches broken
documentation links** — Docusaurus fails the build on them. An agent working on
docs in a per-issue worktree therefore cannot run the one verification that
matters for its change, and the failure surfaces as a missing binary rather than
anything reading as "you are missing a link check". So it gets skipped, not
fixed.

Hit repeatedly on js-common; each time the workaround was a manual
`ln -s /path/to/main/apps/docs/node_modules apps/docs/node_modules`.

## Current state

js-common now carries the nested path as a **local** addition
(rtorcato/js-common#193, merged as #196):

```json
"worktree": { "symlinkDirectories": ["node_modules", "apps/docs/node_modules"] }
```

That survives `repo-tooling fix` — `installClaudeSettings` merges additively
(`...existing`) rather than replacing — so nothing is broken today. But it is a
per-repo hand-edit of a file this package owns, which is exactly the pattern
`doctor` exists to discourage, and every sibling repo with a workspace needs the
same edit made by hand.

Same argument as rtorcato/js-common#191: fix it upstream instead of copying the
workaround into a seventh sibling repo.

## Suggested approach

Derive the list rather than hardcoding a second literal. On install, read the
workspace globs (`pnpm-workspace.yaml`, or `package.json` `workspaces`), resolve
them, and add `<pkg>/node_modules` for each workspace that exists. That keeps it
correct for repos with different layouts instead of special-casing `apps/docs`.

Worth confirming before building:

- Whether Claude Code's `symlinkDirectories` tolerates a path whose parent does
  not yet exist at worktree-creation time. (js-common#193 verified nested paths
  *are* supported, that entries are joined onto both roots, and that a bad entry
  warns and continues rather than breaking worktree creation — so the blast
  radius of getting this wrong is low.)
- Whether to symlink workspaces that have no `node_modules` in the main checkout
  yet. Probably skip silently.

## Related caution worth documenting wherever this lands

Do **not** run `pnpm install` inside a worktree whose `node_modules` is a
symlink. pnpm writes through the link and re-points the *shared* root `.bin`
shims at that worktree's virtual store, which then breaks every other worktree
and the main checkout with a doubled path and a misleading
`tsc: MODULE_NOT_FOUND`. Recovery is `pnpm install --frozen-lockfile` from the
main checkout. Observed twice in one session on js-common; both times it was
misdiagnosed as the symlinking being broken, when the in-worktree install was the
cause.

rtorcato:
_(AI-authored PR — implementer agent via ai-issue-loop, posted under @rtorcato's account.)_

Only the root `node_modules` was symlinked into a fresh Claude Code worktree, so any workspace-scoped command failed there on a missing binary (`pnpm --filter <pkg> build` → `command not found`). That reads as a broken toolchain rather than a missing symlink, so the check it was meant to run gets skipped, not fixed.

## What changed

- `workspaceSymlinkDirs()` in `src/cli/generators/agent-rules.ts` derives `<workspace>/node_modules` from the consuming repo's **own** workspace globs — `pnpm-workspace.yaml` `packages:` (reusing the existing `parseWorkspacePackages`, now exported) and/or `package.json` `workspaces` (both array and `{ packages: [] }` forms). No layout is hardcoded.
- `installClaudeSettings` merges those alongside `node_modules`; still additive, so hand-added entries survive and re-running does not duplicate.
- `checkClaudeWorktreeSettings` reports the same derived list, so a repo carrying only the root entry is nudged instead of reading as ok.
- Docs: nested entries + a warning never to run `pnpm install` inside a worktree whose `node_modules` is a symlink (it re-points the shared root `.bin` shims and breaks every other worktree; recover with `pnpm install --frozen-lockfile` from the main checkout).
- Dogfood: this repo's own `.claude/settings.json` gains `apps/docs/node_modules`.

Workspaces with no `node_modules` in the main checkout are skipped — an entry pointing at nothing warns on every worktree creation, and skipping also keeps `doctor` from flapping on an uninstalled checkout.

## Verification

`biome check`, `tsc --noEmit`, full `vitest run` (882 passed / 18 skipped). New tests cover derivation from both glob sources, the skip, idempotency, and the doctor nudge.

Closes #406

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,7 +1,8 @@
 {
   "worktree": {
     "symlinkDirectories": [
-      "node_modules"
+      "node_modules",
+      "apps/docs/node_modules"
     ]
   }
 }
--- a/apps/docs/docs/guides/ai-setup.md
+++ b/apps/docs/docs/guides/ai-setup.md
@@ -86,10 +86,32 @@ the main checkout instead:
 }
 ```
 
+In a workspace repo the nested `node_modules` are added too, so
+`pnpm --filter <pkg> build` works in a worktree instead of failing on a missing
+binary. The list is derived from your own workspace globs — `pnpm-workspace.yaml`
+`packages:` or `package.json` `workspaces` — and only workspaces that actually
+have a `node_modules` in the main checkout are listed:
+
+```json
+{
+  "worktree": {
+    "symlinkDirectories": ["node_modules", "apps/docs/node_modules"]
+  }
+}
+```
+
 Only the `worktree.symlinkDirectories` key is touched — your `hooks`,
 `permissions` and anything else in that file survive, and entries you added
 yourself are kept. A file that doesn't parse is left alone rather than
 clobbered; `doctor` reports it as drift for you to repair by hand.
+
+:::warning Never run `pnpm install` inside a worktree
+The worktree's `node_modules` is a symlink, so pnpm writes *through* it and
+re-points the shared root `.bin` shims at that worktree's virtual store —
+breaking every other worktree and the main checkout with a misleading
+`tsc: MODULE_NOT_FOUND`. Recover with `pnpm install --frozen-lockfile` from the
+main checkout.
+:::
 
 Skipped for Swift, Python and Perl repos — no `node_modules` to symlink, so the
 key would be noise. `doctor` reports `Claude worktree settings`, and it honours
--- a/src/cli/generators/agent-rules.ts
+++ b/src/cli/generators/agent-rules.ts
@@ -50,9 +50,24 @@ function asObject(value: unknown): Record<string, unknown> | null {
 }
 
 /**
+ * True when `rel` resolves to something at or under `targetDir`. The entries
+ * this guards end up in `worktree.symlinkDirectories`, which Claude Code reads
+ * as symlink targets — an escaping path would point a worktree symlink outside
+ * the repo, so a workspace glob is never trusted to stay inside on its own.
+ */
+function isInside(targetDir: string, rel: string): boolean {
+	const outward = path.relative(path.resolve(targetDir), path.resolve(targetDir, rel))
+	return outward !== '' && !outward.startsWith('..') && !path.isAbsolute(outward)
+}
+
+/**
  * The repo's own workspace globs, from `pnpm-workspace.yaml` and/or
  * package.json `workspaces` (both array and `{ packages: [] }` forms). Both
  * are read because a pnpm repo can keep its globs in either file.
+ *
+ * Negations (`!apps/x`) are dropped, and so is anything absolute or containing
+ * a `..` segment: a `packages: - '../shared-lib'` entry reads as plausible in a
+ * monorepo but would write an out-of-repo symlink target into settings.json.
  */
 async function workspaceGlobs(targetDir: string): Promise<string[]> {
 	const yamlFile = path.join(targetDir, 'pnpm-workspace.yaml')
@@ -63,7 +78,11 @@ async function workspaceGlobs(targetDir: string): Promise<string[]> {
 		: asObject(pkg?.workspaces)?.packages
 	const declared = Array.isArray(field) ? field : []
 	return [...parseWorkspacePackages(yaml), ...declared].filter(
-		(g): g is string => typeof g === 'string' && !g.startsWith('!')
+		(g): g is string =>
+			typeof g === 'string' &&
+			!g.startsWith('!') &&
+			!path.isAbsolute(g) &&
+			!g.split(/[\\/]/).includes('..')
 	)
 }
 
@@ -86,7 +105,9 @@ export async function workspaceSymlinkDirs(targetDir: string): Promise<string[]>
 		globs.map((g) => `${g}/node_modules`),
 		{ cwd: targetDir }
 	)) {
-		found.push(match.split(path.sep).join('/'))
+		// Belt and braces: the glob result is what actually gets written, so it is
+		// re-checked for containment even though the input globs were filtered.
+		if (isInside(targetDir, match)) found.push(match.split(path.sep).join('/'))
 	}
 	return found.sort()
 }
--- a/src/cli/generators/agent-rules.ts
+++ b/src/cli/generators/agent-rules.ts
@@ -1,7 +1,9 @@
+import { glob } from 'node:fs/promises'
 import path from 'node:path'
 import fs from 'fs-extra'
 import { copyPreset, getPackageRoot } from '../utils/copy-preset.js'
 import { LEGACY_TOOL_NAME } from '../utils/lockfile.js'
+import { parseWorkspacePackages } from './misc.js'
 import { installSkillsInstallDocs } from './skills-install.js'
 
 /**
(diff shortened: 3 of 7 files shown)


rtorcato (commented):
🤖 *Automated review — `code-reviewer` via ai-issue-loop. Posted under the owner's account by an agent; not a human review. There is no separate GitHub account for AI agents, so this appears under @rtorcato's avatar.*

Solid fix for #406. Checked the specific concerns flagged for this one:

- **No hardcoded layout.** `workspaceGlobs`/`workspaceSymlinkDirs` derive entirely from the consuming repo's own `pnpm-workspace.yaml` `packages:` and/or `package.json` `workspaces`, resolved with a real glob against the checkout. `apps/docs/node_modules` only shows up because this repo's own settings.json is dogfooded, not because it's baked into the generator.
- **`symlinkDirectories` parent-existence question.** Already answered in the issue itself (js-common#193 verified nested paths are supported and a bad entry just warns), so nothing further needed here — correctly not re-litigated in the implementation.
- **`installClaudeSettings` additive merge.** Still `[...existing, ...wanted.filter(not in existing)]`, so hand-added entries survive and a second run doesn't duplicate (covered by the "without duplicating" test).
- Workspaces with no `node_modules` yet are silently skipped via the glob itself (no match = no entry), matching the issue's suggested approach.

Two things worth a human's judgment, both narrow:

1. `workspaceSymlinkDirs` uses `node:fs/promises` `glob` (stable only in fairly recent Node; experimental with a stderr warning in some 22.x releases). Worth confirming it's within this package's supported `engines` range and doesn't print an experimental-warning during normal CLI/doctor runs — I didn't have the file in scope to check.
2. `workspaceGlobs` filters out `!`-prefixed negation globs from `pnpm-workspace.yaml` entirely (`!g.startsWith('!')`) rather than applying them as excludes against the positive matches. So a workspace deliberately excluded via `- '!apps/excluded'` alongside `- 'apps/*'` would still get a symlink entry if it happens to have a `node_modules` folder. Low blast radius (extra unused symlink, not breakage) and a narrow case, but not what the negation syntax implies.

### Before merging
- Confirm `node:fs/promises` `glob` is compatible with this package's `engines` Node range and doesn't emit an experimental-feature warning on supported versions.

rtorcato (commented):
🤖 *Automated review — `security-expert` via ai-issue-loop. Posted under the owner's account by an agent; not a human review. There is no separate GitHub account for AI agents, so this appears under @rtorcato's avatar.*

**Summary:** 0 critical, 1 medium, 0 low/info — no secrets, no shell/SQL injection, no dependency changes.

## Medium: no path-traversal containment check on workspace-glob-derived paths

`src/cli/generators/agent-rules.ts`: `workspaceGlobs()` reads glob entries from the repo's own `pnpm-workspace.yaml` (`parseWorkspacePackages`) and `package.json` `workspaces`, filtering out only entries starting with `!` (negation). It does not reject entries that are absolute (`/...`) or contain `..` segments. `workspaceSymlinkDirs()` then feeds those globs straight into `node:fs/promises` `glob({ cwd: targetDir })` and pushes every match into the list that ends up in `.claude/settings.json` → `worktree.symlinkDirectories`, with no check that the resolved match stays under `targetDir`.

Per issue #406 itself, Claude Code's worktree feature "joins entries onto both roots" when creating a worktree — i.e. this list is consumed downstream as symlink targets. A `pnpm-workspace.yaml`/`package.json` entry like `../../something` (or an absolute path) would produce an out-of-repo relative path written into a config file whose whole purpose is to drive symlink creation, and nothing here stops it from resolving outside the repo root.

**Exploit scenario:** an external contributor's PR adds an innocuous-looking `packages: - '../shared-lib'` (or similar) to `pnpm-workspace.yaml` — plausible-looking in a monorepo, easy to miss in review. Once merged, the next `repo-tooling fix`/`doctor` run silently writes that traversal path into `.claude/settings.json`, and the next worktree created picks it up as a symlink target outside the repo.

**Fix:** in `workspaceGlobs` (or right before pushing in `workspaceSymlinkDirs`), reject glob entries that are absolute or contain a `..` path segment, and/or verify `path.resolve(targetDir, match)` still starts with `path.resolve(targetDir)` before including a match.

No other issues: the rest of the diff (additive settings merge, doctor check, docs, tests) is straightforward and doesn't introduce injection, secret-leakage, or unsafe shell construction. No dependency/supply-chain changes.

### Before merging
- Add a containment/traversal check on workspace-glob-derived paths in `workspaceGlobs`/`workspaceSymlinkDirs` (`src/cli/generators/agent-rules.ts`) before they're written into `.claude/settings.json`, since that file drives symlink creation for worktrees.

rtorcato:
🤖 *Automated fix round — posted under the owner's account by an agent, not a human comment. There is no separate GitHub account for AI agents, so this appears under @rtorcato's avatar.*

Both review findings addressed in ed64126.

**Blocking (security-expert, medium) — path traversal.** Fixed with both guards:

1. `workspaceGlobs()` now also drops entries that are absolute (`path.isAbsolute`) or contain a `..` path segment, alongside the existing `!` negation filter.
2. `workspaceSymlinkDirs()` re-checks every glob *match* with a new `isInside()` helper (`path.relative` between the resolved target dir and the resolved match) before pushing it. The match is what actually gets written into `worktree.symlinkDirectories`, so it isn't trusted on the strength of the input filter alone.

New test: *"never writes a path outside the repo, however the workspace glob is spelled"* — a fixture with a real sibling `../shared-lib/node_modules`, referenced both as a `..` entry in `package.json` `workspaces` and as an absolute entry in `pnpm-workspace.yaml`, asserts the written list is exactly `['node_modules', 'packages/core/node_modules']`. Verified meaningful: replaying the pre-fix filter against the same fixture matches all three, so the test fails if either guard regresses.

**Secondary (code-reviewer) — `node:fs/promises` `glob` and `engines`.** No change needed, and no `engines` bump. `fs.promises.glob` landed in Node **v22.0.0**; this package declares `"node": ">=22"`, so the entire supported range has it — there is no compat break for consumers. It also emits no experimental warning: checked stderr on v22.23.1 and v24.18.0, both empty. (It's documented as experimental, but `fs.glob` never called `emitExperimentalWarning`, so nothing prints during CLI/doctor runs.)

Not touched, deliberately: the negation-glob semantics the code reviewer raised as item 2 (`!apps/excluded` dropped rather than applied as an exclude). It's non-blocking, low blast radius, and out of scope for this fix — worth its own issue if wanted. `tooling/claude/repo-tooling.md` also left alone; a sibling PR owns it.

Full suite green: 901 tests, 57 files. Typecheck and biome clean (pre-commit + pre-push hooks both passed).

rtorcato (commented):
🤖 *Automated review — `code-reviewer` via ai-issue-loop. Posted under the owner's account by an agent; not a human review. There is no separate GitHub account for AI agents, so this appears under @rtorcato's avatar.*

Reviewed at `ed64126`. Both items from the prior fix round hold.

**1. Path-traversal containment — confirmed fixed.** `workspaceGlobs()` filters out `!`-negated, absolute, and `..`-segment entries before they ever become a glob pattern. The actual glob matches (from `node:fs/promises` `glob(..., { cwd: targetDir })`) are independently re-checked by `isInside()` before being pushed into `found` — verified the relative-path math (`path.relative` + `startsWith('..')` + `isAbsolute`) correctly rejects both `../`-style and absolute escapes and correctly accepts legitimate nested paths like `apps/docs/node_modules`. Belt-and-braces as the comment says: even if a future caller fed `workspaceSymlinkDirs` an unfiltered glob, the output check still holds the line.

**2. `node:fs/promises` glob / engines compat — confirmed non-issue.** `package.json` already declares `"engines": { "node": ">=22" }` (unchanged by this PR), and `fsPromises.glob` runs clean on Node 22.23.1 with no experimental-API warning. No engines bump was made or needed — correct call, no semver implication.

Other things checked and fine: no `@rtorcato`-specific paths baked in (the repo's own dogfooded `apps/docs/node_modules` entry is a derived result, not a hardcoded default); `installClaudeSettings`/`checkClaudeWorktreeSettings` stay additive and idempotent; duplicate glob patterns don't produce duplicate matches (verified `node:fs/promises` glob dedupes internally); tests cover both workspace-glob sources, the skip-when-no-node_modules case, idempotency, and the traversal-safety case at the input-filter level.

### Before merging
Nothing.

rtorcato (commented):
🤖 *Automated review — `security-expert` via ai-issue-loop. Posted under the owner's account by an agent; not a human review. There is no separate GitHub account for AI agents, so this appears under @rtorcato's avatar.*

Re-review of commit `ed64126` against the medium finding from the prior round (`workspaceGlobs()` / `workspaceSymlinkDirs()` writing an escaping path into `.claude/settings.json` → `worktree.symlinkDirectories`).

**The fix holds under adversarial review.** Two layers now guard the write:

1. `workspaceGlobs()` drops `!`-prefixed negations, `path.isAbsolute()` entries, and any glob with a literal `..` segment (split on both `/` and `\`, not a naive substring check — so `..\\foo` is caught even though it isn't a POSIX separator).
2. `workspaceSymlinkDirs()` re-checks every actual glob *match* (not just the input pattern) with `isInside()`, which uses `path.relative(path.resolve(root), path.resolve(root, rel))` and rejects anything starting with `..`, anything absolute, and the empty-string/equal-path case. This is the correct pattern — it avoids the classic `resolved.startsWith(root)` bug (which a sibling directory like `/repo-evil` would defeat); `path.relative` handles that correctly.

I tried to defeat both layers:

- **Symlinked intermediate directory escaping containment** (e.g. `packages/evil -> /outside`, with `/outside/node_modules` existing, matched via the `packages/*/node_modules` pattern): does not reproduce. `workspaceSymlinkDirs()` uses `node:fs/promises` `glob()` with no `followSymlinks` option, and that API's documented default behavior is to **not traverse symbolic links that point to directories** when continuing a pattern (confirmed against the upstream Node issue that later added an opt-in `followSymlinks` flag). Since every constructed pattern here is `${glob}/node_modules` — always one segment past any wildcard — matching requires traversing through the wildcard-matched directory, and Node's glob won't follow a symlink to do that. So a committed symlink can't be used to smuggle an escaping match into the result the way it could with a third-party glob library.
- **Windows drive-absolute paths** (`C:\...`) bypassing `path.isAbsolute()` on a POSIX host: true in principle, but inert — POSIX treats backslash as a literal character, so no real directory would exist to match, and on an actual Windows host `path.isAbsolute()` correctly flags it.
- **`..` embedded via wildcard expansion, case-folding on case-insensitive filesystems, equal-path edge case**: none produce a bypass; the split-based segment check and `path.relative`-based containment check hold up.
- **Every path that reaches the settings file**: `workspaceSymlinkDirs()` output goes through both layers in both call sites (`installClaudeSettings` write path and `checkClaudeWorktreeSettings` read-only doctor check). Pre-existing hand-added entries in `existing` (already in `.claude/settings.json` before this code runs) are preserved unfiltered by design ("hand-added entries survive") — that's a separate, pre-existing trust boundary (whoever can already hand-edit that file has an equivalent primitive) and not a regression introduced by this diff.

No injection, secrets, unsafe shell construction, or dependency changes in this diff — the only new import is the built-in `node:fs/promises` `glob`.

### Before merging
Nothing.

PR #409, opened 2026-08-12
feat(skills): ship the ai-issue-loop skill with a user-level install target

### LINKED ISSUE #404: Ship the ai-issue-loop skill from repo-tooling, installed to a resolved user-level destination
_(AI-authored issue.)_

The `ai-issue-loop` skill currently lives only in the dotfiles repo (GitLab),
symlinked into `~/.claude/skills/ai-issue-loop/SKILL.md` by stow. repo-tooling
already owns everything that skill depends on, and already ships a *different,
drifted* description of the same pipeline (see the companion issue on
`tooling/claude/repo-tooling.md`).

## Why repo-tooling is the right home

It already owns the loop's entire repo-side contract:

- `GITHUB_STANDARD` in `src/base/github-settings.ts` sets
  `required_pull_request_reviews: null` — with a comment explaining it deadlocks
  auto-merge, which is *precisely* the constraint the loop is built around — plus
  squash / auto-merge / delete-on-merge.
- `installClaudeSettings` in `src/cli/generators/agent-rules.ts` owns
  `.claude/settings.json` `worktree.symlinkDirectories`, which the loop relies on
  for per-issue worktrees.
- `skills/<name>/SKILL.md` + `npx skills add` distribution already exists
  (`skills/repo-tooling`, `skills/npm-publish`).

So no new plumbing is needed. The skill becomes
`skills/ai-issue-loop/SKILL.md`, installable with:

```bash
npx skills add https://github.com/rtorcato/repo-tooling --skill ai-issue-loop
```

## The requirement that shapes the design

The skill must **still end up in dotfiles** for anyone who has one — running
`repo-tooling` on a new project or a new machine should land it there, so it
stays version-controlled with the rest of the Claude config. A developer with no
dotfiles repo should simply be asked where to put it.

dotfiles therefore holds a **managed copy, not a fork** — the same contract as
`biome.json` or `tsconfig`: repo-tooling owns the content, `doctor` reports
drift, `fix` updates it.

## Install destination resolution

In order:

1. `--skills-dir <path>` — explicit. **Required** under `--json` / `--yes`, since
   a prompt would corrupt the JSON (existing CLI rule).
2. `~/.claude/skills/<name>/SKILL.md` exists **and is a symlink** → write
   *through* it, so the content lands in dotfiles. Report the resolved real path
   so the user knows to commit it there.
3. `~/.claude/skills/` exists → write directly.
4. Nothing found → prompt for a location.

### Write through the symlink — do not replace it

This is the failure mode most likely to be shipped by accident. Today:

```
~/.claude/skills/            real dir
~/.claude/skills/ai-issue-loop/   real dir
~/.claude/skills/ai-issue-loop/SKILL.md   SYMLINK -> ../../../dotfiles/.../SKILL.md
```

stow symlinks at **file** level. Plain `fs.writeFile` follows the symlink and
updates dotfiles in place. Atomic-rename helpers (`write-file-atomic` and
friends) replace the symlink with a real file, leaving an orphaned dotfiles copy
and **no error** — recreating exactly the split-brain this issue exists to end.

`installClaudeSettings` already uses `fs.outputJson`, which follows symlinks.
Match that.

### 1. Version fighting (the significant one)

Every other fixer is repo-local. This one writes **user-global** state. A repo
pinned to 3.4.0 and another on 3.9.1 will overwrite each other's skill on every
`fix`, and neither is wrong to do so.

Suggested: embed the shipping version in the installed frontmatter, and refuse to
downgrade — write only when shipped > installed, otherwise report and skip.

### 2. Blanket `fix --yes` must not sweep this in

Silently overwriting a developer's global agent skills because they ran a repo
audit is a bad surprise, and violates the drift policy's spirit ("don't surprise
the user"). This should be an explicit `fix claude-skills` target, opt-in — not
part of `fix` all.

### 3. `doctor` needs a sensible verdict when nothing is installed

Probably `optional-missing` rather than `missing` — the loop is a workflow
choice, not a repo requirement, and `optional-missing` already means "opt-in tool
not configured; only fix if the user wants it".

## Also in scope: the guide

A single document covering the whole pipeline end to end, on the docs site
(`apps/docs`), since the contract is public and multiple repos participate:

1. The one constraint — no bot account, so approval is a label and required
   status checks are the real gate
2. Labels and the state machine (issue labels vs PR labels; `ai-changes` is the
   send-back, never `ai-ready`)
3. Repo prerequisites → `repo-tooling fix github-settings`
4. The tick, Pass 0–5
5. Driving it → `/loop 15m /ai-issue-loop`
6. Failure modes and limits — stall reaping, fix-round cap, why `ai-notes` exists

## Open question

Whether the macOS-specific bits (`osascript` notifications, the
`.claude/ai-loop-status` statusline file) ship as-is, get guarded by a platform
check, or move to a dotfiles-local overlay. They currently degrade with
`|| true`, so shipping them is harmless but untidy for a Linux consumer.

rtorcato:
🤖 *AI-authored PR — implementer via ai-issue-loop. Opened under the owner's account by an agent; not a human message. There is no separate GitHub account for AI agents, so this appears under @rtorcato's avatar.*

The `ai-issue-loop` skill lived only in the dotfiles repo, while repo-tooling
already owned every repo-side piece it depends on — `GITHUB_STANDARD`'s
`required_pull_request_reviews: null`, the `.claude/settings.json` worktree
config, and the `skills/<name>/SKILL.md` distribution format.

## What lands

- `skills/ai-issue-loop/SKILL.md` — installable with
  `npx skills add https://github.com/rtorcato/repo-tooling --skill ai-issue-loop`,
  and now included in the npm tarball (`files`).
- `fix claude-skills` — installs to a resolved destination: `--skills-dir` if
  given, else `~/.claude/skills` if it exists, else a prompt.
- `Claude skills` doctor check.
- `apps/docs/docs/guides/ai-issue-loop.md` — the end-to-end pipeline guide.

## The three properties the design turns on

**Writes *through* a stow symlink rather than replacing it**, so the content
lands in dotfiles and stays version-controlled. Plain `fs.writeFile` follows the
link; an atomic-rename helper (`write-file-atomic` and friends) would replace it
with a real file and orphan the dotfiles copy **with no error at all** — the
split-brain the issue exists to end. There is a test that asserts the symlink
survives and the dotfiles copy is the one updated.

**Stamps `repo-tooling-version` into the installed frontmatter and refuses to
downgrade.** Without it, a repo pinned to 3.4.0 and another on 3.9.1 overwrite
each other's global skill on every `fix`, and neither is wrong to do so.

**`explicitOnly` keeps the fixer out of `fix` and `fix --yes`.** It is the only
fixer whose blast radius is outside the repo, and silently rewriting a
developer's global agent skills because they ran a repo audit is exactly the
surprise the drift policy promises not to spring. A blanket `fix --yes` records
it as `skipped` and says to run it by name.

## Why `optional-missing` and never `drift`

`Claude skills` is the only check reporting on state outside the repo being
audited. An out-of-date `~/.claude/skills` is not a defect in *this* repo, and
`drift`/`missing` would fail its CI over a file CI has never seen.
`optional-missing` is the honest verdict and leaves the exit code alone.

## Open question from the issue, resolved

The macOS-specific bits ship as-is. `osascript` already degrades via `|| true`;
the skill now says so explicitly and gives the `notify-send` swap for Linux. A
platform check would be machinery for a line that is already a no-op elsewhere.

## Verification

`pnpm verify` green (910 tests, 58 files). Beyond the unit tests, the built CLI
was driven against a scratch consumer to confirm end to end: `fix --yes` skips
the target, a targeted run installs and is idempotent, a symlinked destination
is written through and stays a symlink, and a `999.0.0` copy is refused as a
downgrade.

Closes #404

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -1,6 +1,6 @@
 {
   "name": "repo-tooling",
-  "description": "AI skills for the @rtorcato/repo-tooling family: drive the tooling CLI and follow the shared semantic-release publishing rules.",
+  "description": "AI skills for the @rtorcato/repo-tooling family: drive the tooling CLI, follow the shared semantic-release publishing rules, and run the label-driven issue → PR pipeline.",
   "owner": {
     "name": "Richard Torcato",
     "url": "https://github.com/rtorcato"
@@ -9,7 +9,7 @@
     {
       "name": "repo-tooling",
       "source": "./",
-      "description": "Drive the @rtorcato/repo-tooling CLI (setup/doctor/fix) and follow the shared release rules. Two skills: repo-tooling (adopt/audit the presets) and npm-publish (never hand-cut a release)."
+      "description": "Drive the @rtorcato/repo-tooling CLI (setup/doctor/fix) and follow the shared release rules. Three skills: repo-tooling (adopt/audit the presets), npm-publish (never hand-cut a release) and ai-issue-loop (the label-driven issue → PR pipeline)."
     }
   ]
 }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "repo-tooling",
-  "description": "Adopt and audit the @rtorcato/repo-tooling shared presets via its CLI, and follow the family's semantic-release publishing rules.",
+  "description": "Adopt and audit the @rtorcato/repo-tooling shared presets via its CLI, follow the family's semantic-release publishing rules, and run the label-driven ai-issue-loop pipeline.",
   "version": "1.0.0",
   "author": {
     "name": "Richard Torcato",
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -82,11 +82,18 @@ npx @rtorcato/repo-tooling fix dependabot --yes --json
 npx @rtorcato/repo-tooling fix engines --yes --json
 npx @rtorcato/repo-tooling fix docs-site --yes --json   # scaffold a Docusaurus docs site under apps/docs
 npx @rtorcato/repo-tooling fix bun --yes --json         # Bun runtime/test config
+
+# Opt-in only — writes user-global state, so a bare `fix` / `fix --yes` skips it.
+# Installs the ai-issue-loop skill to ~/.claude/skills. Override with --skills-dir,
+# which is required alongside --yes/--json when that directory doesn't exist.
+npx @rtorcato/repo-tooling fix claude-skills --yes --json
 ```
 
 ## Drift policy (important)
 
 `fix` defaults the confirm prompt to **No** for drift cases (existing file that doesn't extend our preset). The `--yes` flag is required to overwrite drift. Safe-merge fixers (`engines`, `husky`, `package-json`) never overwrite — they add/merge — and use friendlier prompt wording. `fix --json` implies `--yes` (prompts would corrupt JSON output).
+
+Fixers marked `explicitOnly` are exempt from `fix` all *and* from `fix --yes` — they only run when named as the target. Today that is `claude-skills`, the one fixer whose blast radius is outside the repo.
 
 ## Source-of-truth files in the repo
 
--- a/README.md
+++ b/README.md
@@ -174,13 +174,24 @@ ln -sf ../../node_modules/@rtorcato/repo-tooling/tooling/claude/repo-tooling.md 
 ### Use with Claude Code (plugin)
 
 This repo is also a self-hosted Claude Code marketplace. Install the plugin to
-get two skills — `repo-tooling` (adopt/audit the presets via the CLI) and
-`npm-publish` (the family's release rules) — in any session:
+get three skills — `repo-tooling` (adopt/audit the presets via the CLI),
+`npm-publish` (the family's release rules) and `ai-issue-loop` (the label-driven
+issue → PR pipeline) — in any session:
 
 ```
 /plugin marketplace add rtorcato/repo-tooling
 /plugin install repo-tooling@repo-tooling
 ```
+
+`ai-issue-loop` also installs on its own, user-globally, so every repo on the
+machine shares one copy:
+
+```bash
+npx @rtorcato/repo-tooling fix claude-skills   # → ~/.claude/skills/ai-issue-loop/SKILL.md
+```
+
+It writes outside the repo, so it is opt-in: a bare `fix` skips it. See the
+[AI Issue Loop guide](https://rtorcato.github.io/repo-tooling/guides/ai-issue-loop/).
 
 ### Use with other AI tools (Cursor / Copilot / Codex)
 
@@ -217,7 +228,8 @@ MIT — see [LICENSE](LICENSE).
 Any agent that supports the [`skills`](https://www.npmjs.com/package/skills) CLI can install this repo's skills straight from GitHub — no clone, no package install:
 
 ```bash
-npx skills add https://github.com/rtorcato/repo-tooling --skill repo-tooling
+npx skills add https://github.com/rtorcato/repo-tooling --skill ai-issue-loop
 npx skills add https://github.com/rtorcato/repo-tooling --skill npm-publish
+npx skills add https://github.com/rtorcato/repo-tooling --skill repo-tooling
 ```
 <!-- js-tooling:skills:end -->
--- a/apps/docs/docs/guides/ai-issue-loop.md
+++ b/apps/docs/docs/guides/ai-issue-loop.md
@@ -1,0 +1,189 @@
+---
+title: The AI Issue Loop
+description: The end-to-end label-driven pipeline that turns an ai-ready GitHub issue into a reviewed PR — the one constraint that shapes it, the label state machine, the repo prerequisites, and the limits that keep it cheap.
+---
+
+`ai-issue-loop` is a **label-driven pipeline** that takes a GitHub issue marked
+`ai-ready`, implements it in a per-issue git worktree, opens a PR, has two agents
+review it, and hands it to you to merge. It runs unattended on a timer.
+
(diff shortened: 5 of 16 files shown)


rtorcato (commented):
🤖 *Automated review — `security-expert` via ai-issue-loop. Posted under the owner's account by an agent; not a human review. There is no separate GitHub account for AI agents, so this appears under @rtorcato's avatar.*

Reviewed the symlink-write, path-resolution, and overwrite behavior of the new `claude-skills` fixer (`src/cli/generators/claude-skills.ts`, wired through `src/base/fixers.ts`).

**Symlink following is bounded, unlike #408's gap.** The write target is `path.join(skillsDir, name, 'SKILL.md')` where `name` is the hardcoded constant `SHIPPED_SKILL = 'ai-issue-loop'` — never derived from a glob, CLI arg, or issue body. There is no dynamic segment for a hostile symlink target to redirect via `..` or an absolute value; the only variable is `skillsDir` itself, which controls the file's *directory*, not what it resolves through. If `<skillsDir>/ai-issue-loop/SKILL.md` is itself a symlink, `fs.writeFile` follows it (`installClaudeSkill` in `src/cli/generators/claude-skills.ts:1574-1602`), by design and tested (`tests/cli/generators/claude-skills.test.ts:1792`, "writes through a symlink instead of replacing it"). For that symlink to point somewhere hostile, an attacker needs pre-existing write access to the user's own `~/.claude/skills/` — at which point they already have equivalent-uid filesystem access and easier paths to compromise (`.bashrc`, `.ssh/authorized_keys`, etc. directly). Content written is always the package's own shipped skill markdown, never attacker-supplied. Not a privilege boundary this tool crosses.

**No path traversal in `--skills-dir`.** `resolveSkillsDir` (`src/cli/generators/claude-skills.ts:1477`) does `path.resolve(explicit)` on an explicit CLI flag typed directly by the human running the command — not sourced from an issue body, PR content, or any other untrusted input. This fixer is `explicitOnly` (`src/base/fixers.ts:1288`), so it never runs as part of a bare `fix`/`fix --yes` sweep and is outside the ai-issue-loop's issue-driven codepath entirely.

**Overwrite is guarded and bounded.** `installClaudeSkill` refuses to downgrade (`declined-downgrade` when the installed `repo-tooling-version` stamp is newer, `claude-skills.ts:1592`) and is a no-op when content is already current. Blast radius is a single fixed path; there's no loop over user-supplied names or globs comparable to #408.

**Shadow-run guard is correct.** `riskLevel: 'safe-add'` exempts this fixer from `fix --diff`'s shadow-run (which copies the repo to tmp and executes `run()`) — verified in `fix.ts`: `if (showDiff && (fixer.riskLevel ?? 'destructive') !== 'safe-add')`. Without that exemption a `--diff` preview would write to the real home directory, which the code comment correctly calls out.

No new dependency (`inquirer` was already in `package.json`). No secrets, no shell interpolation, no SQL. `--skills-dir` is process-local CLI input, not attacker-reachable.

### Before merging
Nothing.

rtorcato (commented):
🤖 *Automated review — `code-reviewer` via ai-issue-loop. Posted under the owner's account by an agent; not a human review. There is no separate GitHub account for AI agents, so this appears under @rtorcato's avatar.*

Reviewed against issue #404's install-resolution order, the write-through-symlink requirement, and the public-library no-org-specifics rule.

**Install resolution order** — `resolveSkillsDir` collapses the issue's steps 2+3 (symlink check vs. plain-dir check) into "does `~/.claude/skills` exist", then `installClaudeSkill` decides write-through-symlink vs. direct write based on whether the target `SKILL.md` file itself is a symlink (`src/cli/generators/claude-skills.ts:1477-1485`, `:1554-1601`). Functionally equivalent to the issue's four-step order and cleaner — no divergent case exists since a symlinked `SKILL.md` implies its parent dir already exists.

**Write-through-symlink** — correctly implemented and tested. `fs.writeFile` (not an atomic-rename helper) is used deliberately, with a comment explaining why, and `tests/cli/generators/claude-skills.test.ts` has a dedicated test (`writes through a symlink instead of replacing it`) asserting the link survives and the dotfiles-side file is what changes. This is the failure mode the issue called out as most likely to ship by accident, and it's handled correctly with a regression test guarding it.

**Version-downgrade guard and `explicitOnly`** — both match the issue's stated hazards (version fighting, blanket `fix --yes` sweep-in) and are exercised by tests.

**Semver** — `feat(skills):` is honest: new fixer, new doctor check, new CLI flag, new skill file, all additive with no breaking change visible in the diff.

Two things worth a second look, neither blocking:

1. `src/cli/generators/claude-skills.ts:1252-1269` (`resolveInstallDir`) — under `--yes`/`--json` with no resolvable `~/.claude/skills` and no `--skills-dir`, the fixer logs a `console.error` warning and silently returns `{filesWritten: []}` rather than failing. AGENTS.md and the new guide both say `--skills-dir` is "required" in that combination, but nothing enforces it — a `--yes --json` automation call would see a normal-looking (empty) success rather than a hard failure signaling the install didn't happen. Worth confirming this matches how other "required under --json" flags behave elsewhere in the CLI, or making it an explicit error.
2. `skills/ai-issue-loop/SKILL.md`'s "If you ever switch to real approvals" section hardcodes `@rtorcato/repo-tooling` and `src/base/github-settings.ts` by name. This skill is distributed generically (`npx skills add ... --skill ai-issue-loop`, installed to any consumer's `~/.claude/skills`), so that paragraph only applies to repos that actually depend on `@rtorcato/repo-tooling`. Harmless for everyone else, but it is repo-tooling-specific content baked into a file meant to travel to any machine/repo.

### Before merging
- `resolveInstallDir` should probably hard-fail (non-zero exit / explicit error in the JSON payload) rather than silently skip when `--skills-dir` is unresolvable under `--yes`/`--json`, since docs describe it as "required" in that case — confirm this is intentional or tighten it.

PR #434, opened 2026-08-20
feat(claude): ship sync-agents.mjs as a copy asset

### LINKED ISSUE #427: Ship sync-agents.mjs as a copy asset — three repos carry byte-identical copies
`scripts/sync-agents.mjs` regenerates `AGENTS.md` from `skills/<name>/SKILL.md` so the two can't drift, with a `--check` mode for CI. Three repos have it and repo-tooling ships no asset for it, so all three are hand-maintained copies.

## The copies are identical

`api-common`, `browser-common`, `js-common`. Diffed all three — **every difference is the package name substituted into a path or a log line.** The logic (strip YAML frontmatter, prepend the generated-by header, `--check` compares and exits 1) is byte-identical.

The complete `js-common` → `browser-common` diff, minus prose:

```diff
-const SKILL = join(root, 'skills/js-common/SKILL.md')
+const SKILL = join(root, 'skills/browser-common/SKILL.md')
```

Sizes: 1500 / 1510 / 1526 bytes. The spread is entirely the package name's length and a comment mentioning where `--check` runs (`CI` vs `verify` vs `check:agents`).

## It can be zero-config

The one variable is derivable, so the asset needs no per-repo edit at all:

```js
const { name } = JSON.parse(readFileSync(resolve(root, 'package.json'), 'utf8'))
const SKILL = join(root, 'skills', name.replace(/^@[^/]+\//, ''), 'SKILL.md')
```

`@rtorcato/js-common` → `skills/js-common/SKILL.md`. All three repos already follow that convention, so a copied asset would work unmodified in each — the same property that makes `docusaurus-sync-changelog` copyable.

Worth confirming against `api-common`, which has a workspace layout (`gen-readme-packages.mjs`, `aggregate-changelog.mjs` suggest multiple packages) — if its `skills/` dir isn't keyed on the root package name, the derivation needs a `SKILL_PATH` env override in the same spirit as `CHANGELOG_TARGET`.

## Shape

`tooling/claude/sync-agents.mjs`, copied via `repo-tooling copy claude-sync-agents` — matching `docusaurus-sync-changelog`. `tooling/claude/` currently holds only `repo-tooling.md`, so this would be its first script asset.

Once it lands, the three repos delete their copies and re-copy. I'll open adoption issues in each, held until then.

## Related

- #426 — extracting js-common's docs-generator helpers, same "promote a script into the shared package" shape
- The companion issue on detecting *stale* copied assets, which is what makes this worth centralising rather than just deduplicating once

rtorcato:

`api-common`, `browser-common` and `js-common` each carry a hand-maintained copy of `scripts/sync-agents.mjs`. Diffing them, every difference is the package name spliced into a path and two log lines — the logic is byte-identical. This ships one canonical copy.

## What's here

- `tooling/claude/sync-agents.mjs` — the asset, modelled on `tooling/docusaurus/sync-changelog.mjs`. First script asset under `tooling/claude/`.
- Registered as `claude-sync-agents` in `src/cli/utils/copy-preset.ts` → `scripts/sync-agents.mjs`.
- `package.json` `files` gains `tooling/claude/*.mjs` — the existing entry was `*.md` only, so without this the asset would not ship.
- `tests/cli/claude-helpers.test.ts` — copies the asset into a seeded tmp repo and **runs it with node**, asserting the generated `AGENTS.md` bytes, the unscoped-name path, and that `--check` exits 1 on a stale file.
- Docs: preset list + a short paragraph in `apps/docs/docs/guides/cli.md`.

## Zero-config, as the issue proposed

The skill path comes from the root `package.json` `name` with any npm scope stripped, so the same bytes work unmodified in each repo:

```js
const skillRelative = `skills/${name.replace(/^@[^/]+\//, '')}/SKILL.md`
```

Verified the derived header matches each repo's current `AGENTS.md` first line exactly:

| repo | `package.json` name | derived | current `AGENTS.md` header |
|---|---|---|---|
| js-common | `@rtorcato/js-common` | `skills/js-common/SKILL.md` | ✅ identical |
| browser-common | `@rtorcato/browser-common` | `skills/browser-common/SKILL.md` | ✅ identical |
| api-common | `api-common` (unscoped) | `skills/api-common/SKILL.md` | ✅ identical |

So all three can delete their copies and re-copy with no diff.

## The `SKILL_PATH` override — skipped

The issue flagged it as conditional on api-common's layout. Checked: `api-common/skills/api-common/` is keyed on the root package name like the others, and its unscoped name still derives correctly. Nothing needs the override, so it isn't there. Easy to add later if a consumer diverges.

## Scope

This repo only. No changes to api-common / browser-common / js-common — their adoption is tracked separately per the issue.

`pnpm check`, `pnpm typecheck` and the full suite (912 passed) are green locally.

Closes #427

### DIFF
--- a/apps/docs/docs/guides/cli.md
+++ b/apps/docs/docs/guides/cli.md
@@ -27,7 +27,9 @@ npx @rtorcato/repo-tooling copy biome     # → biome.json
 npx @rtorcato/repo-tooling copy tsconfig  # → tsconfig.json
 ```
 
-Available presets: `biome`, `tsconfig`, `bun`, `nx`, `changesets`, `release-please`, `oxlint`, `claude-skill`, `mcp-example`, `docusaurus-sync-changelog`, `docusaurus-theme-tokens`, `docusaurus-theme`.
+Available presets: `biome`, `tsconfig`, `bun`, `nx`, `changesets`, `release-please`, `oxlint`, `claude-skill`, `claude-sync-agents`, `mcp-example`, `docusaurus-sync-changelog`, `docusaurus-theme-tokens`, `docusaurus-theme`.
+
+`claude-sync-agents` copies `scripts/sync-agents.mjs`, which regenerates `AGENTS.md` from `skills/<package>/SKILL.md` so the two can't drift. It is zero-config — the skill directory comes from the root `package.json` `name` with the npm scope stripped (`@rtorcato/js-common` → `skills/js-common/`). Wire `node scripts/sync-agents.mjs --check` into CI to fail on a stale `AGENTS.md`.
 
 `copy` is for configs you must *own* rather than extend — Biome doesn't support configuration extension, and TypeScript configs resolve more reliably when local. Most other configs (ESLint, Prettier, Vitest, etc.) can be imported or extended directly — see the [Configuration Reference](../reference/biome.md) for usage.
 
--- a/package.json
+++ b/package.json
@@ -100,6 +100,7 @@
 		"tooling/semantic-release/*.mjs",
 		"tooling/semantic-release/*.d.mts",
 		"tooling/claude/*.md",
+		"tooling/claude/*.mjs",
 		"tooling/mcp/mcp.json.example",
 		"tooling/swift/*.yml",
 		"tooling/swift/*.json",
--- a/src/cli/utils/copy-preset.ts
+++ b/src/cli/utils/copy-preset.ts
@@ -18,6 +18,7 @@ export type PresetName =
 	| 'perlcritic'
 	| 'perltidy'
 	| 'claude-skill'
+	| 'claude-sync-agents'
 	| 'mcp-example'
 	| 'docusaurus-sync-changelog'
 	| 'docusaurus-theme-tokens'
@@ -112,6 +113,11 @@ export const PRESETS: Record<PresetName, PresetDefinition> = {
 		target: '.claude/skills/repo-tooling.md',
 		desc: 'Claude Code skill for driving the repo-tooling CLI',
 		legacyTarget: '.claude/skills/js-tooling.md',
+	},
+	'claude-sync-agents': {
+		source: 'tooling/claude/sync-agents.mjs',
+		target: 'scripts/sync-agents.mjs',
+		desc: 'Canonical SKILL.md → AGENTS.md sync script (zero-config; --check mode for CI)',
 	},
 	'mcp-example': {
 		source: 'tooling/mcp/mcp.json.example',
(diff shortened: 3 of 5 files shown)


rtorcato (commented):

Reviewed `tooling/claude/sync-agents.mjs`, its `copy-preset.ts` registration, the `package.json` `files` change, and the new test.

No injection risk: the script only does `fs.readFileSync`/`writeFileSync` on paths built from `package.json`'s `name` field — no `child_process`, `eval`, or template/shell construction anywhere. The new test invokes the script via `execFile(process.execPath, [path, ...args])`, an argv array with no shell, so no command-injection surface there either. No secrets, no new or bumped dependencies.

The only theoretical soft spot is that `name.replace(/^@[^/]+\//, '')` isn't defended against a `name` containing `../`, which could make `skillRelative` escape `root`. Not exploitable in practice: `name` comes from the consuming repo's own `package.json` (already trusted, npm restricts publishable name characters), and worst case is the script reading/writing a path elsewhere in the maintainer's own working tree — not a real trust boundary crossing.

### Before merging
Nothing.

rtorcato (commented):

Matches the issue's proposed shape closely: canonical `tooling/claude/sync-agents.mjs`, zero-config skill-path derivation (verified against all three consuming repos' actual `package.json` names in the PR description), registered as `claude-sync-agents`, and the `tooling/claude/*.mjs` glob is added to `files` (correctly noted as needed since the existing entry was `*.md`-only). Scope is properly limited to this repo per the issue's own "Scope" section — no changes to the three consumer repos.

The script logic itself (frontmatter strip via `/^---\n[\s\S]*?\n---\n+/`, `--check` exit-1-on-mismatch) is straightforward and correct. Test coverage runs the copied asset end-to-end with `node` against a seeded tmp repo, covering scoped name, unscoped name, and the `--check` pass/fail paths — good, matches the stated intent of testing behavior rather than just the copy mechanism.

One thing worth a second look: the generated `AGENTS.md` header hardcodes the remediation command as `` `pnpm sync:agents` ``. The issue itself notes the three existing copies differ in exactly this detail — their log/comment text references `CI`, `verify`, and `check:agents` respectively — implying at least one doesn't use a script literally named `sync:agents`. That wasn't verified the way the skill-path derivation was (no table confirming each repo actually defines a `sync:agents` script). Not a defect in this PR — the three repos are explicitly out of scope here — but worth confirming before those repos adopt the asset, since a wrong command in the generated file's own header is a minor but visible papercut.

### Before merging
- Confirm each of api-common/browser-common/js-common actually defines (or will define) a `sync:agents` npm script before they adopt this asset — the generated `AGENTS.md` header hardcodes that command name, and the issue notes the three repos currently reference it differently (`CI`/`verify`/`check:agents`).

PR #435, opened 2026-08-20
feat(doctor): detect Biome $schema drift from the installed CLI

### LINKED ISSUE #424: doctor does not detect Biome $schema version drift
The Biome config generator writes the schema URL from the version it knows about (`src/cli/generators/linting.ts:103`):

```ts
    $schema: `https://biomejs.dev/schemas/${version}/schema.json`,
```

But the corresponding check (`src/languages/js/checks.ts:113`) only asserts the string *contains* the host path:

```ts
    typeof schema === 'string' &&
    schema.includes('biomejs.dev/schemas/') &&
    linter?.rules?.preset !== undefined
```

So once a repo's `$schema` is written, it is never compared against the Biome actually installed. It goes stale silently and `doctor` reports no drift.

## Observed

`rtorcato/browser-common` pins `2.4.16` in both `biome.jsonc` and `apps/docs/biome.json`, while repo-tooling depends on `@biomejs/biome: ^2.5.0` and the resolved CLI is `2.5.5`. `doctor` is clean, but every Biome run — including the Husky pre-commit hook — prints:

```
biome.jsonc:2:14 deserialize ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  i The configuration schema version does not match the CLI version 2.5.5

  >  2 │   "$schema": "https://biomejs.dev/schemas/2.4.16/schema.json",

  i   Expected: 2.5.5
      Found:    2.4.16

  i Run the command biome migrate to migrate the configuration file.
```

Two infos on every commit in every repo whose Biome has moved on. It is only informational, so nothing fails — which is the problem: it is pure noise that trains people to skim past pre-commit output, and it means editors validate against a schema two minors old.

## Suggestion

Have the check parse the version out of the `$schema` URL and compare it to the installed `@biomejs/biome` (the same value the generator already resolves), reporting drift when they differ. The fixer needs no new logic — regenerating the line already produces the right value.

There is already a `/** The version out of https://biomejs.dev/schemas/2.5.0/schema.json */` comment at `src/languages/js/checks.ts:1431`, so the version is evidently parsed somewhere nearby — this may be a small wiring gap rather than new work.

Low severity, cheap fix, and it silences a recurring diagnostic across every repo in the family.

---

rtorcato:

## Problem

The Biome check only asserted the `$schema` string contained `biomejs.dev/schemas/`, so once `fix biome` had written a version it was never compared against anything again. A repo whose Biome moved on stayed clean in `doctor` while every single Biome invocation — the Husky pre-commit hook included — printed:

```
i The configuration schema version does not match the CLI version 2.5.5
  Expected: 2.5.5
  Found:    2.4.16
```

Informational, so nothing fails — which is the problem. It is recurring noise on a file the consumer never wrote by hand.

## Change

`checkBiome` runs the existing file check and, when it passes, parses the version out of the `$schema` URL and compares it to the `@biomejs/biome` actually installed under the target directory, reporting `drift` when they differ.

The comparison is against the **installed package only**, never the declared range: an exact match is what Biome itself demands, and `^2.5.0` says nothing about which 2.x resolved. With no `node_modules` there is nothing to be wrong about, so the check stays quiet rather than guessing — which is also why every existing tmp-dir test is unaffected.

`installedBiomeVersion` is lifted out of `resolveBiomeSchemaVersion` so the generator and the check read the version from one place.

No new fixer logic: `fix biome` already regenerates the URL from the resolved version, and its existing `canFixDrift: true` path picks the finding up.

## Dogfood

This repo was exactly the described drift — `biome.json` and `apps/docs/biome.json` both targeted 2.5.0 against an installed 2.5.5. `pnpm dogfood` fails on the new finding, so both are bumped to 2.5.5. The shipped preset (`tooling/biome/biome.json`) is deliberately left at 2.5.0, since it is compared against the `^2.5.0` peer floor by a different check.

## Verification

`pnpm run check`, `pnpm run typecheck`, `pnpm test --run` (928 passing), `pnpm run build-cli`, and `pnpm dogfood` are all clean. Two tests added to `tests/cli/commands/doctor.test.ts` cover the drift and the matching case.

Closes #424

### DIFF
--- a/apps/docs/biome.json
+++ b/apps/docs/biome.json
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://biomejs.dev/schemas/2.5.0/schema.json",
+  "$schema": "https://biomejs.dev/schemas/2.5.5/schema.json",
   "root": false,
   "extends": ["@rtorcato/repo-tooling/biome"],
   "files": {
--- a/biome.json
+++ b/biome.json
@@ -1,4 +1,4 @@
 {
-  "$schema": "https://biomejs.dev/schemas/2.5.0/schema.json",
+  "$schema": "https://biomejs.dev/schemas/2.5.5/schema.json",
   "extends": ["@rtorcato/repo-tooling/biome"]
 }
--- a/src/cli/commands/doctor.ts
+++ b/src/cli/commands/doctor.ts
@@ -42,6 +42,7 @@ import type { CheckResult, CheckStatus } from '../../base/types.js'
 import {
 	allDeps,
 	checkAreTheTypesWrong,
+	checkBiome,
 	checkClaudeWorktreeSettings,
 	checkConfigSchemaVersions,
 	checkDocsSite,
@@ -385,7 +386,12 @@ export async function runDoctor(dir: string): Promise<CheckResult[]> {
 	results.push(await checkNodeVersionPin(targetDir))
 	results.push(await checkNodeVersionConsistency(targetDir, pkg))
 	for (const spec of FILE_CHECKS) {
-		results.push(await checkFile(targetDir, spec))
+		// Biome is the one spec that needs the directory as well as the file
+		// contents — it compares the `$schema` version against the installed
+		// CLI (#424). Dispatched here so it keeps its place in the output.
+		results.push(
+			spec.check === 'Biome' ? await checkBiome(targetDir) : await checkFile(targetDir, spec)
+		)
 	}
 	results.push(await checkLintStaged(targetDir, pkg))
 	results.push(await checkVerifyScript(targetDir, pkg))
--- a/src/cli/generators/linting.ts
+++ b/src/cli/generators/linting.ts
@@ -59,6 +59,22 @@ async function readJsonOrNull(filepath: string): Promise<Record<string, unknown>
 }
 
 /**
+ * The `@biomejs/biome` actually installed under `targetDir`, or null when
+ * node_modules isn't populated. This is the binary that will parse the config
+ * and emit the schema-mismatch warning, so it is the only version worth
+ * comparing a written `$schema` against — a declared range like `^2.5.0` says
+ * nothing about which 2.x resolved (#424).
+ */
+export async function installedBiomeVersion(targetDir: string): Promise<string | null> {
+	const installed = await readJsonOrNull(
+		path.join(targetDir, 'node_modules', '@biomejs', 'biome', 'package.json')
+	)
+	const match =
+		typeof installed?.version === 'string' ? /(\d+\.\d+\.\d+)/.exec(installed.version) : null
+	return match?.[1] ?? null
+}
+
+/**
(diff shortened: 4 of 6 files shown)


rtorcato (commented):

**Summary: clean, no criticals/highs/mediums.**

Scope: `doctor`/`fix biome` gain a check that parses a Biome version out of the `$schema` URL already present in `biome.json`/`biome.jsonc`, and compares it against the version resolved from the installed `@biomejs/biome`'s `package.json`.

- **Injection**: none. No shell/exec, no SQL, no template rendering. `readSchemaUrl`/`schemaUrlVersion` (pre-existing, unmodified) and the new regex `/(\d+\.\d+\.\d+)/` only ever produce a version string used for equality comparison and inclusion in a CLI-printed `detail` string. No eval, no dynamic code paths.
- **Untrusted input handling**: the `$schema` string originates from a file inside the target repo (`biome.json`/`biome.jsonc`), which is local, developer-controlled config the tool already trusts elsewhere (same trust level as `package.json`, which this tool already parses). The candidate file list (`['biome.json', 'biome.jsonc']`) is a fixed literal, not derived from the parsed value, so there's no path-traversal surface from the untrusted string.
- **ReDoS**: the regex `\d+\.\d+\.\d+` is linear/non-backtracking on any input length — not exploitable.
- **Secrets**: none introduced; only version strings are read/compared/printed.
- **Dependency/supply chain**: no new dependencies. The two `biome.json` bumps (2.5.0 → 2.5.5) are dogfood fixes to this repo's own config, not the shipped preset (`tooling/biome/biome.json`, deliberately left at 2.5.0 per the PR description) — consistent with not baking org-specific pins into emitted/consumer-facing files.

### Before merging
Nothing.

rtorcato (commented):

The change is correctly wired: `checkBiome` runs the existing file check first and only compares versions when it passes, `installedBiomeVersion` is cleanly lifted out of `resolveBiomeSchemaVersion` so both read from one place, and the fallback priority (installed → declared range → `BIOME_SCHEMA_FALLBACK`) is preserved through the refactor. `checkBiome` returns early and quietly when `node_modules` isn't populated, matching the stated intent of not guessing. Two new tests cover both the drift and matching cases, and the dogfood bump of this repo's own `biome.json`/`apps/docs/biome.json` to 2.5.5 is consistent with the new check actually firing.

No bugs found in the diff itself.

### Before merging
- This `fix:` makes a previously-`ok` doctor check start reporting `drift` in any consumer repo whose installed Biome has moved past its pinned `$schema` version — which per the PR's own description is the common case across the repo family (`browser-common` was cited as already stale). That's a behavior change that turns existing green repos red without any consumer-side action, which is arguably more than a patch-level "fix" from a consumer's perspective (no code broke, but CI status changes). Worth a human call on whether `fix:` is the right semver here, or whether the release notes/changelog should call out the new drift check explicitly so consumers aren't surprised.

Issue #439, opened 2026-08-20
ai-issue-loop: Pass 3 has no claim-first guard, so a concurrent tick double-spawns reviewers


rtorcato:
🤖 *Filed by Claude (AI agent) on rtorcato's behalf, from the repo-tooling `ai-issue-loop` session on 2026-08-20.*

The loop's two spawning passes are not equally protected against a second tick running while the first is still in flight.

## Pass 4 is safe

`skills/ai-issue-loop/SKILL.md:659` — pickup claims **before** spawning:

> Take the first `slots` issues. For each, **claim it first** so a concurrent tick can't double-pick:
>
> ```bash
> gh issue edit <N> --add-label ai-wip
> ```

and the eligibility query filters on `ai-wip`. A concurrent tick sees the claim and skips. Deliberate and correct.

## Pass 3 is not

Pass 3 spawns a reviewer whenever a PR carries `ai-review` and is missing `ai-ok-code` / `ai-ok-sec`. But the reviewer only applies its label **at the end**, after reading the diff and posting its comment. Between spawn and verdict, the label state is indistinguishable from "no reviewer has started".

So a second tick in that window spawns a duplicate reviewer for every PR under review. There is no `ai-reviewing` or equivalent — `grep -n 'ai-reviewing' skills/ai-issue-loop/SKILL.md` returns nothing.

## Cost when it fires

Duplicates are not merely wasteful, they are visible:

- Two `code-reviewer` agents read the same diff and post two review comments, both under the owner's avatar
- The verdict labels race — one may apply `ai-ok-code` while the other applies `ai-changes`, leaving the PR in a contradictory state that Pass 1 then has to interpret
- On a 4-PR queue that is 8 duplicated reviewers, which is a meaningful slice of the monthly cap the limits section exists to protect

This session had a live instance of the window: four PRs were spawned to review at once, and for several minutes none of the eight reviewers had labelled yet. A `/ai-issue-loop` in that gap would have produced eight more.

## Suggested fix

Give Pass 3 the same claim-first shape Pass 4 already has — an `ai-reviewing` label applied immediately before spawning and removed by the verdict step:

| | Applied | Cleared by |
|---|---|---|
| `ai-reviewing-code` | Pass 3, before spawning `code-reviewer` | the reviewer, alongside `ai-ok-code` / `ai-changes` |
[...]

PR #432, opened 2026-08-20
chore(deps): bump @rtorcato/shared-docs to ^1.7.0

rtorcato:
🤖 *Automated — implementer via ai-issue-loop. Posted under the owner's account; not a human message.*

Picks up the two family changes the docs site was missing:

- **1.6.0** added `@rtorcato/infra-x` to `FAMILY`, so the nav dropdown, footer, and sibling grid never rendered it.
- **1.7.0** renamed `db-x` and `infra-x` to `@rtorcato/db-x` / `@rtorcato/infra-x`. `Siblings.tsx` renders `name` verbatim, so `db-x` showed bare next to cards that all read `@rtorcato/…`.

`pnpm-workspace.yaml` needed **no** `minimumReleaseAgeExclude` entry — its existing `'@rtorcato/*'` wildcard already covers this package, so nothing was written and nothing had to be backed out.

## Verification

The lockfile resolves `@rtorcato/shared-docs@1.7.0`. I confirmed against the published 1.7.0 tarball that `dist/family.js` carries the `@rtorcato/infra-x` entry.

I could **not** run \`pnpm --filter '*docs*' build\` locally: this worktree's \`node_modules\` is a symlink into the main checkout, so \`pnpm install\` wanted to purge the shared directory. I used \`pnpm install --lockfile-only\` instead and left the build to CI. Worth a look at the rendered sibling grid before merge.

## Release impact

\`chore(deps):\` — \`shared-docs\` is a devDependency of the private docs app, so this publishes nothing from this repo.

Closes #430

PR #441, opened 2026-08-20
docs(ai-issue-loop): warn implementers not to run pnpm install

### LINKED ISSUE #438: ai-issue-loop: node_modules symlink makes pnpm install fail in agent worktrees, silently

`skills/ai-issue-loop/SKILL.md:673` tells Pass 4 to link the main checkout's `node_modules` into each agent worktree:

```bash
ln -s "$ROOT/node_modules" "$WT_ROOT/$SLUG/node_modules"   # replaces worktree.symlinkDirectories
```

That makes dependencies available without a per-worktree install, which is the point. But it also makes `pnpm install` **fail** inside the worktree, and the skill never says so — so every implementer agent that needs to touch dependencies discovers it the hard way.

## Observed

While implementing #430 (bump `@rtorcato/shared-docs` to `^1.7.0`), the implementer ran `pnpm install` and got:

```
ERR_PNPM_ABORTED_REMOVE_MODULES_DIR_NO_TTY
```

pnpm resolved the symlink, decided the shared `node_modules` was a foreign directory it needed to purge, and refused to continue without a TTY to confirm. The agent fell back to `pnpm install --lockfile-only`, which updated `pnpm-lock.yaml` without touching `node_modules` — correct for the lockfile, but it meant **the verification step in the issue could not run**:

> ```sh
> pnpm --filter '*docs*' build
> ```

The issue asked for confirmation that the sibling grid renders `@rtorcato/infra-x`. That was never checked locally; the agent verified the published tarball's `dist/family.js` instead and left the build to CI. PR #432 shipped carrying an `ai-notes` flag saying exactly that.

Nothing was broken by it, but the loop silently degraded a "verify it works" step into "hope CI catches it", which is the failure mode worth fixing.

## Why it bites this class of issue specifically

Any issue whose work *is* a dependency change — a bump, a new dep, a workspace addition — needs `pnpm install` to verify. Those are also the issues most likely to be handed to an agent, since they look mechanical. So the symlink is most fragile exactly where the loop is most likely to be used.

## Options

| Option | Notes |
|---|---|
| **Document it in the implementer prompt** | Add a line: dependencies are already present via the symlink, do not run `pnpm install`; use `--lockfile-only` if the lockfile must change, and say so in the PR. Cheapest, and it is what was done by hand for #436. Does not restore the ability to build. |
| **Detect and warn in Pass 4** | Same as above, but the loop notes in the PR body that no install was possible, so the reviewer does not have to infer it. |
| **Give dependency-touching worktrees a real `node_modules`** | Skip the symlink and let the agent install. Correct but slow and disk-hungry per worktree; needs a way to know in advance that an issue is dependency-shaped. |
| **Use pnpm's own mechanism** | Investigate whether a per-worktree store/virtual-store config avoids the purge entirely, keeping one shared cache. Most correct if it works; needs verification. |

Option 1 alone closes the silent-degradation part, which is the actual defect — the agent not knowing beats the agent finding out mid-task.

## Note on scope

`skills/ai-issue-loop/SKILL.md` is **published** — `skills/*/SKILL.md` is in `package.json`'s `files`, and the skill is listed in `.claude-plugin/marketplace.json`. So this is a consumer-facing defect, not just a local workflow wart.

## Related

- #432 — the PR that hit it; carries the `ai-notes` flag about the unverified docs build
- The companion Pass 3 issue from the same session

rtorcato:

Implements Option 1 from #438: document the constraint in the Pass 4 implementer prompt.

The Pass 4 worktree links `node_modules` from the main checkout (`SKILL.md:673`). That makes deps available without a per-worktree install, but it also makes `pnpm install` abort with `ERR_PNPM_ABORTED_REMOVE_MODULES_DIR_NO_TTY` — pnpm resolves the symlink, decides the shared directory is foreign and must be purged, and won't proceed without a TTY. Implementers found that out mid-task, fell back to `--lockfile-only`, and quietly shipped without the verification step the issue asked for (#432).

Step 4 of the implementer prompt now says up front: deps are already present, don't run `pnpm install`, use `--lockfile-only` when the lockfile must change, and name in the PR body any verification command that could not run as a result.

Options 2–4 in the issue are deliberately not implemented — Option 1 is what closes the silent-degradation defect.

Scoped to the Pass 4 prompt only; #439 is editing Pass 3 in the same file concurrently.

### Verification

`tests/cli/generators/claude-skills.test.ts` and `tests/cli/commands/fix.test.ts` (the two referencing this skill) pass — 121 tests. Pre-commit biome check + typecheck passed. `AGENTS.md` is generated from `skills/repo-tooling/SKILL.md`, not this skill, so it is unaffected and needs no regeneration.

Closes #438

### DIFF
--- a/skills/ai-issue-loop/SKILL.md
+++ b/skills/ai-issue-loop/SKILL.md
@@ -694,6 +694,16 @@ Then spawn a background implementer agent:
 > 3. Read the repo's `CLAUDE.md` and obey it — especially any pre-commit build
 >    step or committed build output.
 > 4. Do the work. Conventional Commits within the branch.
+>
+>    **Do not run `pnpm install`.** This worktree's `node_modules` is a symlink to
+>    the main checkout, so pnpm sees a foreign directory it must purge first and
+>    aborts with `ERR_PNPM_ABORTED_REMOVE_MODULES_DIR_NO_TTY`. Dependencies are
+>    already present — run tests, lint and build directly. If the work *is* a
+>    dependency change, `pnpm install --lockfile-only` updates `pnpm-lock.yaml`
+>    without touching `node_modules`. When that leaves a verification step you
+>    cannot run, say so in the PR body — name the command you could not run and
+>    why — so the reviewer knows CI is the only check on it rather than assuming
+>    you ran it.
 > 5. Push and open the PR. The title must be a Conventional Commit — it becomes
 >    the squash subject on `main` and, in repos using semantic-release, decides
 >    whether a release goes out at all. Body must contain `Closes #<N>`.


rtorcato (commented):

Docs-only change to `skills/ai-issue-loop/SKILL.md`, adding guidance to the Pass 4 implementer prompt. No shell code, no secrets, no dependency or lockfile changes in this diff — only prose instructing future implementer agents.

- **Injection**: none. The added text is static, maintainer-authored prose, not interpolated from issue/PR body content (untrusted data), so there's no prompt-injection vector here.
- **Unsafe shell construction**: none introduced. The only command named is `pnpm install --lockfile-only`, a fixed flag with no variable interpolation, and it explicitly does not touch `node_modules` (avoids the purge/TTY failure mode described in #438).
- **Secrets**: none present.
- **Supply chain**: no dependency or lockfile edits in this diff.
- **Misreporting risk** (the concern this review was scoped to check): the new text directly *mitigates* rather than introduces this risk. It requires the implementer to name in the PR body any verification command it could not run and why, which is exactly the disclosure that was missing when #432 silently shipped an unverified build. There's no enforcement mechanism beyond agent compliance, but that's inherent to prompt-based instructions generally and not a regression from this change — it's strictly better than the prior silence.

No behavior change to what's technically possible in the worktree (deps were already unable to be freshly installed there); this just makes the constraint and the required disclosure explicit and consumer-visible (this SKILL.md ships in the package).

### Before merging
Nothing.

rtorcato (commented):

Scoped, doc-only change (10 additions, 0 deletions) to `skills/ai-issue-loop/SKILL.md`. It implements Option 1 from #438 as intended — the implementer prompt now tells Pass 4 agents not to run `pnpm install` against the symlinked `node_modules`, explains why (`ERR_PNPM_ABORTED_REMOVE_MODULES_DIR_NO_TTY`), gives `pnpm install --lockfile-only` as the escape hatch for lockfile changes, and requires naming any verification command that couldn't run in the PR body. That matches the issue's stated fix for the silent-degradation defect (the #432 incident).

Checked:
- Content accuracy: `--lockfile-only` is the correct pnpm flag for updating `pnpm-lock.yaml` without touching `node_modules`.
- Markdown formatting: the inserted nested paragraph (blank blockquote line + 3-space indent) is consistent with the blockquote-list continuation style used elsewhere in that prompt block.
- Scope: correctly limited to the Pass 4 implementer prompt; does not touch Pass 3 (left to #439/#442), and Options 2–4 are deliberately absent per the issue's own recommendation.
- No `AGENTS.md` regeneration needed — that's generated from `skills/repo-tooling/SKILL.md`, not this file, as the PR description notes.

No bugs, security issues, or convention violations found.

### Before merging
Nothing.

PR #442, opened 2026-08-20
fix(claude): claim reviewers before spawning in ai-issue-loop Pass 3

### LINKED ISSUE #439: ai-issue-loop: Pass 3 has no claim-first guard, so a concurrent tick double-spawns reviewers

rtorcato:
## Summary

Pass 4 claims an issue with `ai-wip` **before** spawning an implementer, so a concurrent tick cannot double-pick it. Pass 3 had no equivalent — a reviewer applies its verdict label only at the *end*, after reading the diff and posting its comment, so from spawn until then the PR's labels are indistinguishable from "nobody has started". A tick landing in that window spawned a duplicate of every reviewer in flight.

Adds `ai-reviewing-code` / `ai-reviewing-sec` as a transient claim: Pass 3 applies one immediately before each spawn and skips a reviewer already claimed; the reviewer clears its own alongside its verdict label. Two labels rather than one, because the reviewers spawn independently and a single flag could not say which was already running.

Updated consistently across: the Labels table, the `gh label create` block, the state-machine diagram, Pass 3's spawn logic, both reviewer prompt templates (normal and Dependabot), and Pass 2's "Reviewer died" stall row — which now keys on the claim rather than on `ai-review`, so dropping the claim is what lets Pass 3 re-spawn.

`apps/docs/docs/guides/ai-issue-loop.md` carries a mirrored labels table and diagram; both are updated so the published guide does not drift from the skill.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [x] Documentation
- [ ] CI / tooling

## Test plan

- [x] `./node_modules/.bin/vitest run tests/cli/generators/claude-skills.test.ts tests/cli/commands/fix.test.ts` — 121 passed
- [x] Diagram and labels table verified consistent between `skills/ai-issue-loop/SKILL.md` and `apps/docs/docs/guides/ai-issue-loop.md`

## Notes

`skills/*/SKILL.md` is in `package.json`'s `files` and the skill is listed in `.claude-plugin/marketplace.json`, so this is published content — `fix:` is the honest type.

Observed live: a tick spawned eight reviewers across four PRs, and for several minutes none had labelled. A second tick in that gap would have produced eight more.

Closes #439

### DIFF
--- a/apps/docs/docs/guides/ai-issue-loop.md
+++ b/apps/docs/docs/guides/ai-issue-loop.md
@@ -76,6 +76,8 @@ that state, so a missed tick, a crash, or a restart costs nothing.
 | `ai-blocked` | issue | Agent gave up; needs a human. |
 | `holding` | issue | A gate — closes on human judgement, never picked up. |
 | `ai-review` | PR | Awaiting agent review. |
+| `ai-reviewing-code` | PR | `code-reviewer` claimed and running. Cleared with its verdict. |
+| `ai-reviewing-sec` | PR | `security-expert` claimed and running. Cleared with its verdict. |
 | `ai-ok-code` | PR | `code-reviewer` passed. |
 | `ai-ok-sec` | PR | `security-expert` passed. |
 | `ai-changes` | PR | A reviewer requested changes. |
@@ -83,12 +85,17 @@ that state, so a missed tick, a crash, or a restart costs nothing.
 
 ```
 issue: ai-ready ─pickup─> ai-wip ─> PR opened, labelled ai-review
-PR: ai-review ─> reviewers ─┬─> ai-ok-code + ai-ok-sec ─┬─ issue PR  ─> assigned to you
-                            │        (± ai-notes)       │              ─> YOU merge
-                            │                           └─ dependabot ─> auto-merge
-                            └─> ai-changes ─> fix round (max 2) ─> ai-review
-                                                        └─ round 3 ─> ai-blocked
+PR: ai-review ─> ai-reviewing-* ─┬─> ai-ok-code + ai-ok-sec ─┬─ issue PR  ─> assigned to you
+                                 │        (± ai-notes)       │              ─> YOU merge
+                                 │                           └─ dependabot ─> auto-merge
+                                 └─> ai-changes ─> fix round (max 2) ─> ai-review
+                                                             └─ round 3 ─> ai-blocked
 ```
+
+`ai-reviewing-code` / `ai-reviewing-sec` are the claim step. Pass 3 applies one
+immediately before spawning that reviewer and skips spawning a second while it is
+set, so a tick that fires mid-review cannot double-spawn; the reviewer clears its
+own claim alongside its verdict.
 
 `ai-changes` is the send-back — **never** re-apply `ai-ready` to an open PR's
 issue; that is what double-picks it.
--- a/skills/ai-issue-loop/SKILL.md
+++ b/skills/ai-issue-loop/SKILL.md
@@ -383,7 +383,7 @@ work must never be reaped out from under itself.
 | Stalled | Condition | Do |
 |---|---|---|
 | Implementer died | issue `ai-wip` ≥45min, **and no PR exists** for `ai-<N>-<slug>` | `gh issue edit <N> --add-label ai-blocked --remove-label ai-wip --add-assignee @me`, comment, remove the worktree |
-| Reviewer died | PR `ai-reviewing-code` (or `ai-reviewing-sec`) ≥45min with no matching `ai-ok-*` and no `ai-changes` | `gh pr edit <N> --remove-label ai-reviewing-code` — dropping the claim is what lets Pass 3 re-spawn it, and they're cheap and diff-scoped. If that claim has been applied ≥3 times, `ai-blocked` instead |
+| Reviewer died | PR `ai-reviewing-code` (or `ai-reviewing-sec`) ≥45min with no matching `ai-ok-*` and no `ai-changes` | `gh pr edit <N> --remove-label <the claim that stalled>` — drop **that** label, not a fixed one; a stalled `ai-reviewing-sec` cleared as `ai-reviewing-code` leaves the dead claim in place and the reviewer never re-spawns. Dropping the claim is what lets Pass 3 re-spawn it, and they're cheap and diff-scoped. If that claim has been applied ≥3 times, `ai-blocked` instead |
 | Orphan worktree | `"$WT_ROOT"/ai-<N>-*` whose issue is not `ai-wip` and has no open PR | remove the worktree and branch |
 
 The **no PR exists** condition on the first row is what makes reaping safe. An
--- a/skills/ai-issue-loop/SKILL.md
+++ b/skills/ai-issue-loop/SKILL.md
@@ -60,6 +60,8 @@ header names the agent *and* says why it is wearing a human's face:
 | `ai-wip` | issue | Claimed; a worktree exists. |
 | `ai-blocked` | issue | Agent gave up; needs a human. |
 | `ai-review` | PR | Awaiting agent review. |
+| `ai-reviewing-code` | PR | `code-reviewer` claimed and running. Cleared with its verdict. |
+| `ai-reviewing-sec` | PR | `security-expert` claimed and running. Cleared with its verdict. |
 | `ai-ok-code` | PR | `code-reviewer` passed. |
 | `ai-ok-sec` | PR | `security-expert` passed. |
 | `ai-changes` | PR | A reviewer requested changes. |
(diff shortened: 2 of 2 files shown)


rtorcato (commented):

Documentation-only change (`SKILL.md` + mirrored docs guide) — no executable code, scripts, or dependencies touched, so no injection/secrets/supply-chain surface here.

Checked the specific fail-open concern: can a PR reach a pass state (`ai-ok-code`/`ai-ok-sec`) without a reviewer actually having run?

- Pass 3 claims (`ai-reviewing-code`/`ai-reviewing-sec`) immediately before spawning and skips spawning if the claim is already set — same shape as Pass 4's existing `ai-wip` claim, which is the established safe pattern here.
- Each reviewer clears its own claim label in the *same* `gh pr edit` command that applies its verdict (`ai-ok-*` or `ai-changes`), on both the normal and Dependabot templates. There's no path where the claim is cleared without a verdict label being applied alongside it, by the reviewer's own doing.
- Pass 2's reaping of a dead reviewer (`ai-reviewing-*` ≥45min, no matching `ai-ok-*`, no `ai-changes`) only removes the claim label — it never adds `ai-ok-code`/`ai-ok-sec`. That resets the PR to "unclaimed, no verdict," which lets Pass 3 re-spawn it; it cannot be mistaken for a pass since the pass labels are never touched by the reap step.

So the gate does not fail open: the only way `ai-ok-*` gets applied is a reviewer's own command, and reaping a dead claim only clears the claim, never fabricates a pass.

One doc nit, not a defect: the "Reviewer died" row's example command hardcodes `gh pr edit <N> --remove-label ai-reviewing-code` even though the condition it documents covers `ai-reviewing-sec` too — a reader has to infer they should swap the label name for the sec case.

### Before merging
Nothing.

rtorcato (commented):

Checked internal consistency across the Labels table, `gh label create` block, state-machine diagram, Pass 3's claim-before-spawn logic, both reviewer prompt templates, and the Pass 2 stall row — plus cross-checked `skills/ai-issue-loop/SKILL.md` against the mirrored `apps/docs/docs/guides/ai-issue-loop.md`. Labels table, label-create block, diagram, and prompt templates all agree between the two files and correctly use the parameterized `<ai-reviewing-code|ai-reviewing-sec>` form matching the reviewer being cleared.

**Bug — Pass 2's "Reviewer died" row hardcodes the wrong label to remove.**

```
| Reviewer died | PR `ai-reviewing-code` (or `ai-reviewing-sec`) ≥45min with no matching `ai-ok-*` and no `ai-changes` | `gh pr edit <N> --remove-label ai-reviewing-code` — dropping the claim is what lets Pass 3 re-spawn it, ... |
```

The condition column allows either claim label to be the one that stalled, but the "Do" column always removes `ai-reviewing-code`. If `ai-reviewing-sec` (security-expert) is the one that stalled, executing this literally removes a label that was never set and leaves the actual `ai-reviewing-sec` claim in place. Per this PR's own design, a lingering claim is exactly what blocks Pass 3 from re-spawning — so the security-expert path can never self-heal from a died reviewer, silently defeating the reaping this row exists to do. Should read `--remove-label <ai-reviewing-code|ai-reviewing-sec>` (whichever is present), matching the parameterization already used correctly elsewhere in this same diff (e.g. the verdict commands in the reviewer prompt templates).

Everything else — Labels table, label-create block, diagram, Pass 3 spawn logic, both reviewer prompt templates (normal and Dependabot) — is internally consistent and matches the issue's suggested design.

### Before merging
- Pass 2's "Reviewer died" row (SKILL.md) hardcodes `--remove-label ai-reviewing-code` instead of parameterizing on whichever claim label actually stalled, so a died `security-expert` claim would never be cleared and that reviewer could never be re-spawned.

rtorcato (commented):

Re-reviewed after the round-3 fix. The change is documentation-only (`skills/ai-issue-loop/SKILL.md`, `apps/docs/docs/guides/ai-issue-loop.md`) — no scripts, no dependencies, no secrets touched.

The fix itself is correct and closes the gap flagged last round: the "Reviewer died" stall row now reads `--remove-label <the claim that stalled>` instead of a hardcoded `ai-reviewing-code`, with prose explicitly calling out that a stalled `ai-reviewing-sec` cleared as `ai-reviewing-code` would leave the dead claim in place and permanently wedge that reviewer.

Checked the fail-open concern directly: the claim label (`ai-reviewing-code`/`ai-reviewing-sec`) is cleared in exactly two places in this diff, and both preserve the required invariant.

- Reviewer verdict paths (pass and fail) clear the claim in the *same* `gh pr edit` command that adds the real verdict label (`ai-ok-code`/`ai-ok-sec` or `ai-changes`) — the claim can't be dropped without a verdict landing alongside it.
- The stall-reap path (Pass 2) only ever runs `--remove-label <the claim that stalled>` — it never adds `ai-ok-*` or any other pass signal. Reaping unblocks re-spawn; it does not fabricate a verdict.

So a PR cannot reach `ai-ok-code`/`ai-ok-sec` without a reviewer's own command applying it, and the reap path that clears a stuck claim is verdict-free. No gate-fails-open path found in this diff.

No injection, secret, or shell-safety concerns — the `gh pr edit` snippets are prose templates for an agent to fill from known label names, not executed against untrusted input.

### Before merging
Nothing.

rtorcato (commented):

Verified the fix for the prior round's finding. Pass 2's "Reviewer died" stall row (`skills/ai-issue-loop/SKILL.md`) now reads:

> `gh pr edit <N> --remove-label <the claim that stalled>` — drop **that** label, not a fixed one; a stalled `ai-reviewing-sec` cleared as `ai-reviewing-code` leaves the dead claim in place and the reviewer never re-spawns.

This replaces the hardcoded `--remove-label ai-reviewing-code`, and the condition column correctly keys on either claim label (`ai-reviewing-code` (or `ai-reviewing-sec`)) rather than the old shared `ai-review` condition. A dead `security-expert` claim will now be reaped and re-spawnable.

Checked the rest of the state machine for the same class of mismatch (labels table, `gh label create` block, diagram, Pass 3 spawn logic, both reviewer prompt templates) across both `skills/ai-issue-loop/SKILL.md` and `apps/docs/docs/guides/ai-issue-loop.md`:

- Labels table: both files add identical `ai-reviewing-code` / `ai-reviewing-sec` rows.
- `gh label create` block (SKILL.md only, as before — apps/docs doesn't mirror this section): both new labels created.
- Diagrams in both files: consistently show `ai-review ─> ai-reviewing-* ─┬─>` branching to either the ok-labels path or `ai-changes`.
- Pass 3: claims with `gh pr edit <N> --add-label ai-reviewing-code|sec` immediately before spawning, and skips spawning a reviewer whose claim is already set. Order (claim, then spawn) matches Pass 4's existing issue-claim pattern.
- Both reviewer prompt templates (normal and Dependabot): verdict commands now include `--remove-label <ai-reviewing-code|ai-reviewing-sec>` alongside the verdict label, so a reviewer always clears its own claim.

No remaining hardcoded claim label and no other cross-reference drift found in the diff.

### Before merging
Nothing.

PR #443, opened 2026-08-20
chore(repo): dogfood .repo-tooling.json in this repo

### LINKED ISSUE #431: 4 of 10 family repos have no .repo-tooling.json, so doctor cannot record opt-outs

`.repo-tooling.json` is only ever written by `setup`, or by an explicit `fix lockfile`. Nothing creates it retroactively and nothing fails without it, so a repo that adopted the tooling by hand — or predates the lockfile — never gets one, and never will on its own.

## Survey

`.repo-tooling.json` across the family, at CLI version 3.11.0 (`LOCKFILE_VERSION = 2`):

| Repo | Lockfile | `writtenBy` |
|---|---|---|
| js-common | v2 | `@rtorcato/repo-tooling@3.4.0` |
| shared-docs | v2 | `@rtorcato/repo-tooling@3.2.4` |
| browser-common | v1 | `@rtorcato/repo-tooling@2.17.1` |
| cf-common | v1 | `@rtorcato/js-tooling@2.19.0` |
| db-common | v1 | `@rtorcato/js-tooling@2.19.0` |
| react-common | v1 | `@rtorcato/js-tooling@2.18.0` |
| **repo-tooling** | ❌ none | — |
| **api-common** | ❌ none | — |
| **supabase-common** | ❌ none | — |
| **swift-common** | ❌ none | — |

**4 of 10 have no lockfile. repo-tooling is one of them** — it does not dogfood its own.

## Why the gap matters

`demoteDeclined()` short-circuits on a missing lockfile (`src/cli/commands/doctor.ts:189`):

```ts
function demoteDeclined(results: CheckResult[], lock: Lockfile | null): CheckResult[] {
	if (!lock) return results
```

So on those four repos, **every intentional opt-out reads as `not configured` forever**. There is no way to record a decision, which is the lockfile's entire purpose. `checkLockfile` reports the gap as `optional-missing` (`doctor.ts:164`) — deliberately non-failing, which is why it has stayed unnoticed in four repos across two major versions.

It is also the state #428 wants to hang copied-asset hashes off. That issue can't assume a lockfile exists until this is closed.

## The three `js-tooling@2.x` files are separately stale

`readLockfile` migrates v1→v2 **in memory only**; the file is rewritten on the next *write* (`src/cli/utils/lockfile.ts:33-40`). Since nothing writes on a plain `doctor` run, `cf-common`, `db-common` and `react-common` keep a `writtenBy` naming a package that is now 404 on the registry (#393). `writeLockfile` already handles the migration correctly, including removing a leftover `.js-tooling.json` — it just needs to be triggered.

## Ask

1. **Sweep** — run `fix lockfile` on the four repos with none, and the three carrying a `js-tooling` stamp. Mechanical; one commit each.
2. **Dogfood** — repo-tooling should carry its own `.repo-tooling.json`.
3. **Consider promoting the check.** `optional-missing` is arguably too quiet for a file this load-bearing. A repo that has clearly adopted the standard (has `.github/workflows/`, a Biome config, etc.) but no lockfile is drift, not an opt-out. That distinction needs deciding before implementing.

Items 1 and 2 are agent-suitable. Item 3 is a judgement call about check severity and should probably be split out.

## Related

- #428 — copied-asset drift detection, which wants to store hashes in this lockfile
- #393 — stale `js-tooling` naming in generated files

rtorcato:
🤖 *Authored by Claude (AI agent) on rtorcato's behalf, via the repo-tooling `ai-issue-loop`.*

Adds `.repo-tooling.json` to this repo, which shipped the lockfile format but never carried one. Without it, `demoteDeclined()` short-circuits (`src/cli/commands/doctor.ts:191`) and every intentional opt-out here reads as "not configured" forever.

Generated with the CLI's own writer — `writeLockfile()` + `inferProjectConfig()` off the freshly built `dist` — rather than hand-written JSON. `fix lockfile` itself is blocked in this repo by the self-repo guard in `src/cli/index.ts:370` (whose `REPO_TOOLING_ALLOW_SELF` escape hatch is deliberately scoped to read-only `doctor`), so the fixer's exact code path was invoked directly instead of relaxing that guard.

Two corrections to the inferred config, because the inference was wrong for this repo:

- **`bundler: "none"`** — inference says `tsup` (it is a devDependency, as a shipped preset), but this repo builds with plain `tsc` (`build-cli`). A recorded `tsup` would make `fix --resync` scaffold a `tsup.config.ts` that does not belong here.
- **`language: "js"`** — v2 lockfiles record it; `inferProjectConfig` omits it.

## Verification

`doctor` before → after (nothing else moved, 56 checks both runs):

| Check | Before | After |
|---|---|---|
| lockfile | `optional-missing` | `ok` — `.repo-tooling.json v2 (written by @rtorcato/repo-tooling@3.11.0)` |
| ESLint | `optional-missing` | `ok` — intentionally declined |
| Prettier | `optional-missing` | `ok` — intentionally declined |

The two demotions are accurate: this repo lints and formats itself with Biome.

`vitest run` 933/933 pass, `biome check src scripts` clean, `node scripts/dogfood.mjs` clean.

## Deliberately out of scope

Issue #431 lists three asks; **this PR implements only item 2**. `Closes #431` should not be read as items 1 and 3 being done:

- **Item 1 (sweep the other repos)** — not done. It means committing to nine other repositories; an agent working in this repo has no business writing to them. Worth a separate, per-repo pass.
- **Item 3 (promote the check from `optional-missing` to `drift`)** — not done. The issue itself flags it as an unresolved judgement call about check severity and suggests splitting it out. Still open for a decision.

Closes #431

### DIFF
--- a/.repo-tooling.json
+++ b/.repo-tooling.json
@@ -1,0 +1,31 @@
+{
+  "$schema": "https://rtorcato.github.io/repo-tooling/schemas/lockfile.json",
+  "version": 2,
+  "config": {
+    "projectName": "@rtorcato/repo-tooling",
+    "projectType": "library",
+    "typescript": {
+      "enabled": true,
+      "config": "base"
+    },
+    "linting": {
+      "tool": "biome",
+      "eslintConfig": "base"
+    },
+    "formatting": {
+      "tool": "biome"
+    },
+    "testing": {
+      "framework": "vitest",
+      "environment": "node"
+    },
+    "gitHooks": true,
+    "commitLint": true,
+    "semanticRelease": true,
+    "securityAutomation": true,
+    "bundler": "none",
+    "language": "js"
+  },
+  "writtenBy": "@rtorcato/repo-tooling@3.11.0",
+  "writtenAt": "2026-08-20T17:32:17.649Z"
+}


rtorcato (commented):

Scope: this PR's diff is a single new file, `.repo-tooling.json` (+31/-0). No source, CLI, or guard code is touched.

**Self-repo guard (`src/cli/index.ts:370`)**: confirmed untouched. The diff contains zero changes to `src/`, so the `REPO_TOOLING_ALLOW_SELF` scoping to read-only `doctor` is neither weakened, bypassed, nor widened. The PR body's claim that the writer's code path was invoked directly (outside the CLI's `fix lockfile` guard) describes a local generation step, not something reflected in — or reachable through — this diff.

**Committed `.repo-tooling.json` contents**: reviewed line by line. No secrets, tokens, credentials, or machine-local paths. Fields present are: public project name/type, tool choices (biome/vitest/tsc), boolean feature flags, a public GitHub Pages `$schema` URL, a semver `writtenBy` string, and an ISO `writtenAt` timestamp. All non-sensitive.

**Injection / unsafe shell / dependency risk**: none — this is a static JSON data file, no executable code, no `child_process` usage, no new/changed dependencies (`package.json`/lockfiles not touched).

Nothing further to flag; this is a low-risk, data-only addition.

### Before merging
Nothing.

rtorcato (commented):

**Correctness**: The diff is a single generated JSON file — no code changes to review for bugs. The `writtenAt` timestamp matches today's date, `$schema` and `version: 2` match the family's existing v2 lockfiles (js-common, shared-docs) shown in the issue's survey table.

**Hand-corrected fields**: Both overrides hold up on the reasoning given.
- `bundler: "none"` — plausible: `tsup` being a devDependency because repo-tooling *ships* a tsup preset for consumers (not because it builds with tsup itself, which the PR says uses plain `tsc`) is a believable inference false-positive, and the stated failure mode (`fix --resync` scaffolding a stray `tsup.config.ts`) is a real, well-motivated reason to hand-fix rather than trust the inferrer.
- `language: "js"` — makes sense given the visible schema shape: `typescript.enabled` is a nested flag independent of the top-level `language` field, and the repo family includes non-JS members (`swift-common`), so `language` is plausibly the macro bucket ("js" vs "swift") rather than a JS/TS distinction. Consistent with a TS-authored repo still recording `language: "js"`.

I can't fully verify either claim without exploring the repo (out of scope for this review), but neither raises a flag and the reasoning is internally consistent and well-motivated in the PR body.

**Scope / `Closes #431`**: The PR is explicit that it implements only item 2 of the issue's three asks, and the reasons for deferring items 1 and 3 are sound (item 1 requires writing to nine other repos, which an agent shouldn't do unattended; item 3 is a judgement call the issue itself flags as needing to be split out). That said, closing #431 outright means items 1 and 3 have no tracking issue once this merges — the "worth a separate pass" and "still open for a decision" language in the PR body is a stated intent, not a filed issue. That's a process question for a human, not a code defect.

### Before merging
- Before merging, decide whether to close #431 as-is (losing the tracking thread for the item-1 sweep and the item-3 severity decision) or file a follow-up issue for those two items first — the PR body flags both as still needed but doesn't link a tracking issue for either.

PR #444, opened 2026-08-20
feat(docusaurus): add docs-generator helpers as a copy asset

### LINKED ISSUE #426: Extract js-common's docs-generator helpers into a shared asset
`js-common` has three helpers in `scripts/` that are repo-tooling's problem, not a utility library's. They don't belong in `@rtorcato/js-common`'s export surface (build tooling, not runtime utilities), but they're general enough that every `@rtorcato` package with a Docusaurus site will want them.

Filing for later — **not** `ai-ready`, no work wanted yet.

## The helpers

Sources, all in [rtorcato/js-common](https://github.com/rtorcato/js-common):

1. **`escapeForMarkdownTable(text)`** — `scripts/generate-module-docs.mjs:34`. Makes JSDoc summaries safe for a markdown table cell. Two non-obvious behaviours, both already covered by tests in `scripts/generate-module-docs.test.mjs`:
   - escapes `<` *outside* code spans rather than stripping tags, so `Success<T>` survives and `<<b>>` can't leave a live `<b>` behind
   - escapes `\` and `|` in one pass, because escaping only `|` turns `a\|b` into `a\\|b` — an escaped backslash plus a live pipe, which still breaks the row

2. **`collectExportNames(file)`** — the recursive `export function|const|class|type|interface` + `export { … }` + `export * from './x'` parser. Genuinely reusable: any doc generator or boundary check over a subpath-exports package needs it.

3. **`spliceGeneratedBlock(existing, block)` + the `MARKER_START`/`MARKER_END` pair** — `generate-module-docs.mjs:104-117`. Rewrites only the fenced generated region so hand-written prose around it survives, and returns `null` for a page with no markers so a fully hand-written page is left alone.

## Why this is worth doing

**The parser has already been duplicated and has already drifted.** js-common carries two copies — `scripts/check-readme-exports.mjs:20-52` and `scripts/generate-module-docs.mjs:25-98`. `EXPORT_NAMED` and `EXPORT_BRACE` are byte-identical in both. The other two have silently diverged:

| | `check-readme-exports.mjs` | `generate-module-docs.mjs` |
|---|---|---|
| `REEXPORT_FROM` | matches `export *` **and** `export { … } from`, any specifier | `export *` only, and only `./relative` paths |
| alias handling | `.split(/\s+as\s+/).pop()` → `bar` | `.split(/\s+as\s+/)[0]` → `foo` |

For `export { foo as bar }` the second one is simply **wrong** — `bar` is the exported name, so the docs table would list `foo`, a symbol no consumer can import. It doesn't bite today only because js-common happens to have no aliased re-exports; nothing would catch it when one lands.

That's the argument in miniature: one parser, two copies, no test that they agree, and they don't.

## Shape

Probably a copy-asset alongside the existing `docusaurus-sync-changelog` and `docusaurus-theme-tokens`, since consumers need to edit the generator's config (which subpaths, where the docs live) rather than call a fixed API. But an importable module is defensible for `escapeForMarkdownTable` and `collectExportNames`, which are pure functions — worth deciding before anyone writes code.

Whatever the shape, js-common should then delete its copies and consume the shared one, and `check-readme-exports.mjs` and `generate-module-docs.mjs` should share a single parser.

## Related

- js-common#236's sibling pattern — promoting a component into `shared-docs` and importing it back
- rtorcato/repo-tooling#411 (`sync-changelog.mjs` is currently a local fork in js-common awaiting that)

rtorcato:

Closes #426

## What

Adds `copy docusaurus-docs-helpers` → `scripts/docs-helpers.mjs`, registered next to `docusaurus-sync-changelog` and `claude-sync-agents`.

**Shape: a copy-asset, not an importable module.** Consumers edit the generator's config (which subpaths, where the docs live), so a copied script fits better than a fixed API — and it matches the sibling assets.

## The three helpers

| Export | Behaviour preserved |
| --- | --- |
| `escapeForMarkdownTable(text)` | Escapes `<` *outside* code spans, so `` `Success<T>` `` survives and `<<b>>` can't leave a live `<b>` behind. Escapes `\` and `\|` in **one pass** — escaping only `\|` turns `a\|b` into an escaped backslash plus a live pipe, which still breaks the row. |
| `collectExportNames(file)` | The recursive export parser, following relative re-exports into the files they name. |
| `spliceGeneratedBlock(existing, block)` + `MARKER_START`/`MARKER_END` | Rewrites only the fenced region; returns `null` for a page with no markers, so a fully hand-written page is left alone. |

## Bug fixed while porting

js-common's two copies of the parser have diverged, and this ports the correct half of each:

- **Alias handling.** `generate-module-docs.mjs` uses `.split(/\s+as\s+/)[0]`, so `export { foo as bar }` yields `foo` — a symbol no consumer can import; `bar` is the exported name. Ported as `.pop()`, with a test named for exactly that case.
- **`REEXPORT_FROM`.** Ported from `check-readme-exports.mjs`, which matches `export *` **and** `export { … } from` with any specifier, rather than `export *` with `./relative` paths only.

## Notes

- `MARKER_START` / `MARKER_END` are byte-identical to js-common's, so pages it has already generated keep splicing after adoption.
- Nothing org-specific is baked into the asset — no package names, no paths, no I/O beyond reading the files it is handed.
- **js-common's adoption is a follow-up in that repo**: deleting its two copies and rewiring `check-readme-exports.mjs` and `generate-module-docs.mjs` onto this single parser is out of scope here.

## Tests

10 new tests in `tests/cli/docusaurus-docs-helpers.test.ts`, exercising the asset the way consumers get it (copied out, then imported): the copy target, both `escapeForMarkdownTable` behaviours, the splice and its `null` case, and the parser across declaration/brace/aliased exports, both re-export forms, a bare specifier, a directory `index.ts`, an import cycle, and a missing file.

Full suite: 925 passed / 18 skipped. Biome and `tsc --noEmit` clean.

### DIFF
--- a/apps/docs/docs/guides/cli.md
+++ b/apps/docs/docs/guides/cli.md
@@ -27,7 +27,7 @@ npx @rtorcato/repo-tooling copy biome     # → biome.json
 npx @rtorcato/repo-tooling copy tsconfig  # → tsconfig.json
 ```
 
-Available presets: `biome`, `tsconfig`, `bun`, `nx`, `changesets`, `release-please`, `oxlint`, `claude-skill`, `claude-sync-agents`, `mcp-example`, `docusaurus-sync-changelog`, `docusaurus-theme-tokens`, `docusaurus-theme`.
+Available presets: `biome`, `tsconfig`, `bun`, `nx`, `changesets`, `release-please`, `oxlint`, `claude-skill`, `claude-sync-agents`, `mcp-example`, `docusaurus-sync-changelog`, `docusaurus-docs-helpers`, `docusaurus-theme-tokens`, `docusaurus-theme`.
 
 `claude-sync-agents` copies `scripts/sync-agents.mjs`, which regenerates `AGENTS.md` from `skills/<package>/SKILL.md` so the two can't drift. It is zero-config — the skill directory comes from the root `package.json` `name` with the npm scope stripped (`@rtorcato/js-common` → `skills/js-common/`). Wire `node scripts/sync-agents.mjs --check` into CI to fail on a stale `AGENTS.md`.
 
--- a/apps/docs/docs/reference/docusaurus.md
+++ b/apps/docs/docs/reference/docusaurus.md
@@ -33,6 +33,29 @@ hooks reliably, so chain it in `build`/`start` rather than relying on `prebuild`
 }
 ```
 
+## `copy docusaurus-docs-helpers`
+
+Drops `scripts/docs-helpers.mjs` into your repo:
+
+```bash
+npx @rtorcato/repo-tooling copy docusaurus-docs-helpers
+```
+
+The pure pieces of a docs generator for a subpath-exports package, so the
+generator script above them stays small and project-specific:
+
+| Export | What it does |
+| --- | --- |
+| `escapeForMarkdownTable(text)` | Makes a JSDoc summary safe for a table cell — escapes `<` outside code spans (so `` `Success<T>` `` survives and `<<b>>` can't leave a live `<b>`), and escapes `\` and `\|` in one pass. |
+| `collectExportNames(file)` | Recursively parses `export function\|const\|class\|type\|interface\|enum`, `export { … }`, and both re-export forms, following relative specifiers into the files they name. For `export { foo as bar }` the exported name is `bar`. |
+| `spliceGeneratedBlock(existing, block)` | Rewrites only the region between `MARKER_START` and `MARKER_END`, so hand-written prose around it survives. Returns `null` for a page with no markers — a fully hand-written page is left alone. |
+
+Wire it into whatever generator your repo needs:
+
+```js
+import { collectExportNames, escapeForMarkdownTable, spliceGeneratedBlock } from './docs-helpers.mjs'
+```
+
 ## `copy docusaurus-theme-tokens`
 
 Drops the shared design-token base into `apps/docs/src/css/_jt-tokens.css`:
--- a/package.json
+++ b/package.json
@@ -87,6 +87,7 @@
 		"tooling/docusaurus/index.mjs",
 		"tooling/docusaurus/index.d.mts",
 		"tooling/docusaurus/sync-changelog.mjs",
+		"tooling/docusaurus/docs-helpers.mjs",
 		"tooling/docusaurus/theme-tokens.css",
 		"tooling/docusaurus/theme.css",
 		"tooling/biome/biome.json",
--- a/src/cli/utils/copy-preset.ts
+++ b/src/cli/utils/copy-preset.ts
@@ -21,6 +21,7 @@ export type PresetName =
 	| 'claude-sync-agents'
 	| 'mcp-example'
 	| 'docusaurus-sync-changelog'
+	| 'docusaurus-docs-helpers'
 	| 'docusaurus-theme-tokens'
 	| 'docusaurus-theme'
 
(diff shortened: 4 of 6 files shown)


rtorcato (commented):

Reviewed `tooling/docusaurus/docs-helpers.mjs` (new copy-asset), the `copy-preset.ts` registration, `package.json` files entry, and the new test file.

**No injection, secret-leak, or unsafe-shell issues.** The script does file I/O only (`readFileSync`/`existsSync`) — no `child_process`, no `eval`/`new Function`, no dynamic `import()` of anything but the test's own copied fixture. No dependencies were added; the only `package.json` change is a `files` array entry (asset manifest), not a dependency bump.

**`escapeForMarkdownTable`** — verified it can't be induced to emit live HTML. `<` is replaced with `&lt;` on every character outside a matched `` `...` `` span; a matched span (no backtick inside, single-backtick delimited) is exactly a valid CommonMark code span too, so there's no case where content classified as "code" here is *not* also treated as code (and thus escaped) by the downstream renderer — I couldn't construct an input where the split misclassifies live prose as a code span and lets `<` through. `<<b>>` → `&lt;&lt;b>>` as documented, no residual tag. The backslash/pipe pass runs across the whole joined string (including inside backticks), which over-escapes rather than under-escapes — a correctness nit, not a security regression (fails toward more escaping, not less).

**`spliceGeneratedBlock`** — uses `indexOf` for both markers and explicitly guards `end < start` → returns `null`, so it can't produce a negative-length splice or corrupt content when markers are missing/out of order. With a single marker pair (the documented contract) it only ever replaces the region between them; content outside is untouched.

**`collectExportNames`** — resolves relative re-export specifiers via `path.resolve`/`readFileSync` against the local source tree the generator is pointed at. This mirrors the read-only, developer-invoked behavior it's ported from in js-common (referenced in issue #426) and only ever reads files to regex-scan for export syntax — never writes, execs, or reflects raw file content back into output.

Tests cover the two escaping behaviours, the splice's null case, and the aliased-export bug fix (`.pop()` vs `.split()[0]`) called out in the PR description.

### Before merging
Nothing.

rtorcato (commented):

Verified the three ported helpers against the issue's documented requirements, diff-scoped (`gh pr diff 444` + `gh issue view 426`).

**`escapeForMarkdownTable`** — correct. `<` is escaped only in the non-code-span halves of the `split(/(\`[^\`]*\`)/)` result (so `` `Success<T>` `` survives, `<<b>>` → `&lt;&lt;b>> `, no live tag), and `\`/`|` are escaped in a single `.replace(/[\\|]/g, '\\$&')` pass, so `a\|b` → `a\\\|b` rather than the broken `a\\|b` a naive pipe-only escape would produce. Tests cover both cases directly.

**`collectExportNames`** — takes the fix correctly. The brace-export loop uses `.split(/\s+as\s+/).pop()`, so `export { foo as bar }` yields `bar` (the importable name), not the `.split(...)[0]` bug from `generate-module-docs.mjs` that the issue calls out. There's a test named for exactly this case (`'takes the alias, not the local name, for an aliased re-export'`), and `REEXPORT_FROM` matches both `export *` and `export { … } from` with any specifier (filtered to relative afterward), matching the `check-readme-exports.mjs` half the issue says is correct.

One thing worth a maintainer's eye but not blocking: for a *named* re-export (`export { picked } from './named.js'`), the code both adds `picked` directly at the call site *and* recurses into `named.js` via `REEXPORT_FROM`, which pulls in that file's full export set (e.g. a `notReexported` sibling const) rather than just the named symbol. This matches the function's own docstring ("following relative re-exports into the files they name") and isn't one of the two bugs the issue asked to fix, so it reads as intentional over-inclusion rather than a regression — but it does mean a doc generator built on this could list a symbol that isn't actually re-exported. Flagging as a question, not a defect.

**`spliceGeneratedBlock`** — correct. Returns `null` when either marker is missing or out of order, leaving hand-written pages untouched; only the fenced region is replaced otherwise. Test covers both the splice and the `null` path.

Copy-asset registration (`package.json` files array, `PresetName` union, `PRESETS` record, docs in `cli.md` and `reference/docusaurus.md`) all match the sibling `docusaurus-sync-changelog` pattern. Scope notes in the PR body (copy-asset shape, js-common adoption as a follow-up) match what the issue left open/out of scope.

### Before merging
Nothing.

PR #445, opened 2026-08-20
fix(dependabot): refuse to silently delete repo-local ignore rules

### LINKED ISSUE #422: fix dependabot silently deletes repo-local ignore rules
`fix dependabot` writes `.github/dependabot.yml` wholesale from `dependabotConfig()` in `src/cli/generators/security.ts`. That template emits `schedule`, `cooldown`, `groups` and `commit-message` — but **no `ignore:` block**, and no mechanism for a repo to keep one.

So any repo-local `ignore` rule is deleted the next time the fixer runs, with no warning and no drift report.

## Live example

`rtorcato/browser-common` currently carries this, added by hand when adopting the grouped config (rtorcato/browser-common#136):

```yaml
    ignore:
      # apps/docs is deliberately pinned to TypeScript ~5.6.3 — TS 7 removed
      # `baseUrl`, so a TS major is a migration to do by hand, not a dep bump.
      - dependency-name: typescript
        update-types:
          - version-update:semver-major
```

That rule is load-bearing. `apps/docs` is pinned to `typescript@~5.6.3` because `@docusaurus/tsconfig@3.10.1` sets `baseUrl`, which TS 7 removed (`TS5102`) — the root of that repo is on `^7.0.2`, the docs workspace deliberately is not. Drop the `ignore` and Dependabot opens a TS 7 major against `apps/docs`, which breaks the docs build.

One `repo-tooling fix dependabot` removes it silently. Nothing in `doctor` reports the loss, because from the fixer's point of view the file now matches the standard exactly — the drift *is* the standard.

This is a general problem, not a browser-common one: `ignore` is precisely the key that is inherently repo-specific, so a template that owns the whole file can never carry it.

## Options

1. **Merge instead of overwrite** — parse the existing `dependabot.yml` and preserve `ignore` (and any other keys the template doesn't own) when regenerating. Most correct, most work.
2. **Own an `ignore` input** — let `.repo-tooling.json` declare ignore rules so they're regenerated as part of the standard. Keeps the file fully generated and makes the constraint explicit and reviewable.
3. **Detect and warn** — cheapest: if the existing file has an `ignore:` block the template won't reproduce, have `fix` refuse (or print a loud warning naming the rules it is about to drop) rather than silently clobbering.

Option 3 alone would have prevented the failure mode; 2 is the tidiest long-term fit with how the rest of the standard is configured.

## Blast radius

Any repo that has hand-added an `ignore` rule since adopting the grouped config. Worth a quick sweep before this lands — the rules are invisible once deleted, and the symptom (a surprise major PR that breaks a build) shows up much later and looks unrelated.

---

rtorcato:

Closes #422

## What

`fix dependabot` wrote `.github/dependabot.yml` wholesale from `dependabotConfig()`, which emits no `ignore:` block. Any hand-added ignore rule was therefore deleted on the next run, silently — and `doctor` reported nothing afterwards, because the file then matches the standard exactly. The drift *is* the standard.

`ignore` is the one key that is inherently repo-local, so a template owning the whole file can never carry it.

## Scope — Option 3 only

The issue offered three options; a triage comment on the issue pinned **Option 3 (detect and refuse)**. Refusing rather than warning is the load-bearing choice: an unattended `fix --yes` walks straight past a warning, and the rules are unrecoverable once written over.

**Option 2 — owning an `ignore` input in `.repo-tooling.json` — remains open as a follow-up.** It is the tidier long-term fit (the file stays fully generated, and the constraint becomes reviewable), but it is a `ProjectConfig` schema change plus a migration story for every repo that already hand-edited the file, and it closes the data-loss window no faster than this does.

## Behaviour

| Path | Before | After |
|---|---|---|
| `fix dependabot` | overwrites, rules gone | exits 1, names each rule, file untouched |
| `fix dependabot --json` | same | `{"error": "dependabot-ignore-rules", "hint": ...}` |
| `fix --yes` (bulk) | overwrites | records the target `skipped`, applies every other fixer |
| `fix dependabot --diff` | — | previews the refusal instead of crashing in the shadow run |
| `doctor` | silent | names the rules in the Dependabot detail line |

`doctor` deliberately does **not** call an ignore block drift. A hand-held pin is correct, not a defect — it is just the reason `fix` refuses, so it has to be visible before anyone runs `fix`. A repo with canonical grouping plus an ignore rule still reads `ok`.

Nothing here is specific to `typescript` or to any one repo: `dependabotIgnoreRules()` names whatever `dependency-name`s it finds under any `ignore:` block, in any ecosystem, and reports a rule as `<unnamed rule>` when it cannot read a name — which still stops the overwrite.

## Notes for review

- The detector is a line scanner, not a YAML parse — the package ships no YAML parser, and this only needs to answer "is there something to lose, and what is it called". Marked with a `ponytail:` comment naming the ceiling.
- `fix.ts` needed two new catch sites. The bulk loop previously had no `FixerAbort` guard (its comment said as much) because the only aborting fixer was `explicitOnly` and unreachable there; `dependabot` is neither. One refusal must not abandon the rest of the run.

## Verification

`vitest run` — 928 passed, 18 skipped; `tsc --noEmit`; `biome check src scripts` clean. New coverage: parser unit tests (browser-common's real rule, nested `update-types`, multi-ecosystem, quoted names, unnamed rules, `ignore: []`, commented-out blocks), the four `fix` paths above, and two `doctor` cases.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -95,6 +95,8 @@ npx @rtorcato/repo-tooling fix claude-skills --yes --json
 
 Fixers marked `explicitOnly` are exempt from `fix` all *and* from `fix --yes` — they only run when named as the target. Today that is `claude-skills`, the one fixer whose blast radius is outside the repo.
 
+A fixer may also **refuse** — the target file holds something the generator cannot reproduce, so overwriting would destroy it. Today that is `dependabot` against a config with repo-local `ignore:` rules (#422). A targeted `fix dependabot` then exits 1 with `error: dependabot-ignore-rules`; a bulk `fix --yes` records it `skipped` and carries on with the rest. Neither `--yes` nor `--json` overrides it — resolve the named rules by hand and re-run.
+
 ## Source-of-truth files in the repo
 
 - `src/cli/commands/setup.ts` — `ProjectConfig` interface and the setup orchestrator
--- a/apps/docs/docs/guides/dependabot-strategy.md
+++ b/apps/docs/docs/guides/dependabot-strategy.md
@@ -89,6 +89,21 @@ CI. Closing stale PRs is the normal, expected hygiene step — not a loss of wor
 - `repo-tooling doctor` flags drift from the canonical config; `repo-tooling fix`
   re-applies it. A strategy change propagates to every repo via `fix`.
 
+## 7. Repo-local `ignore:` rules
+
+`fix dependabot` writes the file wholesale, and the canonical config has no
+`ignore:` block — so a hand-added ignore rule (a dependency held back on purpose,
+usually with the reason in a comment above it) would be deleted by the next `fix`
+with nothing left to report the loss.
+
+It isn't. `fix dependabot` **refuses** when the existing config carries `ignore:`
+rules, naming each one, and `doctor` reports them in the Dependabot detail line so
+they're visible beforehand. Deliberate ignore rules are not drift: the check still
+reads `ok`.
+
+To proceed anyway, deal with the block by hand — copy it back into the regenerated
+file, or delete it to accept the loss — then re-run `fix dependabot`.
+
 ## Canonical `dependabot.yml`
 
 ```yaml
--- a/src/base/checks.ts
+++ b/src/base/checks.ts
@@ -3,6 +3,7 @@ import fs from 'fs-extra'
 import { BLOCK_START, hasStaleAgentBlock } from '../cli/generators/agent-rules.js'
 import { BADGE_START, hasPublicOnlyBadges } from '../cli/generators/badges.js'
 import { claudeSkillStatus, SHIPPED_SKILL } from '../cli/generators/claude-skills.js'
+import { DEPENDABOT_CONFIG_PATHS, dependabotIgnoreRules } from '../cli/generators/security.js'
 import { type DetectedLanguage, detectNestedLanguages } from '../cli/utils/detect-language.js'
 import type { CheckResult } from './types.js'
 
@@ -280,7 +281,7 @@ export async function checkGitHubActions(
 }
 
 export async function checkDependabot(dir: string): Promise<CheckResult> {
-	for (const candidate of ['.github/dependabot.yml', '.github/dependabot.yaml']) {
+	for (const candidate of DEPENDABOT_CONFIG_PATHS) {
 		const candidatePath = path.join(dir, candidate)
 		if (await fs.pathExists(candidatePath)) {
 			// The canonical standard (apps/docs/docs/guides/dependabot-strategy.md) is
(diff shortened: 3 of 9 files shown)


rtorcato (commented):

Reviewed against the diff and issue #422 only (no repo exploration).

**Detection is generic, not typescript-specific.** `dependabotIgnoreRules()` scans for any `^\s*ignore:\s*(#.*)?$` header and walks its indented block for `dependency-name:` values, regardless of ecosystem. Traced it by hand against the exact browser-common snippet quoted in the issue (nested comment lines, `update-types:` sub-list) — it correctly returns `['typescript']` and doesn't miscount the nested `- version-update:semver-major` line as a second rule. Test coverage backs this up across multiple ecosystems, quoted/unquoted names, `ignore: []`, and commented-out blocks.

**Holds under `fix --yes`.** Targeted `fix dependabot --yes` still exits 1 and leaves the file untouched (verified by `fix.test.ts`'s "refuses rather than drop" test, which explicitly passes `yes: true`). Bulk `fix --yes`/`fix` (no target) catches `FixerAbort` in the loop, records the fixer as `skipped`, and continues applying every other fixer rather than crashing or silently overwriting — matching the PR's stated behavior table and the issue's concern that an unattended run would "walk past" a mere warning.

**No obvious false positives.** The template never emits `ignore:`, so any detected block is by construction something the template can't reproduce; canonical output and `ignore: []` both correctly return no rules, so a repo the template can fully regenerate won't start refusing.

**Minor nit (non-blocking):** in `dependabotIgnoreRules`, when a block mixes named and unnamed entries (e.g. one rule with only `versions:`, another with `dependency-name:`), the padding logic (`while (names.length < items) names.push('<unnamed rule>')`) appends `<unnamed rule>` at the end regardless of which item actually lacked a name, rather than tracking the association per-item. The rule *count* stays correct, so the refusal still fires — this only risks a misleading label in the message naming which specific rule couldn't be read. No test exercises this mixed case.

Everything else — doctor's "not drift" framing, the `--diff` preview swallowing the shadow-run's `FixerAbort`, the AGENTS.md/dependabot-strategy.md doc updates — reads correctly and consistently with the pinned Option 3 decision.

### Before merging
- Nit only: `dependabotIgnoreRules` can mislabel *which* rule is `<unnamed rule>` when a block mixes named and unnamed entries — the refusal still correctly fires, only the rule name in the message could be wrong. Not blocking.

rtorcato (commented):

**Summary: 0 critical, 0 high, 1 low, 0 info — no criticals.**

Scope: `src/cli/generators/security.ts` (`dependabotIgnoreRules`, `findDependabotIgnoreRules`), `src/base/fixers.ts` / `src/base/checks.ts` / `src/cli/commands/fix.ts` wiring, and their tests.

## Low

**`dependabotIgnoreRules` is a total function (never throws), but its line-scanner heuristic can silently miss an `ignore:` block on flow-style or oddly-indented YAML — which is the exact failure mode this PR exists to close.**

- `src/cli/generators/security.ts` — the header regex `^(\s*)ignore:\s*(#.*)?$` only matches block-style `ignore:` followed by nothing (or a comment) on the same line. A rule written in flow style, e.g. `ignore: [{dependency-name: typescript}]`, never matches the header regex, so `dependabotIgnoreRules` returns `[]` for it and `fix dependabot` proceeds to overwrite — the same silent-deletion bug #422 reports, just reached by a different YAML dialect.
- Similarly, indentation is compared by raw character count (`header[1].length` vs. `next.search(/\S/)`), so a header indented with a tab and items indented with spaces (or vice versa) could make `indent <= blockIndent` trigger on the first item line, causing the block to be treated as empty.
- Neither case is a crash/hang — no ReDoS-prone patterns here (all character classes are linear), and there's no code path where a malformed string throws — so there's no unhandled-exception risk propagating past the fixer. The risk is purely the detector under-counting, not misbehaving.
- Mitigating factors: this operates on the repo's own `.github/dependabot.yml`, not attacker-controlled/untrusted network input; the tool's own generator always emits block style; and the gap is disclosed in-code via the `ponytail:` comment as a known ceiling ("reach for a parser if the message ever has to reproduce the rules rather than name them"). Real-world dependabot configs are essentially always block-style, matching what GitHub's own docs and this repo's canonical template emit.

No findings on injection, secrets, shell construction, or dependencies — no `child_process`/`exec`, no path built from parsed file content (candidate paths are the two hardcoded `DEPENDABOT_CONFIG_PATHS` strings), and no new dependencies (the PR deliberately avoids adding a YAML parser). The only thing that reaches an error message or `--json` payload is the extracted `dependency-name` values themselves, which is the intended, non-sensitive output.

### Before merging
- Low-severity, disclosed gap: flow-style `ignore: [...]` YAML (single-line list) bypasses `dependabotIgnoreRules` detection entirely and would still be silently overwritten. Worth a follow-up if any repo in the fleet uses that style — otherwise the existing `ponytail:` ceiling comment already covers it.

PR #447, opened 2026-08-20
fix(security): gate Dependabot auto-merge on the dev-minor group

### LINKED ISSUE #423: Scaffolded automerge merges production dependency bumps unreviewed
`generateDependabotConfig()` writes two files as a deliberate pair (`src/cli/generators/security.ts`):

- `.github/dependabot.yml` — groups bumps into `production-minor`, `dev-minor`, `major-updates`
- `.github/workflows/dependabot-automerge.yml` — merges on green CI

The workflow's only gate is the semver type:

```yaml
      - name: Auto-merge patch and minor updates
        if: |
          steps.metadata.outputs.update-type == 'version-update:semver-patch' ||
          steps.metadata.outputs.update-type == 'version-update:semver-minor'
        run: gh pr merge --auto --squash "$PR_URL"
```

Majors are correctly excluded. But `production-minor` is `dependency-type: production` — so **minor/patch bumps of dependencies that ship to consumers merge with no human or agent review**, on CI alone.

## Confirmed working exactly as designed

In `rtorcato/browser-common`, minutes after the grouped config merged:

```
$ gh pr view 139 --json mergedBy,autoMergeRequest
139 merged by: app/github-actions at 2026-08-18T19:15:08Z | autoMerge: was armed

$ gh run list --workflow "Dependabot auto-merge"
19:13:56  dependabot/npm_and_yarn/production-minor-7  completed/success
```

`react`, `react-dom` and `@easyops-cn/docusaurus-search-local`, opened and merged to `main` in about two minutes. No review, agent or human.

That instance was harmless — those are dependencies of `apps/docs`, which is `private: true`, and browser-common's published package ships zero runtime dependencies.

## Where it actually bites: js-common

`rtorcato/js-common` has the identical `dependabot.yml` grouping and the identical automerge workflow, and it is **not** private:

```
deps:     pino, short-uuid, uuid, zod
optional: @inquirer/prompts, chalk, chalk-animation, commander,
          figlet, gradient-string, pino-pretty
```

npm installs `optionalDependencies` for consumers too, so all eleven of those reach every consumer of the published package. Under the current pairing, a minor bump to any of them auto-merges unreviewed and ships in the next release.

That is a meaningfully different risk from auto-merging dev tooling: it is the supply chain of a published package, and the 7-day `cooldown` is the only thing standing between a compromised release and `main`.

## Suggestion

Narrow the automerge gate so it does not cover production dependencies of a publishable package. Options, roughly in order of effort:

- Restrict the workflow to the `dev-minor` group only (`steps.metadata.outputs.dependency-group`), leaving `production-minor` for manual triage.
- Keep auto-merging production bumps **only** where the manifest is `private: true`, which preserves today's convenience for docs/app workspaces.
- Make it a `.repo-tooling.json` switch, defaulting to the safe behaviour, so publishing repos and app repos can differ deliberately.

Worth noting the `cooldown: default-days: 7` already does real work here and should stay whatever else changes — it is what prevents a same-day malicious release from being merged automatically.

---

rtorcato:
🤖 *Automated — implementer via ai-issue-loop. Posted under the owner's account by an agent; not a human message.*

Closes #423

## What changed

`generateDependabotConfig()`'s auto-merge workflow gated only on the semver type, so minor/patch bumps of **runtime** dependencies merged to `main` on green CI alone. For a published package that is its supply chain, and the 7-day `cooldown` was the only thing in the way.

```diff
-      - name: Auto-merge patch and minor updates
+      - name: Auto-merge dev-dependency patch and minor updates
         if: |
-          steps.metadata.outputs.update-type == 'version-update:semver-patch' ||
-          steps.metadata.outputs.update-type == 'version-update:semver-minor'
+          steps.metadata.outputs.dependency-group == 'dev-minor' &&
+          (steps.metadata.outputs.update-type == 'version-update:semver-patch' ||
+          steps.metadata.outputs.update-type == 'version-update:semver-minor')
```

Applied to the generator, to this repo's own copies (dogfood), and to the strategy guide. No group was renamed — the pairing with `dependabot.yml` is unchanged.

## Decisions worth reviewing

**Both conditions, not group alone.** `dev-minor` is already declared `minor`+`patch` in the `dependabot.yml` this generator writes, so a group-only gate would be sufficient *given that pairing*. Kept both anyway: this workflow is the security gate, and the file it would be trusting lives in the consumer's repo where it can be edited independently of the workflow. Two extra lines to stop trusting a config we don't control is the right trade on a gate whose failure mode is an unreviewed publish.

**The gate fails closed.** A PR outside any group reports an empty `dependency-group` and never matches. That was the stated reason dev-only was pinned over private-only, and it holds here.

**GitHub Actions bumps now need a human too.** The `github-actions` ecosystem block has no `groups:`, so those PRs report an empty `dependency-group` and stop auto-merging. This is a narrowing beyond production npm deps and was not called out in the issue — flagging it explicitly. I think it's correct rather than collateral: a compromised action runs in CI holding `GITHUB_TOKEN`, which is not obviously a smaller blast radius than a runtime dep. If you'd rather keep them flowing, the carve-out is one `||` on `package-ecosystem`. Docs table updated to say ❌ manual either way.

**`doctor` — extended the existing check, not a new one.** The suggestion was a separate clearly-named check; I put it in `checkDependabot` as a named delta instead:

```
Dependabot  drift  .github/dependabot.yml drifts from canonical
                   (auto-merge workflow merges production bumps (missing dev-minor group gate))
```

Reason: a second check would emit a second `drift` row for the same file pair with the same remedy (`fix dependabot` rewrites both). Happy to split it if you'd rather see it standalone.

## Release type

`fix:` → patch. This **narrows behaviour consumers currently rely on** — any repo that takes the update stops auto-merging production bumps it previously merged. I didn't mark it breaking because nothing in the public surface moves: same CLI flags, same `ProjectConfig` schema, same `filesWritten` list. Only the *content* of two scaffolded files, which `fix dependabot` exists to rewrite. If you read a scaffolded-output change of this size as major-worthy, say so and I'll redo the subject as `feat!:`.

Also worth noting the issue thread: your comment said "dev-only and private-only", and the triage comment that lifted the hold pinned **dev-only** only. This PR implements dev-only. Private-only is not foreclosed and would layer on top.

## Verification

`biome check src scripts`, `tsc --noEmit`, and the full vitest suite (918 passed, 18 skipped) all clean. Three new tests: the gate's shape, that the gated group name is one the paired `dependabot.yml` actually declares, and the new `doctor` drift row.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -17,13 +17,16 @@ updates:
       prefix: chore
       include: scope
     groups:
-      # Safe tier: runtime + dev minor/patch auto-merge on green (see the
-      # dependabot-automerge workflow). Grouped so react/react-dom move together.
+      # Runtime minor/patch. Grouped so react/react-dom move together, but NOT
+      # auto-merged — these ship to consumers of a published package, so they
+      # get a human (see the dependabot-automerge workflow).
       production-minor:
         dependency-type: production
         update-types:
           - minor
           - patch
+      # The only tier that auto-merges on green — dev tooling never reaches a
+      # consumer of the published package.
       dev-minor:
         dependency-type: development
         update-types:
--- a/.github/workflows/dependabot-automerge.yml
+++ b/.github/workflows/dependabot-automerge.yml
@@ -17,10 +17,14 @@ jobs:
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Auto-merge patch and minor updates
+      # Belt and braces: the dev-minor group is already declared minor+patch in
+      # dependabot.yml, but this workflow is the security gate and shouldn't
+      # trust a config file a consumer repo can edit independently.
+      - name: Auto-merge dev-dependency patch and minor updates
         if: |
-          steps.metadata.outputs.update-type == 'version-update:semver-patch' ||
-          steps.metadata.outputs.update-type == 'version-update:semver-minor'
+          steps.metadata.outputs.dependency-group == 'dev-minor' &&
+          (steps.metadata.outputs.update-type == 'version-update:semver-patch' ||
+          steps.metadata.outputs.update-type == 'version-update:semver-minor')
         run: gh pr merge --auto --squash "$PR_URL"
         env:
           PR_URL: ${{ github.event.pull_request.html_url }}
--- a/apps/docs/docs/guides/dependabot-strategy.md
+++ b/apps/docs/docs/guides/dependabot-strategy.md
@@ -25,17 +25,32 @@ Dependabot opens roughly **3–4 PRs per cycle** instead of one per package:
 
 | Group | Contents | Update types | Auto-merge |
 | --- | --- | --- | --- |
-| `production-minor` | runtime `dependencies` | minor, patch | ✅ on green |
+| `production-minor` | runtime `dependencies` | minor, patch | ❌ manual |
 | `dev-minor` | `devDependencies` | minor, patch | ✅ on green |
 | `major-updates` | all packages | major | ❌ manual |
-| `github-actions` | workflow actions | all | ✅ on green |
+| `github-actions` | workflow actions | all | ❌ manual |
 
-## 2. Auto-merge — safe tier only
+## 2. Auto-merge — dev dependencies only
 
-Patch and minor updates (prod and dev) **merge themselves once CI is green** —
-no human in the loop. Implemented by `.github/workflows/dependabot-automerge.yml`
-using `dependabot/fetch-metadata` + `gh pr merge --auto --squash`, gated to
-`version-update:semver-patch` and `version-update:semver-minor`.
+Patch and minor updates **of `devDependencies`** merge themselves once CI is
+green — no human in the loop. Implemented by
+`.github/workflows/dependabot-automerge.yml` using `dependabot/fetch-metadata` +
+`gh pr merge --auto --squash`, gated on **both** `dependency-group == 'dev-minor'`
+and `version-update:semver-patch` / `version-update:semver-minor`.
+
+Everything else waits for a human, and the gate **fails closed**: a PR outside a
+group reports an empty `dependency-group` and so never matches.
+
+> **Production bumps are deliberately excluded.** A minor bump to a runtime
+> dependency of a published package ships to every consumer on the next release.
+> Auto-merging those on green CI alone means the 7-day `cooldown` is the only
+> thing between a compromised upstream release and `main`. GitHub Actions bumps
+> are excluded for the same reason — a compromised action runs in CI holding
+> `GITHUB_TOKEN`.
+>
+> The gate names the group that `dependabot.yml` declares, so the two files are
+> a paired unit: rename a group in one and auto-merge silently stops (which is
+> the safe direction, but still drift — `doctor` flags it).
 
 > **Requires branch protection.** `gh pr merge --auto` only gates correctly when
 > the repo has auto-merge enabled and `main` has required status checks
--- a/src/base/checks.ts
+++ b/src/base/checks.ts
@@ -295,7 +295,15 @@ export async function checkDependabot(dir: string): Promise<CheckResult> {
 				}
 			}
 			const automergePath = path.join(dir, '.github', 'workflows', 'dependabot-automerge.yml')
-			if (!(await fs.pathExists(automergePath))) {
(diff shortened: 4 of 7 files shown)


rtorcato (commented):

Reviewed the diff-scoped changes: `.github/dependabot.yml`, `.github/workflows/dependabot-automerge.yml`, `apps/docs/docs/guides/dependabot-strategy.md`, `src/base/checks.ts`, `src/cli/generators/security.ts`, and tests.

**1. Fails closed.** The new condition is `steps.metadata.outputs.dependency-group == 'dev-minor' && (...semver-patch || ...semver-minor)`. GitHub Actions `if:` expressions treat an empty/missing/unexpected `dependency-group` output as simply not equal to `'dev-minor'`, so the whole `&&` short-circuits false and the merge step never runs — no separate `!= ''` check needed. A Dependabot PR outside any group (or a group renamed later) correctly stops auto-merging rather than falling through.

**2. Majors exclusion intact.** `semver-major` never matches either arm of the `update-type` disjunction, so it's still excluded — the new group check is additive (`&&`), not a replacement for the semver-type gate.

**3. No new injection or bypass surface.** Both `dependency-group` and `update-type` come from `dependabot/fetch-metadata`'s parsed output, referenced only inside a YAML `if:` expression (not shell-interpolated), so there's no command-injection vector here. The `run:` line still passes the PR URL via `"$PR_URL"` env var, unchanged from before. Nothing in this diff lets PR-supplied content reach a shell string or otherwise influence its own merge decision beyond what already existed pre-PR (trigger/actor gating is outside this diff).

Belt-and-braces choice to gate on both the group name and the semver type (rather than trusting `dependabot.yml`'s group definition alone) is sound defense-in-depth since that file lives in a consumer repo. The `doctor` drift check and generator tests correctly assert the new gate string is present, so regressions to the old broad gate will be caught. No secrets, no new dependencies, no unsafe shell construction introduced.

### Before merging
Nothing.

rtorcato (commented):

**Scope:** read `gh pr view 447`, `gh pr diff 447`, and issue #423 only.

**Pairing (dependabot.yml ↔ workflow) — matches.** The workflow gates on `steps.metadata.outputs.dependency-group == 'dev-minor'`, and `dependabot.yml`'s group key is `dev-minor:` — no typo, no rename. A test (`tests/cli/generators/security.test.ts`) even extracts the group name from the workflow via regex and asserts `dependabot.yml` declares it, which is a good guard against this exact class of drift in future edits.

**Hole is closed correctly.** `production-minor` no longer matches (wrong group), majors still never matched (semver-major was never in the OR list, unchanged), and the semver-type condition was kept alongside the new group condition rather than replacing it — so an ungrouped or misclassified PR still can't slip through on group alone.

**Fails closed.** An ungrouped PR (e.g. `github-actions`, which has no `groups:` block) reports an empty `dependency-group` and never matches `'dev-minor'`. Consistent with the dev-only decision already pinned for this issue.

**Old-pairing consumer repos degrade safely.** A repo that pulls only the new workflow keeps its existing `dev-minor` group (that group already existed pre-#423, only its comment changed), so the new gate still resolves correctly against an old `dependabot.yml`. A repo that keeps the *old* workflow with an updated `dependabot.yml` is still exposed until `fix dependabot` runs — but the new `checkDependabot` drift check (`src/base/checks.ts`) now detects exactly that gap by scanning the workflow content for the `dev-minor` gate string, and `fix dependabot` rewrites both files together. The PR body doesn't spell this consumer-drift scenario out explicitly, but the `doctor` change is the actual mechanism that closes it, and it's covered by a new test (`tests/cli/commands/doctor.test.ts`).

**Everything else checks out:** YAML `if:` block preserves the pre-existing multi-line literal-block pattern (not a new syntax risk), and the `checks.ts` if/else restructure doesn't introduce a double-push or fallthrough bug from what's visible in the diff context.

Two things the PR author already surfaces transparently as open questions rather than defects — noted below for the human merge decision, not blocking.

### Before merging
- `fix:` vs `feat!:` for the commit type: this narrows auto-merge behaviour consumers currently rely on (production bumps that merge today will stop). Nothing in the public CLI surface changes, so `fix:` is defensible, but worth a deliberate call given semantic-release will cut a patch either way.
- The `github-actions` ecosystem also loses auto-merge as a side effect of the group-only gate (empty `dependency-group`), which is broader than what #423 asked for. The author flags this explicitly and offers a one-line carve-out if that's unwanted — worth a yes/no before merge.

PR #448, opened 2026-08-20
feat(doctor): tell a stale copied asset from a deliberate local fork

### LINKED ISSUE #428: doctor: detect stale copied assets — sync-changelog.mjs exists in 6 different versions across the family
`repo-tooling copy <asset>` is a one-shot: it writes the file and nothing ever checks it again. Consumers drift silently, and the drift is invisible until someone diffs by hand.

## Measured drift

`scripts/sync-changelog.mjs` across the family, by git blob SHA:

| Repo | Size | Blob | |
|---|---|---|---|
| repo-tooling `tooling/docusaurus/` (canonical) | 4982 | `ed2bcaf` | source |
| repo-tooling `scripts/` | 4982 | `ed2bcaf` | dogfooded ✅ |
| browser-common | 4982 | `ed2bcaf` | ✅ in sync |
| js-common | 3863 | `541ae2d` | deliberate fork, now obsolete |
| cf-common | 1746 | `133e9c4` | 
| swift-common | 1746 | `133e9c4` | 
| supabase-common | 1456 | `6683ad5` | 
| react-common | 1223 | `5e4f45c` | 
| db-common | 1007 | `f17aaf0` | 

**8 copies, 6 distinct versions. One consumer is current.**

## What the drift does and does not mean

The five smaller copies are the pre-Releases generation — they copy `CHANGELOG.md` into the docs page verbatim, rather than building it from GitHub Releases.

That is only a *bug* where `CHANGELOG.md` has stopped being updated. Per the canonical asset's header, the shipped semantic-release preset drops `@semantic-release/git` because the `code-scanning-main` ruleset rejects the release commit with GH013 (#417) — and on those repos nothing commits `CHANGELOG.md` back, so it freezes at the last release that ran with the plugin. It freezes at a *real* release, so the page reads as current and nobody notices.

**I checked, and that is not currently happening on the drifted repos.** `cf-common`, `db-common` and `browser-common` still use `@semantic-release/git`, and their `CHANGELOG.md` top entry matches their latest release (`1.5.1`, `1.4.1`). `swift-common`, `react-common` and `supabase-common` have no root `release.config.mjs` and a different layout, so the old asset is not obviously wrong for them either.

So this issue is about the **mechanism**, not a live outage: the five stale copies are fine today and become wrong the moment their repo moves onto the GH013-affected preset, with no signal when that happens. That transition is the failure this check should catch.

## Ask

A `doctor` check that compares each copied asset against the version currently shipped by the installed `@rtorcato/repo-tooling` and reports drift as `stale` (re-copy) vs `modified` (locally edited — say so and leave it).

Distinguishing the two is the part that matters. A local edit is sometimes correct: js-common forked this file deliberately and documented why in the header. A check that can't tell a fork from a stale copy will either nag about the fork forever or stay quiet about the stale ones.

Options, roughly in order of effort:
- record the source version in a generated header comment on copy, and compare that — cheap, but a local edit that keeps the header lies
- hash the pristine asset at copy time into a lockfile (`.repo-tooling-assets.json`); drift from that hash = modified, hash matches but differs from installed = stale. Distinguishes both cases correctly
- content hash against every published version of the asset — no extra state, but needs history

`fix` could then re-copy the `stale` ones. The `modified` ones stay a human decision.

## Related

- #427 — `sync-agents.mjs`, three hand-maintained copies with no asset at all. Centralising it without drift detection just moves where the drift starts

rtorcato:
🤖 *Automated — implementer via ai-issue-loop. Posted under the owner's account by an agent; not a human message.*

Closes #428

Implements the lockfile-hash option pinned in triage — the only one of the three that classifies a deliberate fork correctly.

## What changed

`copy <preset>` now records the asset's pristine sha256 in `.repo-tooling.json` under a new `assets` map. `doctor`'s **Copied assets** check compares three hashes and reports:

| State | Condition | doctor status |
|---|---|---|
| `stale` | file matches the record, shipped asset differs | `drift` — `fix copied-assets` re-copies |
| `modified` | file differs from the record | `ok`, named in the detail, left alone |
| untracked | no recorded hash | `ok`, named in the detail, never drift |
| `ok` | file matches the record and the shipped asset | `ok` |

`js-common`'s deliberate fork of `sync-changelog.mjs` lands in `modified` and stays there — it is reported once per run and never nagged as fixable.

## No-record behaviour

Four family repos have no `.repo-tooling.json`, and every copy made before this lands has no recorded hash. Both report as untracked, which is `ok` and cannot fail a build. `copy` deliberately does **not** create a lockfile as a side effect — `recordAssetHash` returns `false` and the copy still succeeds.

## Lockfile migration

`LOCKFILE_VERSION` 2 → 3 through the existing `migrate()`: a v2 file gains `assets: {}` on read, everything else untouched. `writeLockfile` gained an optional third argument and otherwise carries existing hashes forward, so `updateLockfileConfig` (called from a dozen fixers) can't silently drop them.

## The one risk worth naming

Bumping the version means an *older* CLI reading a v3 lockfile reports the existing `lockfile` check as `drift` ("newer than this CLI supports"). That is the designed behaviour of the version field rather than a regression, and the fix is the upgrade the message already suggests.

## Tests

`tests/cli/utils/copied-assets.test.ts` covers all four states plus the two no-record shapes (record stripped, and no lockfile at all), that untracked never fails a build, and that the check reaches `runDoctor`. `tests/cli/utils/lockfile.test.ts` covers the v2→v3 migration, hash preservation across `updateLockfileConfig`, and that `recordAssetHash` creates no file.

947 tests pass; `biome check` and `tsc --noEmit` clean.

### DIFF
--- a/apps/docs/docs/guides/cli.md
+++ b/apps/docs/docs/guides/cli.md
@@ -33,6 +33,18 @@ Available presets: `biome`, `tsconfig`, `bun`, `nx`, `changesets`, `release-plea
 
 `copy` is for configs you must *own* rather than extend — Biome doesn't support configuration extension, and TypeScript configs resolve more reliably when local. Most other configs (ESLint, Prettier, Vitest, etc.) can be imported or extended directly — see the [Configuration Reference](../reference/biome.md) for usage.
 
+### Copied assets don't drift silently
+
+A copy used to be a one-shot: nothing looked at the file again, so a repo could sit on an ancient copy of `scripts/sync-changelog.mjs` for a year without a signal. `copy` now records the asset's hash at copy time in `.repo-tooling.json`, and doctor's **Copied assets** check reads it back:
+
+| State | Means | doctor |
+|---|---|---|
+| `stale` | Untouched since the copy, but the package ships a newer version | `drift` — `fix copied-assets` re-copies it, and can't lose anything, because nothing local is in it |
+| `modified` | The file differs from what was copied — someone forked it deliberately | reported, never overwritten; a human decision |
+| untracked | No recorded hash: the copy predates this, or the repo has no `.repo-tooling.json` | reported, never `drift` — a missing record isn't evidence of drift |
+
+`copy` never creates a lockfile just to have somewhere to write. Run `fix lockfile` first if you want a repo's copies tracked.
+
 ## list / ls
 
 Prints all available tooling configurations.
--- a/src/base/fixers.ts
+++ b/src/base/fixers.ts
@@ -30,6 +30,7 @@ import {
 	generateDependabotConfig,
 	generateRenovateConfig,
 } from '../cli/generators/security.js'
+import { classifyCopiedAssets } from '../cli/utils/copied-assets.js'
 import { copyPreset } from '../cli/utils/copy-preset.js'
 import { detectLanguage } from '../cli/utils/detect-language.js'
 import type { Lockfile } from '../cli/utils/lockfile.js'
@@ -139,6 +140,25 @@ async function resolveInstallDir(explicit: string | undefined, assumeYes: boolea
 }
 
 export const BASE_FIXERS: Fixer[] = [
+	{
+		target: 'copied-assets',
+		description:
+			'Re-copy presets that are unchanged since they were copied but older than the version this package ships',
+		appliesTo: ['Copied assets'],
+		outputs: ['(the stale copied presets, re-copied in place)'],
+		canFixDrift: true,
+		async run({ targetDir }) {
+			// Only the `stale` ones: their content provably still matches what was
+			// copied, so overwriting loses nothing. `modified` assets are somebody's
+			// deliberate fork and stay a human decision (#428).
+			const stale = (await classifyCopiedAssets(targetDir)).filter((a) => a.state === 'stale')
+			const filesWritten: string[] = []
+			for (const asset of stale) {
+				filesWritten.push((await copyPreset(asset.preset, targetDir)).target)
+			}
+			return { filesWritten }
+		},
+	},
 	{
 		target: 'editorconfig',
 		description: 'Scaffold .editorconfig (UTF-8, LF, tab indent)',
--- a/src/cli/commands/doctor.ts
+++ b/src/cli/commands/doctor.ts
@@ -15,6 +15,7 @@ import { type DetectedLanguage, detectLanguage } from '../utils/detect-language.
 import { checkGitHubSettings } from '../../base/github-settings.js'
 import { checkMilestones } from '../../base/milestones.js'
 import { checkGitIdentity } from '../../base/git-identity.js'
+import { checkCopiedAssets } from '../utils/copied-assets.js'
 import { type Lockfile, LOCKFILE_VERSION, readLockfile } from '../utils/lockfile.js'
 import { declinedInLock, getFixTargetForCheck } from './fix-targets.js'
 import {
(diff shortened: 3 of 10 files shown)


rtorcato (commented):

Reviewed the diff for #428 (copied-asset drift detection).

**Hash choice/use**: full sha256 hex digests, compared with strict `===` (`recorded !== current`, `shipped !== recorded`). No truncation, no loose/type-coercing comparison, no comparison of formatted/derived strings — this is sound for a change detector.

**What gets recorded**: `.repo-tooling.json` gains `assets: { <presetName>: <sha256 hex> }`. Preset names come from the static `PresetName` enum, not user input, and the value is just a hex digest — no absolute paths, usernames, or other machine-local data land in the committed file.

**Path handling**: `preset.target`/`preset.source` are resolved from the hardcoded `PRESETS` map (existing code, unchanged here), never from external/asset-name input reachable by an attacker. All reads go through `fs.readFile`/`hashFile`; nothing is passed to a shell (`node:crypto` + `fs-extra` only, no new deps, no `exec`/`spawn` anywhere in this diff).

**Failure direction**: checked the three "can't verify" paths —
- No recorded hash → `unknown`, explicitly listed by name in the `Copied assets` detail (never silently folded into the "in sync" count).
- File missing entirely → skipped from results (never fabricated as `ok`).
- Installed package's own source unreadable while a recorded hash exists → falls back to treating it as `ok` (comment: "no news, not drift"). This is a deliberate, documented default for a case that shouldn't happen (failing to read your own installed package), and it degrades to *not reporting drift* rather than fabricating a false positive — consistent with this being a change detector, not an attestation.

No secrets, no injection surface, no new dependencies/supply-chain change (stdlib `node:crypto` + already-used `fs-extra` only). Lockfile migration (v2→v3) only adds an empty `assets: {}` on read, doesn't touch existing fields.

### Before merging
Nothing.

rtorcato (commented):

Reviewed against the pinned lockfile-hash approach from #428.

**Classification logic (the core of the PR) is correct.** In `classifyCopiedAssets` (`src/cli/utils/copied-assets.ts`):
- `recorded !== current` → `modified` (file diverges from the hash taken at copy time — this is exactly the js-common-fork case, and it correctly stays `modified` forever rather than getting nagged or silently re-copied).
- `recorded === current && shipped !== recorded` → `stale` (untouched locally, package moved on — the only state the `copied-assets` fixer re-copies).
- No `recorded` entry → `unknown`, which maps to doctor status `ok`, never `drift`. Confirmed this can't fail a build — `checkCopiedAssets` only returns `status: 'drift'` when `stale.length > 0`, `unknown` assets just get listed in the detail string.

**No-record / no-lockfile case is handled deliberately, not accidentally.** `recordAssetHash` (`src/cli/utils/lockfile.ts`) returns `false` and creates no file when the repo has no `.repo-tooling.json`, and `copyPreset` doesn't fail when that happens (`if (hash) await recordAssetHash(...)` — return value is discarded, copy still succeeds). Matches the "four family repos have no lockfile" requirement.

**Lockfile migration follows the existing pattern.** `LOCKFILE_VERSION` 2 → 3, `migrate()` only touches files below the current version and adds `assets: lock.assets ?? {}` — same shape as the v1→v2 migration it sits next to. `writeLockfile`'s new optional `assets` param defaults to carrying the existing file's hashes forward via `readLockfile`, and there's a direct test (`leaves recorded asset hashes alone (#428)`) proving `updateLockfileConfig` — called from a dozen fixers — doesn't silently drop them.

**One open question, not a defect:** `hashFile` hashes raw bytes (`fs.readFile` with no encoding, straight into `createHash('sha256')`), and nothing in this PR normalizes line endings before hashing. If a target repo has `core.autocrlf` on (or a `.gitattributes` that rewrites the copied asset's line endings on checkout), an untouched copy could hash differently than the value recorded at copy time and get misclassified as `modified` instead of `ok`/`stale`. This doesn't fail a build (`modified` is doctor status `ok`), and I can't tell from the diff alone whether any target repo in this family actually has autocrlf enabled — it's a question worth a maintainer's judgment call on whether it's worth guarding against, not something I can resolve from the diff.

Nothing else stood out — `copy` only stamping a hash it can actually compute, the fixer only touching `stale` (never `modified`) via a sequential loop (no race on the lockfile write), and the doc table in `cli.md` all match the implementation.

### Before merging
- Confirm whether any consuming repo in the family uses `core.autocrlf`/line-ending-rewriting `.gitattributes` on the asset files this hashes — if so, an unmodified copy could misreport as `modified` on checkout. Not a blocker; `modified` never fails a build, it just under-reports `stale`/`ok`.

Issue #429, opened 2026-08-20
Scaffold and check the release environment gate — only 1 of 10 repos has one


rtorcato:
`repo-tooling` scaffolds squash + auto-merge + required status checks + `required_pull_request_reviews: null`, but nothing scaffolds a gate between **merging `main`** and **publishing to npm**. On the shipped semantic-release preset those are the same event.

## Survey

`release` environment with `required_reviewers`, across the family:

| Repo | Gate |
|---|---|
| js-common | ✅ `required_reviewers` |
| browser-common, cf-common, db-common, react-common, supabase-common, api-common, swift-common, repo-tooling, shared-docs | ❌ none |

**1 of 10.** On the other nine, anything that lands on `main` publishes. js-common has it because it was set up by hand, not because the tooling produced it — so it is one `doctor`/`fix` run or one repo re-bootstrap away from being lost, with no signal.

## Ask

1. **Scaffold it.** Create the `release` environment with `required_reviewers` as part of the same bootstrap that sets branch protection, and wire the release workflow job to `environment: release`. A merge to `main` should leave the pipeline `waiting`, not publish.
2. **Check it.** A `doctor` rule that reports a repo whose release workflow publishes to npm with no environment gate — and, separately, one that has the environment but whose release job does not reference it, which looks protected and is not.

The second is the failure mode worth catching. An environment nothing references is decorative, and it reads as a gate in the GitHub UI.

## Why now

This is the precondition for letting `ai-issue-loop` auto-merge anything beyond Dependabot (rtorcato/dotfiles!N — GitLab). The loop's rule today is *"only Dependabot PRs merge unattended"*, and the stated reason is that merging `main` fires semantic-release and publishes.

With this gate in place that reason no longer holds: a human still stands between the merge and npm, so an unattended merge costs a revert at worst. Without it the loop's rule is exactly right and should stay.

The loop-side change is designed to fail closed — it will probe for `required_reviewers` on the `release` environment and hand the PR over when the probe comes back empty:

```bash
gh api repos/OWNER/REPO/environments \
  --jq '[.environments[] | select(.name=="release")
         | .protection_rules[]? | select(.type=="required_reviewers")] | length'
```

So shipping this changes nothing on its own; it just makes a repo *eligible*. Which is the right order — the gate should exist because publishing unreviewed is bad, not because an agent wants to merge.

## Related

- #427, #428 — the other two asset/drift issues from the same sweep

rtorcato:
[...]

PR #449, opened 2026-08-20
feat(doctor): check the release environment gate

rtorcato:

Implements the **checks half** of #429. The scaffolding half — creating the `release` environment and wiring `environment: release` into the release workflow — stays open, pending a decision on who the required reviewers are. `Refs #429`, deliberately not `Closes`.

## What lands

Two checks in `checkGitHubSettings`, alongside branch protection / merge settings / workflow permissions / code-scanning:

| Check | Drifts when |
|---|---|
| **Release gate** | the publishing job runs behind no `environment:` and the repo has no `release` environment; or it names an environment the repo doesn't have (Actions creates it unprotected on first run); or the named environment has no `required_reviewers` |
| **Release environment** | the repo **has** a `release` environment and no job references it — the failure mode the issue calls out: decorative, and it reads as a gate in the GitHub UI |

The two never both fire on one repo — "no `environment:` and no `release` environment" belongs to the first, "no `environment:` but the environment exists" to the second, and the first defers. One root cause, one finding.

## Statuses

- A repo that publishes nothing, or a private package → `ok`, `not applicable — no workflow job publishes to a registry`. Not `drift`; nothing is broken.
- gh missing / unauthenticated / no GitHub remote → `ok` skipped, with the other four, via the existing `skipAll`.
- The environments endpoint unreadable (403) → `ok` skipped. A 404 is read as "no environments", not as unreadable.

## Job-scoped, not file-scoped

The publishing job is found by splitting the workflow's `jobs:` blocks. A whole-file grep for `environment:` reports any repo with a `github-pages` deploy job as gated — the exact false negative this check exists to prevent. There's a test for it.

Dogfooding the check against this repo's own `ci.yml` caught a second one: a `varcheck` job was reported as the publisher because its shell block ends with `# npm publish uses OIDC trusted publishing`. Comment lines are now stripped before matching, which also stops a commented-out `environment:` reading as a live gate. Both cases are regression-tested.

Hand-split rather than parsed because the package ships no YAML dependency and the only question asked is where one job's keys begin and end.

## No fixer

Creating the environment needs a `required_reviewers` list only a human can supply, so both checks report and their hints describe the manual step. The three `base + <lang> fixers cover every check doctor emits` tests record them as intentionally uncovered.

## Please look at this bit

`Release gate` reports **drift on this repo** — true, and precisely the point. But CI's dogfood step runs doctor against this repo with `GH_TOKEN: ${{ github.token }}`, so the finding would turn CI red. I added `'Release gate'` to `ACCEPTED` in `scripts/dogfood.mjs`, with a comment saying it is *not* a self-repo divergence and that the line should be deleted the moment the `release` environment exists. That is a judgement call and the one thing here worth a second opinion — the alternative was knowingly opening a red PR.

## Release impact

`feat:` → minor. Honest type: repos that were previously clean will now report drift, which is a visible behaviour change, not a fix.

## Verification

`vitest run` 947 passed / 59 files · `biome check src scripts` clean · `tsc --noEmit` clean · `pnpm dogfood` (built CLI, authenticated) clean.

## Explicitly out of scope

- Creating the `release` environment via the GitHub API, or wiring `environment: release` into any workflow.
- Any write to a repository other than this one.

### DIFF
--- a/scripts/dogfood.mjs
+++ b/scripts/dogfood.mjs
@@ -14,6 +14,12 @@ const ACCEPTED = new Set([
 	'Vitest', // vitest.config.mjs imports the local ./tooling source, not the export
 	'Commitlint', // commitlint.config.mjs re-exports the local ./tooling source
 	'semantic-release', // release.config.mjs extends the local ./tooling source
+	// NOT a self-repo divergence — a real finding this repo cannot close yet.
+	// The check landed (rtorcato/repo-tooling#429, checks half); creating the
+	// `release` environment is the scaffolding half, still open because it needs
+	// a `required_reviewers` list only a human can supply. Delete this line the
+	// moment that environment exists — that is the whole point of the check.
+	'Release gate',
 ])
 
 function runDoctor() {
--- a/src/base/github-settings.ts
+++ b/src/base/github-settings.ts
@@ -76,11 +76,15 @@ export const GITHUB_STANDARD = {
 } as const
 
 const CODE_SCANNING_CHECK = 'Code-scanning gate'
+const RELEASE_GATE_CHECK = 'Release gate'
+const RELEASE_ENV_CHECK = 'Release environment'
 const CHECK_NAMES = [
 	'Branch protection',
 	'Merge settings',
 	'Workflow permissions',
 	CODE_SCANNING_CHECK,
+	RELEASE_GATE_CHECK,
+	RELEASE_ENV_CHECK,
 ] as const
 
 // Recommended CodeQL alert thresholds — GitHub's UI defaults. This is the
@@ -186,6 +190,7 @@ export async function checkGitHubSettings(dir: string, exec?: GhExec): Promise<C
 		checkMergeSettings(info),
 		await checkWorkflowPermissions(gh, info.nwo),
 		await checkCodeScanningRuleset(gh, info.nwo, info.branch, dir),
+		...(await checkReleaseGate(gh, info.nwo, dir)),
 	]
 }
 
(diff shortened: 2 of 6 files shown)


rtorcato (commented):

**1 low/info, 0 high, 0 medium — no criticals.** No secrets leak, no shell/command injection, no new dependencies (hand-rolled job splitter avoids adding a YAML parser — the right call for this narrow use).

## Findings

**Low — `findPublishJob`'s catch-all can conflate a real read error with "nothing publishes" (`src/base/github-settings.ts`, `findPublishJob`)**

```ts
try {
    for (const f of (await fs.readdir(workflowsDir)).sort()) {
        ...
        const content = await fs.readFile(path.join(workflowsDir, f), 'utf-8')
        ...
    }
} catch {
    return null
}
```

`findPublishJob` returning `null` is read upstream as "not applicable — no workflow job publishes to a registry" → status `ok`. That's correct when `workflowsDir` genuinely doesn't exist (handled explicitly above the try). But the `catch` around `fs.readdir`/`fs.readFile` also swallows real I/O failures (permission errors, a broken symlink, an encoding issue) and reports them identically — `ok`, gate not needed — rather than surfacing them as "could not read workflows" the way `readEnvironments` already does for its own unreadable case (`skip`). That's the exact fail-open shape the task called out: an unreadable input should not resolve to `ok`.

*Exploit scenario:* low — this isn't attacker-reachable (workflow files are local repo content, not user/issue-controlled), so it's a robustness gap rather than something a hostile actor can trigger. Still worth tightening since the whole point of this check is not to silently report a gate that isn't there.

*Fix:* distinguish "directory absent" (already handled) from "directory present but unreadable/unparseable" — have the latter propagate to `skip(...)` in `checkReleaseGate` instead of falling through to "no publish job."

## Other things checked, no issues
- **Token/credential handling**: uses the existing `GhExec`/`gh api` abstraction already used by the sibling checks in this file; no new auth path. `r.stderr` from the environments call is only pattern-matched (`/404|not found/i`), never placed into a `detail`/hint string — no risk of a stderr-leaked token or error reaching the JSON report.
- **Fail-open on unreadable API data**: correctly conservative — only an explicit 404 is read as "no environments"; any other failure (403, 5xx, malformed JSON, unexpected response shape) resolves to `skip`, and `skip` on both checks, not a silent `ok`-as-compliant. Malformed/unexpected `environments` payload shape defaults to an empty list, which pushes the *gate* check toward `drift` (over-reporting), not `ok` (under-reporting) — the safe direction.
- **YAML "parsing"**: hand-split, regex-based, no catastrophic-backtracking patterns; operates only on local workflow file content, never reaches a shell exec or a path join.
- **Dogfood ACCEPTED entry**: adding `'Release gate'` to `scripts/dogfood.mjs`'s `ACCEPTED` set is a documented, reasoned exception (with a note on when to remove it), not a case of suppressing a real security control.

### Before merging
- `findPublishJob`'s catch-all in `src/base/github-settings.ts` should not report `ok`/"not applicable" for a genuine directory-read failure — route it through the same `skip` path used when the environments API is unreadable. Low severity, not blocking.

rtorcato (commented):

Scope check: the PR body correctly says `Refs #429`, not `Closes #429` — the triage comment on #429 explicitly splits the scaffolding half (creating the `release` environment, wiring `environment: release`) out as still-held, so leaving the issue open is right.

**Second check (the point of the PR).** Verified it detects the decorative-gate case, not just absence: `checkReleaseGate` computes `hasRelease = envs.has('release')` independently of what the publish job references. The test `flags a 'release' environment that no job references (#429, the decorative gate)` exercises exactly this — a repo with a `release` environment (with required reviewers) whose publish job declares no environment gets `Release environment: drift` / `Release gate: ok` (deferred), matching the doc comment that the two checks never both fire for one root cause.

**Not-applicable status.** A repo that publishes nothing (or ships `private: true`) returns `status: 'ok'` with detail `not applicable — no workflow job publishes to a registry`, not `drift`. Consistent with the `ok`/`drift`/`skip` pattern already used elsewhere in this file (e.g. `skip()` for the 403/unauthenticated case, 404 correctly read as "no environments" rather than unreadable).

**Commit-type honesty.** `feat(doctor):` is correct — this is a new check, and the PR body is explicit that clean repos will now report drift, which is a real behavior change, not silently swept under a `fix:`.

Job-splitting (`workflowJobs`) and comment-stripping are well-reasoned and regression-tested against the real false positive found dogfooding on this repo's own `ci.yml` (the `varcheck` job's trailing publish-related comment).

One thing worth a second look, which the author already flagged themselves: adding `'Release gate'` to `ACCEPTED` in `scripts/dogfood.mjs` is a judgment call to keep CI green on a genuinely-drifting self-repo finding. The comment explaining it is clear and self-deleting once the environment exists, but it's a human call on whether that's acceptable versus e.g. gating dogfood differently. Passing this through as a note, not a defect.

Minor edge case, not blocking: `workflowJobs` ends the `jobs:` block at the first non-blank, indent-0 line (`if (l.trim() !== '' && indentOf(l) === 0)`), with no exemption for a stray top-level comment. A workflow with a column-0 comment between `jobs:` and its first job (unusual, but not invalid YAML) would truncate the jobs block early and silently miss the publish job. Given real-world Actions files rarely write comments at that exact indent, this is unlikely to bite anyone in practice.

### Before merging
- Confirm the `scripts/dogfood.mjs` `ACCEPTED` entry for `'Release gate'` is the desired way to keep CI green on a real, currently-unfixable self-repo finding (flagged by the author as needing a second opinion, not a defect).

rtorcato:

## What to change

Both reviewers passed this and graded the finding below as non-blocking. **The orchestrator is overriding that and sending it back for one fix round.** The reviewers are not wrong about severity — the disagreement is about which axis to grade on. Reasoning below the fix, so it does not get in the way.

`findPublishJob` in `src/base/github-settings.ts`:

```js
} catch {
    return null
}
```

That `catch` wraps `fs.readdir` and `fs.readFile`. Returning `null` is read upstream as *"no workflow job publishes to a registry"* → status `ok`. That is correct when the workflows directory genuinely does not exist — which is already handled explicitly above the `try` — but the same `catch` also swallows real I/O failures: a permission error, a broken symlink, an encoding problem.

**Route the unreadable case to `skip(...)` in `checkReleaseGate`, the way `readEnvironments` already handles its own unreadable case in this file.** Distinguish "directory absent" (already handled, stays `ok`) from "directory present but unreadable" (should `skip`). The correct pattern exists a few lines away — follow it rather than inventing a second one.

While in there, the `code-reviewer` also noted a minor edge case in `workflowJobs`' block-boundary detection. Fix that in the same round if it is a small change; if it turns out to be structural, leave it and say so.

<details>
<summary>Why this blocks when the reviewers said it doesn't</summary>

Both reviewers weighed exploitability, and on that axis they are right: workflow files are local repo content, not attacker-controlled, so nobody can trigger this deliberately. As a *security* finding it is genuinely Low.

But this check's entire purpose is to tell a human whether a gate exists between merging `main` and publishing to npm. "I could not read your workflows" reported as `ok` means the tool says *you have a gate* when it does not know. A diagnostic that fails open in its core question is worse than not having the diagnostic, because it converts an unknown into a false reassurance — and nothing downstream distinguishes the two.

The security review said as much in its own words: *"the exact fail-open shape the task called out."* It then graded on exploitability and passed. That is a reasonable severity call and the wrong stopping point for this particular feature.

It also meets the bar for `ai-changes` rather than `ai-notes`: there is a concrete change an agent can make, and the reviewer already named it precisely. `ai-notes` is for questions only a human can answer; this is not one.

Everything else in both reviews stands and needs no action. In particular the environments-API path already fails closed correctly — only an explicit 404 reads as "no environments", while 403, 5xx and malformed JSON all resolve to `skip`, and a malformed payload pushes toward `drift` rather than `ok`. That is the right shape; this one path just does not match it yet.
</details>

rtorcato:
🤖 *Automated — fix round via `ai-issue-loop`. Posted under the owner's account by an agent; not a human message.*

Both items addressed in `ba55b61`.

**Fail-open in `findPublishJob` (the blocking one).** `findPublishJob` now returns `PublishJob | null | 'skip'`, mirroring `readEnvironments`' `Environments | 'skip'` a few lines away — the catch around `fs.readdir`/`fs.readFile` returns `'skip'` instead of `null`, and `checkReleaseGate` routes it to `skip(...)` on both checks with the reason `could not read .github/workflows`. The absent-directory case is untouched: still handled by the `fs.pathExists` guard above the `try`, still `ok`/not-applicable.

**`workflowJobs` block boundary — small, so fixed in the same round.** The end-of-block scan now skips column-0 comments (`!l.trimStart().startsWith('#')`), so a stray top-level comment between `jobs:` and its first job no longer truncates the block and hides every job below it. One predicate, not structural.

**Tests.** Two added: an unreadable workflows directory must report `skipped` / `could not read .github/workflows` and explicitly *not* `not applicable` — so it cannot regress to `ok`; and a jobs block with a column-0 comment still yields its jobs. The unreadable case is provoked by writing a *file* at `.github/workflows` (ENOTDIR), which avoids `chmod 000` — that would silently pass under a root test runner.

Full suite 949 passed / 59 files, `tsc --noEmit` clean, biome clean on `src scripts`. (Pre-existing lint errors exist under `tests/` in unrelated files; the repo's `lint` script only covers `src scripts`, and the file I edited lints clean.)

rtorcato (commented):

Verified all three points from the prior round's fix:

1. **Fail-open direction is correct.** `findPublishJob` returns `null` only when `.github/workflows` genuinely doesn't exist (`pathExists` false, checked before the `try`); any error inside the `try` (unreadable dir, broken symlink) returns `'skip'`. The one consumer, `checkReleaseGate`, checks `publish === 'skip'` before the `!publish` null-check, so the truthy `'skip'` string can never fall through and get treated as "nothing publishes" — confirmed by the new `readdir` ENOTDIR test that asserts `detail` says "could not read .github/workflows" and explicitly not "not applicable".

2. **No new information leak.** Both skip reasons are static strings (`'could not read .github/workflows'`, `'could not read environments'`) — no interpolated error message, errno, or path beyond what the check already reports. The caught exceptions in `findPublishJob` and `readEnvironments` are discarded, not surfaced.

3. **`workflowJobs` column-0 comment handling is the correct direction.** The boundary scan for the next top-level key now skips blank and `#`-prefixed lines before deciding a line is a new top-level key, so a stray comment between jobs no longer truncates the `jobs:` block and drops subsequent jobs (previously the under-reporting failure mode). Regression-tested (`is not truncated by a column-0 comment inside the jobs block`). Within a job body the per-line indent/header matching is unchanged, and a comment line can't masquerade as a job header since the header regex requires `[\w.-]` at the start, which `#` never satisfies.

No injection, secrets, unsafe shell construction, or dependency changes in this diff — `gh` is invoked with argv arrays, not shell strings, and `nwo` flows into a REST path argument, not a shell command.

### Before merging
Nothing.

rtorcato (commented):

Reviewed the latest commit against the two things the send-back asked for.

**1. `findPublishJob` fail-open — closed correctly.**
- Absent `.github/workflows` (`fs.pathExists` false) → `return null` → surfaces as `ok` / "not applicable — no workflow job publishes to a registry". Correct: absence isn't drift.
- Present-but-unreadable (the test uses a *file* at the `.github/workflows` path, so `readdir` throws `ENOTDIR`) → caught by the `try/catch` around the loop → `return 'skip'` → surfaces as `ok` / "skipped … could not read .github/workflows", never "not applicable". This is exactly the absent-vs-unreadable split the send-back required, and it's asserted directly in `tests/base/github-settings.test.ts` (`'skips, not "nothing publishes", when the workflows directory is unreadable'`), checking both the positive (`toContain('skipped')`) and negative (`not.toContain('not applicable')`) assertions.
- One theoretical gap, not a regression from this change: `fs.pathExists` uses `fs.access` with `F_OK`, which doesn't require read permission on the leaf directory, only traversal permission on its parents. A `.github` directory itself unreadable (not just `.github/workflows`) would make `pathExists` resolve `false` and fall through to `null`/"not applicable" rather than `skip`. Narrow edge case, outside what this round targeted, not worth blocking on.

**2. `workflowJobs` block-boundary handling — sound.**
The end-of-block scan now skips blank and column-0 comment lines and only stops at a real, non-comment, indent-0 line:
```js
if (l.trim() !== '' && !l.trimStart().startsWith('#') && indentOf(l) === 0) {
```
Traced both directions:
- A column-0 comment *before* the first job (between `jobs:` and the first job header) no longer truncates the block — covered by `'is not truncated by a column-0 comment inside the jobs block'`.
- A genuine next top-level key (e.g. `permissions:`) still ends the block regardless of preceding comments — the scan only advances past comments/blanks, so the first real key it hits still wins. Covered generally by `'splits jobs and stops at the next top-level key'`, though that test doesn't combine a preceding comment with the real boundary in one case. Valid YAML can't put job-body content at column 0 without dedenting out of the block scalar, so a comment can't smuggle extra job content past a genuine boundary. Any stray column-0 comment that survives inside a job's buffer gets stripped later by `withoutComments` before matching, so it's inert either way. Not a bug, just a slightly one-sided test — not blocking.

**3. Scope.** PR body says "`Refs #429`, deliberately not `Closes`" and confirms the scaffolding half (creating the `release` environment, wiring `environment: release`) stays out. Matches the issue's two-part ask and the stated split.

No correctness issues found in the areas this round addressed.

### Before merging
Nothing.

PR #456, opened 2026-08-20
feat(doctor): audit ai-issue-loop label colours and descriptions

### LINKED ISSUE #446: doctor: audit ai-issue-loop label colors — gh label create cannot repair existing labels
`doctor` already audits GitHub-side repo state — `github-settings.ts` for merge/protection settings, `milestones.ts` for milestone hygiene (#397). Labels are the same class of drift and are currently unowned, so they rot silently.

## The drift, measured

`ai-ready` was red (`#B60205`) in six of eight repos and green (`#0e8a16`) in two:

| Color | Repos |
|---|---|
| `B60205` (wrong) | browser-common, js-common, cf-common, db-common, react-common, api-common |
| `0e8a16` (spec) | shared-docs, repo-tooling |

`#B60205` is byte-identical to `ai-blocked`. The two labels a maintainer most needs to distinguish — "an agent should pick this up" vs "an agent gave up, a human must look" — rendered identically. That is not cosmetic; it misread as blocked during a live `ai-issue-loop` tick on browser-common.

## Root cause

The case is the tell. Every wrong value is uppercase `B60205`; every correct label in the same repos is lowercase (`b60205`, `fbca04`, `0e8a16`). Uppercase is what GitHub's web colour picker writes, lowercase is what `gh label create` writes — so `ai-ready` was hand-created in those six *before* the `ai-issue-loop` skill's setup block ever ran.

And the setup block uses:

```bash
gh label create ai-ready -c '#0e8a16' -d 'Eligible for an AI agent to implement'
```

`gh label create` **errors as a no-op when the label already exists** — the skill even instructs the reader to ignore that error. So it can create a missing label but can never repair an existing one's colour or description. The drift is permanent by construction, and re-running the skill's bootstrap in all six repos would have changed nothing.

Descriptions drifted the same way: browser-common read `Approved for AI-agent execution` against the spec's `Eligible for an AI agent to implement`.

## Proposal — `src/base/labels.ts`

Same shape and same `gh` seam as `milestones.ts`:

- A canonical table of the nine `ai-issue-loop` labels (name, colour, description).
- `doctor` reports drift read-only: missing labels, wrong colour, wrong description.
- `fix labels` repairs with **`gh label edit`**, not `create` — idempotent and able to correct existing labels, which is the whole point.
- Colour comparison must be **case-insensitive**, or every repo reports drift forever against `B60205` vs `b60205`.

Canonical table (from `skills/ai-issue-loop/SKILL.md`):

| Name | Color | Description |
|---|---|---|
| `ai-ready` | `0e8a16` | Eligible for an AI agent to implement |
| `ai-wip` | `fbca04` | Claimed by an agent; worktree exists |
| `ai-blocked` | `b60205` | Agent gave up; needs a human |
| `ai-review` | `1d76db` | PR awaiting agent review |
| `ai-ok-code` | `0e8a16` | code-reviewer passed |
| `ai-ok-sec` | `0e8a16` | security-expert passed |
| `ai-changes` | `d93f0b` | Reviewer requested changes |
| `ai-notes` | `fbca04` | Passed, but a reviewer left something to read before merging |
| `holding` | `5319e7` | Gate/holding issue — human judgement, never auto-picked |

Owning it here also lets the block come out of `SKILL.md`, where it cannot self-heal.

## Watch for

**Do not create missing labels by default.** Not every repo runs the loop — `supabase-common` has no `ai-*` labels at all, and `cf-common`/`db-common` have only `ai-ready`. Creating the full set everywhere would put nine labels into repos that will never use them. Suggest `doctor` only checks labels that already exist, and `fix` only creates the full set behind an explicit opt-in (or when two or more `ai-*` labels are already present, implying the loop is in use).

**Get the table right before enforcing it.** Per #422, `fix` commands PUT canonical state back silently; a wrong entry here would propagate to every repo with nothing in the output tying it to this change.

## Status

The immediate drift is already repaired by hand — all eight repos now read `0e8a16`. This issue is only for the durable check, so it cannot drift again unnoticed. Descriptions were deliberately left alone.

rtorcato:

Closes #446

## The bug

`gh label create` **errors as a no-op when the label already exists** — the skill's bootstrap block even tells the reader to ignore that error. So it can add a missing label but can never repair an existing one's colour or description. That is why `ai-ready` sat at `B60205` in six of eight repos, byte-identical to `ai-blocked`: the two states a maintainer most needs to tell apart rendered the same, and re-running the bootstrap would have changed nothing. Repair here goes through **`gh label edit`**.

## What landed

`src/base/labels.ts`, modelled on `src/base/milestones.ts` — same `gh` seam, same check/fix split, same self-skip conventions.

- **`AI loop labels`** doctor check. Read-only, `ok` outside a git repo or when `gh` can't answer. Reports wrong colour, wrong description, and missing labels.
- **`fix labels`** fixer. `gh label edit` for drift, `gh label create` only for labels the set is genuinely missing. Idempotent; `safe-add` risk level so the `--diff` shadow-run doesn't fire real writes during a preview, matching `github-settings` and `milestones`.

### Colour comparison is case-insensitive

Normalised (`#` stripped, lowercased) before comparing. GitHub's web picker writes uppercase, `gh` writes lowercase, and the two render identically — comparing raw would report every hand-created label as drift forever and have the fixer PATCH a colour that was already right. A label whose only difference is hex case is not reported and not touched; there is a test for each half.

### A repo that doesn't run the loop is *not applicable*, not drift

Fewer than two of the eleven labels present → `ok` with `not applicable — repo does not use the ai-issue-loop labels`, matching how `Release gate` says "not applicable" and `Milestones` skips a repo that uses none. The fixer honours the same gate, so a bare `fix --yes` cannot push eleven labels into a repo that opted out. One stray label (the observed `cf-common` / `db-common` half-state) is still not applicable.

## Source of truth — the choice you asked about

**`LOOP_LABELS` in `src/base/labels.ts` owns the table; `SKILL.md` keeps its bash block, and a test asserts the block matches the table exactly.**

Deleting the block was the tempting option, but it is the *only* bootstrap path for a repo with zero loop labels — precisely the repo `fix labels` deliberately refuses to touch. Removing it would trade one gap for another. Instead the block stays, gains a note that `create` cannot repair and that `fix labels` is what does, and `tests/base/labels.test.ts` parses it and compares it to `LOOP_LABELS`. The two cannot diverge without a red test.

Note the table has **eleven** labels, not the nine in the issue: `SKILL.md` is the live spec and also defines `ai-reviewing-code` / `ai-reviewing-sec`. Every colour and description is taken verbatim from that block — nothing was invented, which is the #422 caution about a `fix` command PUTting a wrong canonical value back into every repo.

## Verification

- `vitest run` — 988 passed, 18 skipped, 61 files. 17 of those are new.
- `tsc --noEmit --project src/cli/tsconfig.json` — clean.
- `biome check` on every touched file — clean.
- No label writes were run against any repository, this one included.

### DIFF
--- a/apps/docs/docs/guides/ai-issue-loop.md
+++ b/apps/docs/docs/guides/ai-issue-loop.md
@@ -83,6 +83,16 @@ that state, so a missed tick, a crash, or a restart costs nothing.
 | `ai-changes` | PR | A reviewer requested changes. |
 | `ai-notes` | PR | Passed, but a reviewer left something to read before merging. |
 
+Colours carry meaning here — `ai-ready` is green and `ai-blocked` red precisely
+so the two states a maintainer must tell apart are legible at a glance. `doctor`
+audits colour and description as the **`AI loop labels`** check, and `fix labels`
+repairs drift with `gh label edit`. The distinction matters: the skill's
+bootstrap block uses `gh label create`, which errors as a no-op on a label that
+already exists — so it can add a missing label but can never repair a
+hand-created one. A repo with fewer than two of these labels is reported as *not
+applicable* rather than drift: not running the loop is a choice, and neither the
+check nor the fixer pushes labels into a repo that opted out.
+
 ```
 issue: ai-ready ─pickup─> ai-wip ─> PR opened, labelled ai-review
 PR: ai-review ─> ai-reviewing-* ─┬─> ai-ok-code + ai-ok-sec ─┬─ issue PR  ─> assigned to you
--- a/skills/ai-issue-loop/SKILL.md
+++ b/skills/ai-issue-loop/SKILL.md
@@ -96,6 +96,19 @@ gh label create ai-changes -c '#d93f0b' -d 'Reviewer requested changes'
 gh label create ai-notes   -c '#fbca04' -d 'Passed, but a reviewer left something to read before merging'
 ```
 
+Bootstrap only. `gh label create` **cannot repair a label that already exists** —
+re-running this block against a hand-created `ai-ready` leaves whatever colour
+the web picker gave it, which is how six repos ended up with `ai-ready` rendering
+identically to `ai-blocked` (rtorcato/repo-tooling#446). To repair drift:
+
+```bash
+npx @rtorcato/repo-tooling doctor --json   # "AI loop labels" reports colour/description drift
+npx @rtorcato/repo-tooling fix labels      # repairs it with `gh label edit`
+```
+
+`src/base/labels.ts` in repo-tooling owns the canonical table and a test asserts
+this block matches it, so the two cannot diverge.
+
 Also once per repo, keep the status file out of git:
 
 ```bash
--- a/src/base/fixers.ts
+++ b/src/base/fixers.ts
@@ -37,6 +37,7 @@ import { detectLanguage } from '../cli/utils/detect-language.js'
 import type { Lockfile } from '../cli/utils/lockfile.js'
 import { resolveLanguageModule } from '../languages/registry.js'
 import { applyGithubSettings } from './github-settings.js'
+import { applyLoopLabels } from './labels.js'
 import { closeCompletedMilestones } from './milestones.js'
 import type { CheckResult } from './types.js'
 
@@ -277,6 +278,20 @@ export const BASE_FIXERS: Fixer[] = [
 		canFixDrift: true,
 		async run({ targetDir }) {
 			return { filesWritten: await closeCompletedMilestones(targetDir) }
+		},
+	},
+	{
+		target: 'labels',
+		description:
+			'Repair ai-issue-loop label colours and descriptions on GitHub via `gh label edit` (mutates the remote repo, not files). No-ops on a repo that does not use the loop',
+		appliesTo: ['AI loop labels'],
+		outputs: ['GitHub labels (remote, via gh label edit)'],
+		// safe-add for the same reason github-settings is: it exempts this fixer
+		// from the `--diff` shadow-run, which executes run() for a mere preview.
+		riskLevel: 'safe-add',
+		canFixDrift: true,
+		async run({ targetDir }) {
+			return { filesWritten: await applyLoopLabels(targetDir) }
 		},
 	},
 	{
--- a/src/base/labels.ts
+++ b/src/base/labels.ts
@@ -1,0 +1,234 @@
+import path from 'node:path'
+import chalk from 'chalk'
+import fs from 'fs-extra'
+import { type GhExec, realGhExec } from './github-settings.js'
+import type { CheckResult } from './types.js'
+
+/**
+ * `ai-issue-loop` label hygiene (#446). Same class of GitHub-side drift as
+ * github-settings.ts and milestones.ts, on the same `gh` seam.
+ *
+ * The bug this exists for: the skill's bootstrap uses `gh label create`, which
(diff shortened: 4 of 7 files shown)


rtorcato (commented):

Checked the three risk points from the issue, all handled correctly:

1. **Case-insensitive colour comparison.** `normalizeColor` strips `#` and lowercases before every comparison (`src/base/labels.ts`). Tests cover both an uppercase hex match and a leading-`#` match reporting `ok`, not drift — the exact `B60205` vs `b60205` case from #446 is verified not to false-positive.

2. **Fixer uses `gh label edit`, not `create`, for repair.** `applyLoopLabels` builds a `toEdit` set from `wrongColor`/`wrongDescription` and calls `gh label edit`; `gh label create` is only reached for genuinely `missing` labels. A dedicated test asserts zero `create` calls occur when repairing an existing label's colour — directly guards against reintroducing the root-cause bug.

3. **Not-applicable vs drift.** `IN_USE_THRESHOLD = 2` treats a repo with 0–1 matching labels as opted out (`ok`, "not applicable"), matching the issue's own suggested heuristic ("two or more ai-* labels already present"). Covered by tests for zero labels and the observed one-label half-state (`cf-common`).

Also addressed: the label spec's second copy. `LOOP_LABELS` in `src/base/labels.ts` is now canonical; `skills/ai-issue-loop/SKILL.md` keeps its bootstrap block (still the only path to seed a zero-label repo) but `tests/base/labels.test.ts` parses the block via regex and asserts it equals `LOOP_LABELS` exactly, so the two can't diverge silently. Reasonable choice given the block is a bootstrap-only alternate path, not pure duplication.

No correctness issues found. `readLabels`/`classifyLabels` are straightforward, the `Set` dedup for labels with both wrong colour and description relies on reference identity from the shared `LOOP_LABELS` array (correct, since deltas are built from that same array). Test coverage is thorough for a labels-only change.

### Before merging
Nothing.

rtorcato (commented):

**Shell construction.** `src/base/labels.ts` imports and reuses `realGhExec`/`GhExec` directly from `github-settings.ts` — it's the same seam, not a reimplementation, so there's no divergence to lose their protections. All `gh label edit`/`gh label create` calls pass args as array elements (`['label', 'edit', spec.name, '--color', spec.color, '--description', spec.description]`), never a concatenated shell string. Notably, the values written are *always* the hardcoded `LOOP_LABELS` table entries — the code never echoes back a value read from `gh label list` into an edit/create call (`classifyLabels` only uses `existing` for comparison). That means even a label with an adversarial description (quotes, backticks, `$(...)`, leading `-`) on the remote can't be replayed into a command; it just gets overwritten with the canonical spec string.

**Scope of the write.** No repo name is threaded through as a parameter — `gh` operates against whatever repo the local `dir` resolves to, the same target-scoping already used by `github-settings.ts`/`milestones.ts`. Nothing here lets a repo be supplied or inferred from untrusted input.

**Blast radius.** `applyLoopLabels` never passes `--name`, so it edits color/description only — it cannot rename a label, so pipeline eligibility (which is keyed on label *name*) can't be altered by this fixer. It only iterates the fixed 11-entry `LOOP_LABELS` set (no arbitrary label targeting), and `IN_USE_THRESHOLD = 2` blocks it from creating the full set in a repo that hasn't opted into the loop. Idempotent and read-gated the same way the existing checks are.

No secrets or tokens appear in output, and no new dependencies are introduced.

### Before merging
Nothing.

PR #462, opened 2026-08-20
fix(dependabot): stop the dev-minor auto-merge reaching peerDependencies

### LINKED ISSUE #458: The dev-minor Dependabot group contains 13 peerDependencies, so #447's auto-merge gate does not hold

PR #447 (`bb32f69`, closing #423) narrowed the scaffolded `dependabot-automerge.yml` to auto-merge only the `dev-minor` group, on the reasoning that dev-only bumps are safe to merge unattended while production dependencies must wait for a human.

**The `dev-minor` group is not dev-only.** It contains packages that ship to consumers, so the gate does not do what it was merged to do.

## Evidence

Dependabot PR #455 (`chore(deps-dev): bump the dev-minor group with 18 updates`) was reviewed by an agent that resolved each package's type from the root `package.json` rather than trusting the group name. Result:

- **13 of 18 are `peerDependencies`** of `@rtorcato/repo-tooling`, which is publishable (`private` is unset):
  `@biomejs/biome`, `@commitlint/cli`, `@commitlint/config-conventional`, `@next/eslint-plugin-next`, `@typescript-eslint/eslint-plugin`, `@typescript-eslint/parser`, `conventional-changelog-conventionalcommits`, `esbuild`, `eslint`, `eslint-plugin-jest`, `rollup`, `semantic-release`, `typescript-eslint`
- 5 are genuinely dev-only: `@playwright/test`, `@types/node`, `@vitejs/plugin-react`, `knip`, `@types/react`

None are majors, so the pre-existing `update-type` condition would not have stopped them either.

## Why this matters

`peerDependencies` are part of the published contract — a consumer installs against them. Under the merged gate, a `dev-minor` group PR auto-merges on green CI alone, so **13 consumer-facing packages would have merged unattended**, which is precisely the class #423 was filed to stop.

The gate reads as safe and is not. That is worse than the state before #447, where at least the name `production-minor` matched what it contained — here the group name actively misleads anyone auditing the workflow.

This PR did not auto-merge, but only because the pipeline reviewed it first and the reviewer checked the manifest instead of the group name. That check is not part of the scaffolded workflow; it is a property of the review agent's prompt. Consumers taking the scaffolded pair get no such protection.

## Root cause to confirm

The grouping is defined by the same generator (`dependabotConfig()` in `src/cli/generators/security.ts`). Dependabot's `dependency-type: development` classification appears not to treat `peerDependencies` as production — worth confirming against Dependabot's own semantics before choosing a fix, since the behaviour may differ by ecosystem.

## Options

| Option | Notes |
|---|---|
| **Exclude peer deps from `dev-minor`** | Most direct — make the grouping match the name. Needs a `dependabot.yml` pattern that Dependabot actually honours for peer deps. |
| **Gate on package identity, not group** | Have the workflow resolve each package's type from the manifest, the way the review agent does. Robust against grouping surprises, but much more workflow logic. |
| **Drop auto-merge entirely for this repo** | The `dev-minor` group is only 5 packages once peers are removed; the convenience may not be worth the failure mode. |

## Also

- The PR body was **truncated** (`_Description has been truncated_`), so the `updated-dependencies:` trailers were not visible. Classification came from the package/from/to table at the top, which survives truncation. Anything auditing these PRs from the trailers alone would have seen nothing to flag — failing open.
- #455 is currently `DIRTY` (conflicts) and its checks had not reported at review time.

## Related

- #423 / PR #447 — the gate this undermines
- #452 — the other two open questions from #447

rtorcato:
🤖 *Opened by Claude (AI agent) on rtorcato's behalf. Posted under the owner's account; there is no separate GitHub account for AI agents.*

Closes #458.

## The hole

`dependency-type: development` is Dependabot's classification, not a safety property. A package listed in **both** `devDependencies` and `peerDependencies` is filed as *development*, so it lands in the `dev-minor` group — while `peerDependencies` are part of what a published package hands its consumers.

This repo has **32** such packages:

```
$ jq -r '(.peerDependencies|keys) - ((.peerDependencies|keys) - (.devDependencies|keys)) | length' package.json
32
```

PR #455 bumped 18 in one `dev-minor` group PR, **13 of them peers of the published package**. The #423 gate would have auto-merged the lot on green CI alone. It only didn't because the branch was conflicting, so GitHub could not compute a merge ref and no checks ever ran — the gate was never actually tested against a live case.

## The fix

The workflow now resolves every bumped name against the tracked manifests instead of trusting the group name:

```yaml
ships=$(git ls-files -- 'package.json' '*/package.json' | while read -r f; do
  if [ "$(jq -r '.private // false' "$f")" = 'true' ]; then continue; fi
  jq -r '((.dependencies // {}) + (.optionalDependencies // {}) + (.peerDependencies // {})) | keys[]' "$f"
done | sort -u)
```

and the merge step gains `steps.gate.outputs.safe == 'true' &&`.

Three details worth reviewing:

- **Private workspaces are skipped.** They publish nothing, so their deps reach nobody. Flagging `apps/docs`'s React bump would train the reader to ignore the gate.
- **No dependency names reported ⇒ refuse.** Nothing verified is not the same as nothing to verify; that direction is where #423 failed open.
- **npm-specific by design.** A repo with no `package.json` (Swift, Python) finds nothing here and falls back to the group + semver gate, as before. Documented in the workflow comment rather than left to be discovered.

`doctor` gains the matching drift check, so a repo that took the #423 pair but not this one is told it still has the hole.

## Testing

The gate is shell, so **the test runs the shell** — extracts the `run:` body from the generated workflow and executes it under `bash` against a throwaway git repo. A string assertion would pass on a script that never sets `safe` at all.

Six cases: dev-only passes; a peer dep stands down; a runtime dep stands down; a private workspace's dep is ignored; a publishing workspace's dep stands down; empty names fails closed.

`vitest run` — 996 passed. The repo's own `action-pins` test caught `actions/checkout@v4` in the first draft and forced `@v7`, which is what the rest of the repo pins.

## Not done here

- **#455 itself is still stuck.** Rebasing it before this lands is what arms the trap. Once this is on `main`, `@dependabot rebase` is safe — the gate will see 13 peers and stand down.
- The `dev-minor` **grouping** is unchanged. It still collects peer deps; this makes that harmless rather than tidy. Narrowing the group in `dependabot.yml` would need a pattern Dependabot honours for peer deps, which is #458's option 1 and a separate question.

### DIFF
--- a/.github/workflows/dependabot-automerge.yml
+++ b/.github/workflows/dependabot-automerge.yml
@@ -17,11 +17,50 @@ jobs:
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
 
+      - uses: actions/checkout@v7
+
+      # The group name is Dependabot's opinion, not a safety property: a package
+      # in both devDependencies and peerDependencies is classified "development"
+      # and lands in dev-minor, while peerDependencies ship to consumers of a
+      # published package (#458). So resolve every bumped name against the
+      # tracked manifests and stand down if any of them reaches a consumer.
+      #
+      # npm-specific by design. A repo with no package.json (Swift, Python) finds
+      # nothing here and falls back to the group + semver gate below.
+      - name: Check whether any bumped package ships to consumers
+        id: gate
+        env:
+          NAMES: ${{ steps.metadata.outputs.dependency-names }}
+        run: |
+          set -euo pipefail
+
+          # A private workspace publishes nothing, so its deps reach nobody.
+          ships=$(git ls-files -- 'package.json' '*/package.json' | while read -r f; do
+            if [ "$(jq -r '.private // false' "$f")" = 'true' ]; then continue; fi
+            jq -r '((.dependencies // {}) + (.optionalDependencies // {}) + (.peerDependencies // {})) | keys[]' "$f"
+          done | sort -u)
+
+          safe=true
+          # No names reported means we cannot verify anything — fail closed.
+          if [ -z "${NAMES//[, ]/}" ]; then
+            echo "::notice::no dependency names reported — leaving this PR for a human"
+            safe=false
+          fi
+          for name in ${NAMES//,/ }; do
+            if printf '%s\n' "$ships" | grep -qxF -- "$name"; then
+              echo "::notice::$name ships to consumers — leaving this PR for a human"
+              safe=false
+              break
+            fi
+          done
+          echo "safe=$safe" >> "$GITHUB_OUTPUT"
+
       # Belt and braces: the dev-minor group is already declared minor+patch in
       # dependabot.yml, but this workflow is the security gate and shouldn't
       # trust a config file a consumer repo can edit independently.
       - name: Auto-merge dev-dependency patch and minor updates
         if: |
+          steps.gate.outputs.safe == 'true' &&
           steps.metadata.outputs.dependency-group == 'dev-minor' &&
           (steps.metadata.outputs.update-type == 'version-update:semver-patch' ||
           steps.metadata.outputs.update-type == 'version-update:semver-minor')
--- a/apps/docs/docs/guides/dependabot-strategy.md
+++ b/apps/docs/docs/guides/dependabot-strategy.md
@@ -26,7 +26,7 @@ Dependabot opens roughly **3–4 PRs per cycle** instead of one per package:
 | Group | Contents | Update types | Auto-merge |
 | --- | --- | --- | --- |
 | `production-minor` | runtime `dependencies` | minor, patch | ❌ manual |
-| `dev-minor` | `devDependencies` | minor, patch | ✅ on green |
+| `dev-minor` | `devDependencies` | minor, patch | ✅ on green, if nothing in it ships |
 | `major-updates` | all packages | major | ❌ manual |
 | `github-actions` | workflow actions | all | ❌ manual |
 
@@ -40,6 +40,20 @@ and `version-update:semver-patch` / `version-update:semver-minor`.
 
 Everything else waits for a human, and the gate **fails closed**: a PR outside a
 group reports an empty `dependency-group` and so never matches.
+
+> **The group name is not the gate.** `dependency-type: development` is
+> Dependabot's classification, and it files a package listed in *both*
+> `devDependencies` and `peerDependencies` as development — so it lands in
+> `dev-minor` while still being part of what a published package hands its
+> consumers. This repo has 32 such packages; 13 of them once rode a single
+> "dev-only" group PR. The workflow therefore resolves every bumped name against
+> the tracked manifests (`dependencies` + `optionalDependencies` +
+> `peerDependencies` of every non-private `package.json`) and stands down if any
+> of them ships. A PR reporting no dependency names at all is also refused —
+> nothing verified means nothing waved through.
+>
+> This step is npm-specific. A repo with no `package.json` finds nothing and
+> falls back to the group + semver gate alone.
 
 > **Production bumps are deliberately excluded.** A minor bump to a runtime
 > dependency of a published package ships to every consumer on the next release.
--- a/src/base/checks.ts
+++ b/src/base/checks.ts
@@ -304,6 +304,15 @@ export async function checkDependabot(dir: string): Promise<CheckResult> {
 				if (!automerge.includes("dependency-group == 'dev-minor'")) {
 					deltas.push('auto-merge workflow merges production bumps (missing dev-minor group gate)')
 				}
+				// #458: the group gate alone lets peerDependencies through, because
+				// Dependabot files a package that is both dev and peer under
+				// "development". The canonical workflow re-checks the names against the
+				// manifests; a workflow without that step is the #423 hole reopened.
+				if (!automerge.includes("steps.gate.outputs.safe == 'true'")) {
+					deltas.push(
+						'auto-merge workflow trusts the dev-minor group name (missing peerDependencies check)'
+					)
+				}
 			} else {
 				deltas.push('missing dependabot-automerge workflow')
 			}
--- a/src/cli/generators/security.ts
+++ b/src/cli/generators/security.ts
@@ -39,8 +39,11 @@ export function dependabotConfig(ecosystem: string | null): string {
         update-types:
           - minor
           - patch
-      # The only tier that auto-merges on green — dev tooling never reaches a
-      # consumer of the published package.
+      # The tier that can auto-merge on green. "development" is Dependabot's
+      # classification, not a safety property: a package listed in both
+      # devDependencies and peerDependencies lands here, and peer deps are part
+      # of a published package's contract (#458). The auto-merge workflow
+      # re-checks every name against the manifests rather than trusting this.
       dev-minor:
         dependency-type: development
         update-types:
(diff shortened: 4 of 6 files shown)


rtorcato (commented):

**Summary: no criticals or highs. 1 item worth a human's eyes (workspace-glob depth), everything else checks out — the gate is written defensively.**

### Injection / secrets
- `NAMES` (`steps.metadata.outputs.dependency-names`) is passed via `env:` and referenced only as `$NAMES` inside the `run:` block — it is never interpolated as `${{ ... }}` directly into the script text. That's the correct mitigation for the classic GitHub Actions script-injection pattern (untrusted `${{ }}` expansion landing in shell source); this PR does it right.
- The step never runs `npm install`/`npm ci` or executes anything from the checked-out tree — it only reads `package.json` through `jq`. So even though the job holds `contents: write, pull-requests: write` and checks out the Dependabot branch, there's no code-execution surface from the manifest content, only JSON parsing.
- `grep -qxF -- "$name"` uses fixed-string, whole-line matching, avoiding regex-metacharacter or partial-match tricks in package names.
- No hardcoded secrets in the diff; `GITHUB_TOKEN` usage is unchanged from the existing step.
- Residual/theoretical: `for name in ${NAMES//,/ }` is unquoted, so it's subject to word-splitting and pathname expansion. Not exploitable in practice — valid npm package names can't contain glob metacharacters (`*?[]`) or unescaped whitespace — but if you want it airtight, quoting into an array (`IFS=',' read -ra names <<<"$NAMES"`) removes the theoretical edge case for free.

### Fail-open / fail-closed behavior
- Empty/whitespace-only `NAMES` → `safe=false`. Correct fail-closed per the issue's stated goal (#423's failure mode was exactly "nothing verified, waved through anyway").
- Malformed/missing `package.json` under `set -euo pipefail`: a `jq` parse failure inside the `ships=$(...)` command substitution propagates through `pipefail` and trips `errexit`, aborting the step before `safe` is ever written. `steps.gate.outputs.safe` is then unset, and `steps.gate.outputs.safe == 'true'` evaluates false — so the merge step is skipped. Confirmed fail-closed, not fail-open.
- Private workspaces are correctly excluded only from the "ships" side (their deps can't leak into the safe-list); a bumped package's own privacy status is irrelevant since matching is by name against the consumer-facing set.
- One thing I can't verify without exploring the repo (out of scope for this review), so flagging as a question rather than a defect: `git ls-files -- 'package.json' '*/package.json'` — as a git pathspec, `*` does not cross `/`, so this matches the root manifest and exactly one level of nested workspaces (e.g. `packages/ui/package.json`), but **not** a package.json nested two levels deep (e.g. `packages/scope/name/package.json`). If this repo's workspace layout ever nests that deep, a peer/prod dependency there would silently miss the "ships" set and the gate would report `safe=true` for a consumer-facing bump — the exact failure mode #458 exists to close. Worth a maintainer confirming the actual workspace depth (or widening the pathspec to `**/package.json` while still excluding `node_modules`) before relying on this in a deeply-nested monorepo.

### Dependency / supply chain
- `actions/checkout@v7` is pinned by tag, not SHA, but that matches this repo's existing convention (the PR body notes the repo's own `action-pins` test enforces this) — not a new risk introduced here, just the repo's already-accepted posture.
- No new npm dependencies added; the `tests/cli/generators/security.test.ts` addition runs the extracted gate script under `bash` in a throwaway git repo rather than asserting on strings — good, since it actually exercises `set -euo pipefail` behavior instead of just checking for the presence of a string.

### Before merging
Nothing that blocks. One thing worth a maintainer's confirmation (not a concrete code change I can prescribe without seeing the repo's actual workspace layout): whether any package.json in this monorepo sits more than one directory level below root — if so, `'*/package.json'` in the gate's `git ls-files` pathspec won't see it, and a peer/prod dependency bump there would pass the gate as "safe" when it shouldn't.

rtorcato (commented):

Reviewed the `dependabot-automerge.yml` gate logic and its byte-identical twin in `src/cli/generators/security.ts`, plus the `checks.ts` drift check and doc update.

**Verified the two flagged risk areas directly, not just by inspection:**

- `git ls-files -- 'package.json' '*/package.json'` — tested empirically in a throwaway repo: git's default pathspec glob does *not* apply `FNM_PATHNAME` semantics, so a single `*` matches across `/`. It correctly picks up `apps/docs/package.json` and `packages/ui/package.json` (two levels deep), not just one level. The test fixtures in `tests/cli/generators/security.test.ts` exercise this same shape and are correct.
- TS vs YAML byte-identity — extracted both `run:` blocks from the diff and diffed them after mechanically undoing the template-literal escaping (`\${` → `${`, `\\n` → `\n`). They come out **identical**. Every `${` in the script is escaped in the TS version (`\${{ steps.metadata... }}`, `\${NAMES//[, ]/}`, `\${NAMES//,/ }`) and nothing else needed escaping (`$f`, `$(...)`, `$GITHUB_OUTPUT`, `$ships`, `$name`, `$safe` are bare `$`, not `${`, so they're inert in a JS template literal). `printf '%s\\n'` correctly double-escapes so the emitted file gets a literal `\n` for `printf`, matching the YAML.

**`set -euo pipefail` + the `while read` subshell:** if `jq` ever fails inside the loop (e.g. malformed JSON), bash's plain assignment `ships=$(...)` propagates the pipeline's non-zero exit under `pipefail`, and `set -e` (inherited into the piped subshell) would abort the step entirely — the step fails, `steps.gate.outputs.safe` is never set, and the `if:` gate on the merge step evaluates false. That's safe-direction failure, not a hole, and not a realistic case since these are already-committed, already-valid manifests at HEAD.

**`${NAMES//,/ }` unquoted in `for name in ...`:** technically subject to word-splitting *and* pathname expansion, but npm forbids `*`, `?`, and `[` in package names, so this can't be turned into an unintended glob in practice.

**No names reported → fails closed**, confirmed: `${NAMES//[, ]/}` strips both `,` and ` `, and an empty result correctly sets `safe=false` without being overwritten by the (then-empty, no-op) `for` loop.

`echo "safe=$safe" >> "$GITHUB_OUTPUT"` is written exactly once, after the loop, as a single `key=value` line — correct `$GITHUB_OUTPUT` usage, no multiline-value concerns since the value is always `true`/`false`.

The `private` check (`.private // false`), the dependency-type union (`dependencies + optionalDependencies + peerDependencies`, deliberately excluding `devDependencies`), and the new `checks.ts` drift detection (`steps.gate.outputs.safe == 'true'` substring match) all match the stated intent and the linked issue's root cause.

### Before merging

Nothing.

rtorcato:
🤖 *Automated — `ai-issue-loop` Pass 1 (orchestrator). Posted under the owner's account by an agent; not a human message. There is no separate GitHub account for AI agents, so this appears under @rtorcato's avatar.*

## Correcting the `security-expert` review above

Its one flagged finding — that `git ls-files -- 'package.json' '*/package.json'` *"only reaches one workspace level deep, so a peer/prod dep in a two-level-nested package.json would silently bypass the gate"* — **is not correct**, and `ai-notes` has been removed on that basis.

The two reviewers directly contradicted each other here (`code-reviewer` verified the glob does match nested manifests), so it was settled by running it:

```console
$ git init -q . && mkdir -p a/b/c
$ for f in package.json a/package.json a/b/package.json a/b/c/package.json; do printf '{}' > "$f"; done
$ git add -A
$ git ls-files -- 'package.json' '*/package.json'
a/b/c/package.json
a/b/package.json
a/package.json
package.json
```

Git pathspecs use `fnmatch` **without** `FNM_PATHNAME`, so `*` matches `/`. `*/package.json` is therefore depth-unbounded, not one level. No workspace depth is exempt from the gate and no maintainer confirmation is needed.

## Why the label was removed rather than left with this correction

`ai-notes` means *open the comments before merging*, and its stated failure mode is appearing on every PR until the reader stops looking. A note that has been disproven is exactly that noise. The review comment stays in the timeline so the record is intact — only the label is gone.

## Handing over

Both reviewers passed, `mergeStateStatus` is `CLEAN`, and this repo has no `release` environment with `required_reviewers` (the probe returns `0`), so it does not merge unattended. Assigned to @rtorcato.

⏳ **#455 is waiting on this.** Until this is on `main`, a `@dependabot rebase` there arms the very auto-merge this PR closes.

One thing genuinely worth knowing, which neither reviewer raised: the gate reads **tracked** files only. A manifest that is untracked or gitignored is invisible to it. That is the correct behaviour for a repo's published contract, but it is the assumption the check rests on.

PR #464, opened 2026-08-20
fix(skills): stop routing held Dependabot PRs through ai-changes

### LINKED ISSUE #459: ai-issue-loop: a Dependabot PR held for a human is labelled ai-changes and assigned to nobody
🤖 *Filed by Claude (AI agent) on rtorcato's behalf, from a `repo-tooling` session on 2026-08-20. Posted under the owner's account; there is no separate GitHub account for AI agents.*

`skills/ai-issue-loop/SKILL.md` ships to consumers, so this is a defect in published content, not a local workflow wart.

## Two bugs, one symptom

A Dependabot PR that trips the reviewer's "ships to consumers" rule ends up **labelled with a contradiction and assigned to nobody**, so it never reaches the human it is waiting on.

Observed on a `hono` bump in another repo of the family: `ai-changes`, `ai-notes`, `ai-ok-sec`, no `ai-review`, **Assignees: No one**, sitting for two weeks. The review comment's own closing line was *"merge promptly since 4.12.34 is a security release."* Nothing in the label set says which arm is holding it, and it appears in no queue.

### Bug 1 — the skill contradicts itself on what `ai-changes` means

The generic reviewer prompt (line ~507) is unambiguous:

> **If what you found is a question only a human can answer — pass it and note it. Never `ai-changes`.** `ai-changes` dispatches an implementer agent, and an agent cannot answer [...] Route it to the human directly: pass + `ai-notes`.

The Dependabot reviewer prompt (line ~578) then instructs the opposite:

> Apply `ai-changes` if **either** holds:
> - a package's major version differs between the `from` and `to` columns
> - a package ships to consumers [...]
>
> Those wait for a human

"Those wait for a human" is precisely the case the generic prompt reserves for `ai-notes`. So a reviewer following its prompt correctly emits a label whose documented meaning ("an agent should change this") is the opposite of its intent ("a human must decide"). The skill then needs a special case at line ~598 to undo the damage:

> **A Dependabot PR labelled `ai-changes` is terminal — never spawn a fix round for it.**

That paragraph exists only to neutralise a label that should not have been applied.

### Bug 2 — Pass 1 assigns nobody, so the PR is invisible

Pass 1's handoff covers exactly one case:

> for every **non-Dependabot** PR carrying both `ai-ok-code` and `ai-ok-sec` and not `ai-changes`, assign it

and separately assigns Dependabot PRs carrying `ai-notes`. **No branch assigns a Dependabot PR carrying `ai-changes`**, so it lands in no *Assigned to you* view and is discoverable only by scanning open PRs — which the skill itself calls out as the reason the handoff exists at all:

> "Merge it yourself" is only actionable if the user can find it, and a PR sitting in a list of open PRs looks identical to one still being worked.

This gap has a second producer. Pass 1's non-`CLEAN` check applies `ai-changes` to any both-passed PR that GitHub will not merge, Dependabot included — `#455` in this repo took that route (`DIRTY`) and is likewise unassigned.

## Proposed fix — deletes more than it adds

**Stop using `ai-changes` on Dependabot PRs.** Have the "ships to consumers" / major-version rules apply the *pass* label plus `ai-notes`, with the reason in `### Before merging`.

The gate is unchanged. The skill already states twice that `ai-notes` suppresses auto-merge:

- line ~120: *"Only the Dependabot arm merges itself, and only when no reviewer left `ai-notes`."*
- line ~291: *"A Dependabot PR carrying `ai-notes` is **not** auto-merged — assign it to the human."*

So nothing production-facing merges unattended either way, and the PR gains an assignee.

Consequences:

- The line ~598 "terminal, never spawn a fix round" special case can be **deleted** — nothing produces the state it guards against.
- The Dependabot prompt stops contradicting the generic prompt; one rule about `ai-changes` applies everywhere: *use it only when you can name a concrete change an agent could make.*
- The label set becomes legible without opening the PR:

| Labels | Means |
|---|---|
| `ai-ok-code, ai-ok-sec` | auto-merged, or will be on green CI |
| `ai-ok-code, ai-ok-sec, ai-notes`, assigned | human decides — read the comment |

**One caveat worth weighing.** The Dependabot prompt currently warns *"Be sparing with `ai-notes` here specifically: it suppresses auto-merge, so a reflexive note on every dependency bump wedges the one path that runs unattended."* Routing the consumer-facing rule through `ai-notes` raises its volume by design. That is the intended outcome — those bumps are meant to stop — but the "be sparing" wording needs rewriting so it discourages *decorative* notes rather than the gate itself.

**Still needed regardless:** Pass 1 should assign any Dependabot PR carrying `ai-changes` from the non-`CLEAN` path, so the invisible state cannot recur through that door.

## Related

- #458 — the scaffolded `dependabot-automerge.yml` auto-merges `dev-minor` on green CI alone, no labels involved. Different layer, same failure direction.
- #455 — currently `ai-changes` + `DIRTY` + unassigned in this repo; an instance of the Bug 2 gap.

rtorcato:
🤖 *Written by Claude (AI agent) via ai-issue-loop, on rtorcato's behalf. Posted under the owner's account; there is no separate GitHub account for AI agents.*

Closes #459

`skills/ai-issue-loop/SKILL.md` ships to consumers (`skills/*/SKILL.md` in `package.json` `files`), so this is a fix to published content.

## What changed

**The Dependabot reviewer no longer applies `ai-changes`.** The major-version and "ships to consumers" rules now apply the reviewer's **pass** label plus `ai-notes`, with the rule that fired stated in `### Before merging`.

The gate is unchanged — the skill already routes `ai-notes` Dependabot PRs away from auto-merge and into the human's *Assigned to you* view. What changes is that the label set stops contradicting itself: `ai-changes` means "an agent can act on this" everywhere, matching the generic reviewer prompt.

**Pass 1 now assigns any Dependabot PR carrying `ai-changes`.** Pass 1's own non-`CLEAN` check (`BLOCKED`/`DIRTY`/`BEHIND`) can still apply that label to a Dependabot PR, and no branch assigned it — which is how #455 ended up unassigned. This half was needed regardless of the reviewer change.

**The Pass 3 "terminal — never spawn a fix round" guard stays.** The issue proposed deleting it, but Pass 1's non-`CLEAN` path is still a producer of that state, so removing the guard would let a fix round spawn against a bot's lockfile. Its rationale is updated to name the remaining producer instead.

Carried through for consistency: the `ai-changes` row in the Labels table, the `ai-notes` "never blocks" paragraph, the state diagram, and the reviewer prompt's closing advice — rewritten to discourage *decorative* notes rather than the gate itself, since the consumer-facing rule now routes through it by design.

Also dropped a stale `⚠<n>held` claim in the sentence being rewritten: `held` appears in no Pass 5 `SUMMARY` segment, and Pass 5 reserves `⚠` for `blocked`/`ci-red`.

## Verification

- `biome lint --config-path=biome.json src scripts` — clean
- `vitest run` — 988 passed, 18 skipped

`tests/base/labels.test.ts` asserts the skill's `gh label create` bootstrap block against `LOOP_LABELS`; that block is untouched.

### DIFF
--- a/skills/ai-issue-loop/SKILL.md
+++ b/skills/ai-issue-loop/SKILL.md
@@ -64,13 +64,15 @@ header names the agent *and* says why it is wearing a human's face:
 | `ai-reviewing-sec` | PR | `security-expert` claimed and running. Cleared with its verdict. |
 | `ai-ok-code` | PR | `code-reviewer` passed. |
 | `ai-ok-sec` | PR | `security-expert` passed. |
-| `ai-changes` | PR | A reviewer requested changes. |
+| `ai-changes` | PR | A reviewer requested changes. Reviewers never apply it to a Dependabot PR. |
 | `ai-notes` | PR | Passed, but a reviewer left something to read before merging. |
 | `holding` | issue | A gate — closes on human judgement, never picked up. |
 
 **`ai-notes` is advisory and never blocks.** It rides *alongside* a pass label,
 never instead of one, and it never sends a PR back — a finding that should block
-is `ai-changes`. It exists because a pass label currently means both "clean" and
+an issue PR is `ai-changes`. On a Dependabot PR there is nothing to send back to,
+so `ai-notes` is the hold itself: it suppresses auto-merge and routes the PR to
+the human. It exists because a pass label currently means both "clean" and
 "I found something real but would not hold the PR over it", and those two are
 indistinguishable in the *Assigned to you* view where merges actually happen.
 The bar is a finding that **changes what a human would do**: a semver
@@ -121,8 +123,8 @@ PR: ai-review ─> ai-reviewing-* ─┬─> ai-ok-code + ai-ok-sec ─┬─ is
                                  │        (± ai-notes)       │              ─> YOU merge ─> worktree removed
                                  │                           └─ dependabot ─┬─ no ai-notes ─> auto-merge ─> worktree removed
                                  │                                          └─ ai-notes ───> assigned to you
-                                 └─> ai-changes ─> fix round (max 2) ─> ai-review
-                                                             └─ round 3 ─> ai-blocked
+                                 └─> ai-changes (issue PRs only) ─> fix round (max 2) ─> ai-review
+                                                                               └─ round 3 ─> ai-blocked
 ```
 
 `ai-reviewing-code` / `ai-reviewing-sec` are the *claim* step: Pass 3 applies one
@@ -289,6 +291,17 @@ gh pr edit <N> --add-label ai-changes \
 ```
 
 Count it as `rev`, not `ready`. A merge conflict (`DIRTY`) takes the same route.
+
+**Assign any Dependabot PR carrying `ai-changes`.** Pass 3 never spawns a fix
+round for one, so it is waiting on a human from the moment the label lands — and
+no other branch of this pass assigns it, which leaves it in no *Assigned to you*
+view at all:
+
+```bash
+gh pr edit <N> --add-assignee @me
+```
+
+Count it as `rev`. Idempotent, so it also picks up ones an earlier tick stranded.
 
 So: every open PR **authored by `dependabot[bot]`**, labelled both `ai-ok-code`
 and `ai-ok-sec`, **not** `ai-changes`, **not** `ai-notes`, that has no
@@ -577,15 +590,24 @@ package/from/to table survives because it sits at the top; classify from that.
 > the label. If you cannot tell whether a workspace publishes, treat it as
 > production.
 >
-> Apply `ai-changes` if **either** holds:
+> Apply your **pass** label *plus* `ai-notes` if **either** holds:
 > - a package's major version differs between the `from` and `to` columns
 > - a package ships to consumers — it appears in `dependencies`,
 >   `optionalDependencies`, or `peerDependencies` of a **non-private** package
 >
 > Those wait for a human — a runtime dependency of the published package, or a
-> major, is not something an automated verdict should wave through. Dev-only
-> minor/patch bumps with green CI get the pass label. **If you cannot determine a
-> package's type, treat it as production and block**; failing closed is correct here.
+> major, is not something an automated verdict should wave through. `ai-notes` is
+> the gate that holds them: it suppresses auto-merge outright and gets the PR
+> assigned to the human, so nothing production-facing lands unattended. Dev-only
+> minor/patch bumps with green CI get the pass label alone. **If you cannot
+> determine a package's type, treat it as production and note it**; failing
+> closed is correct here.
+>
+> **Never apply `ai-changes` to a Dependabot PR.** It dispatches an implementer
+> agent, and there is no change an agent could make — rewriting a bot's lockfile
+> is not its business, and the decision here is a human's either way. That is the
+> same rule as the generic prompt above: `ai-changes` only when you can name a
+> concrete change an agent could make.
 >
 > State in your comment which rule fired, name the packages that tripped it, and say
 > whether the body was truncated so the reader knows what you could and couldn't see.
@@ -595,9 +617,10 @@ package/from/to table survives because it sits at the top; classify from that.
 > Pass 3 claimed you with it before spawning you, and a claim left behind wedges
 > your half of the review until Pass 2 reaps it.
 >
-> Be sparing with `ai-notes` here specifically: it suppresses auto-merge, so a
-> reflexive note on every dependency bump wedges the one path that runs
-> unattended. A truncated body you could not fully read **is** worth a note; a
+> Keep `ai-notes` load-bearing here: it suppresses auto-merge, so a *decorative*
+> note on a bump you would otherwise wave through wedges the one path that runs
+> unattended. A major, a package that ships to consumers, or a truncated body you
+> could not fully read **is** worth a note; restating the version table on a
 > routine dev-only patch bump is not.
 
 Be honest about what this buys: an agent reading a version table catches majors,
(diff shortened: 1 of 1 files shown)


rtorcato (commented):

Reviewed as a control document (this SKILL.md governs merge decisions in an unattended pipeline), not as prose.

**Does anything now auto-merge that did not before?** No. The actual auto-merge gate for Dependabot PRs — "every open PR ... labelled both `ai-ok-code` and `ai-ok-sec`, **not** `ai-changes`, **not** `ai-notes`, that has no `autoMergeRequest` yet" — is untouched by this diff and still excludes `ai-notes` explicitly. Consumer-facing/major bumps move from `pass + ai-changes` to `pass + ai-notes`, but both labels are checked by the same exclusion, so the block that used to ride on `ai-changes` now rides on `ai-notes` instead. Confirmed in the diff's unmodified context, not inferred from the prose.

**Pass label alone can't satisfy the merge condition.** Verified — the condition requires `not ai-notes` as a separate clause, so a pass label without `ai-notes` is required for auto-merge to proceed; a pass label with `ai-notes` is excluded either way.

**Fail-closed / half-labelled state.** If a reviewer emits the pass label but omits `ai-notes` on a consumer-facing bump, it would slip through — but that failure mode already existed under the old scheme (omitting `ai-changes` had the identical effect). This diff doesn't change that risk surface, it just moves which label the reviewer must remember to attach.

**Pass 3 terminal guard.** Correctly kept rather than deleted (the issue proposed deleting it) — Pass 1's own non-`CLEAN` path can still emit `ai-changes` on a Dependabot PR, and the diff adds an explicit Pass 1 assignment step for that case so it no longer goes unassigned. This is the more conservative of the two options the issue raised and matches its "still needed regardless" note.

**`--approve` / `required_pull_request_reviews` / untrusted-issue-body rules.** Not touched by this diff.

I did not find a "release-environment exception" section in this diff to check against — nothing in the diff or its surrounding context references one, and I stayed diff-scoped per instructions rather than reading the full file.

### Before merging
Nothing.

rtorcato (commented):

Reviewed against the linked issue (#459) and the diff only, as scoped.

**Internal consistency — checked.** Every remaining `ai-changes` mention reads correctly against the new rule:
- Labels table: `ai-changes` row now notes reviewers never apply it to a Dependabot PR.
- `ai-notes` "never blocks" paragraph: now distinguishes issue PRs (ai-changes is the block signal) from Dependabot PRs (ai-notes itself is the hold, since there's nothing to send back to).
- State diagram: the `ai-changes` branch off the reviewer-claim node is now annotated `(issue PRs only)`, matching the rule. The second line's leading spaces shift by exactly the width of the inserted `(issue PRs only) ` text, so the ASCII-art alignment is preserved, not broken.
- Dependabot reviewer prompt: the major-version/ships-to-consumers rule now applies pass label + `ai-notes` instead of `ai-changes`, including the "cannot determine package type" fallback (now "note it" instead of "block"), plus an explicit new "Never apply `ai-changes` to a Dependabot PR" rule that matches the generic prompt's wording.

**Terminal guard kept, not deleted.** The "Dependabot PR labelled `ai-changes` is terminal — never spawn a fix round" paragraph survives and is correctly reworded to name Pass 1's non-`CLEAN` check as the remaining producer, rather than being removed as the issue's proposed fix suggested. Nothing in the diff dispatches an implementer at a Dependabot PR.

**Pass 1 assignment gap — actually fixed.** A concrete new block (`gh pr edit <N> --add-assignee @me`) assigns any Dependabot PR carrying `ai-changes`, placed right after the non-CLEAN/DIRTY handling in Pass 1, addressing the exact "no branch assigns it" gap from #459 (and #455's stranded state, since it's called idempotent).

**`ai-notes` volume.** The "be sparing" warning is rewritten to "keep `ai-notes` load-bearing" — it now explicitly still counts majors/consumer-facing/truncated-body findings as worth a note, while only discouraging *decorative* notes on routine dev-only bumps. This matches the issue's caveat: raises volume by design without re-opening the door to reflexive notes.

Also caught the stale `⚠<n>held` claim removed from the Pass 3 paragraph, matching the PR description's note that `held` never appears in a Pass 5 `SUMMARY` segment.

No bugs found. This is a markdown-only change to `SKILL.md`; verification commands in the PR description (biome/tsc/vitest) don't exercise this text but there's nothing here that could regress them.

### Before merging

Nothing.

rtorcato:

## What to do

**Merge it.** Both reviewers passed with nothing to read first, `mergeStateStatus` is `CLEAN`, and every check is green.

This repo has **no `release` environment with `required_reviewers`** — the probe returns `0` protection rules:

```bash
gh api repos/rtorcato/repo-tooling/environments \
  --jq '[.environments[] | select(.name=="release") | .protection_rules[]? | select(.type=="required_reviewers")] | length'
```

So merging `main` fires semantic-release and publishes to npm with nobody in the loop, and only Dependabot PRs — whose `chore(deps)` squash subject cuts no release — merge unattended here. That gate is #429, still `holding`: it needs you to name who the reviewers should be before an agent can scaffold it.

The squash subject on this one is `fix(skills):`, so merging cuts a **patch release**.

## What the reviewers checked

Both were briefed specifically, because this PR edits the document that governs this pipeline — an internal contradiction in it becomes a behaviour bug on the next tick, not just a docs wart.

- `security-expert` — verified the actual auto-merge condition is untouched and still excludes **both** `ai-changes` and `ai-notes`. That is the load-bearing check: moving consumer-facing Dependabot bumps from `ai-changes` to `ai-notes` is only equivalent if `ai-notes` suppresses auto-merge on every path. It does.
- `code-reviewer` — verified every remaining `ai-changes` mention stays internally consistent, that the "terminal — never spawn a fix round" guard was **kept** rather than deleted (Pass 1's non-`CLEAN` path can still produce that label), that the Pass 1 assignment gap is fixed with a real `gh` command rather than just described, and that the `ai-notes` volume warning was narrowed to decorative notes instead of discouraging the gate.

## Note

`skills/ai-issue-loop/SKILL.md` ships to npm (`package.json` `files` carries `skills/*/SKILL.md`), so this changes published content for every consumer, not just this repo's own loop. Your local copy under `~/.claude/skills/` will not pick it up until a `fix claude-skills` run after the next release.

Issue #411, opened 2026-08-17
fix claude-skills fails open: --yes/--json reports success when nothing was installed


rtorcato:
_(AI-authored issue.)_

`fix claude-skills` under `--yes` / `--json` reports success when it installed
nothing.

`resolveInstallDir` (`src/base/fixers.ts:99`) returns `null` when no skills
directory resolves and a prompt is unavailable, printing a yellow advisory. The
fixer turns that `null` into `{ filesWritten: [] }` (`src/base/fixers.ts:334`),
which the command records as an ordinary applied-with-no-files action and exits
`0`.

## Why that's the wrong shape

`--json` promises a machine-readable result, and the payload for "I could not
find anywhere to install this" is byte-identical to "already up to date":

```json
{ "directory": "…", "target": "claude-skills", "actions": [ { …, "filesWritten": [] } ] }
```

A script driving `fix claude-skills --json` in CI cannot tell the two apart, and
the advisory that explains the difference goes to stderr where the JSON consumer
isn't looking.

## The docs already describe the intended behaviour

Both say the flag is *required* in exactly this case:

- `apps/docs/docs/guides/ai-issue-loop.md:28` — "**`--skills-dir <path>`** overrides the destination. It is required alongside `--yes` / `--json` when `~/.claude/skills` does not exist"
- `src/cli/index.ts:353` (the flag help) — "Required with --yes/--json when that directory does not exist"

So this is the code diverging from the documented contract, not a docs bug.

## Done when

`fix claude-skills --yes` / `--json` with no resolvable destination and no
`--skills-dir` exits non-zero with a distinguishable error — under `--json`, an
error payload in the shape of the existing `no-lockfile` / `unknown-target`
cases (`src/cli/commands/fix.ts:374`, `:463`). The interactive path is
unchanged: an empty prompt answer is a user declining, not a failure.

Found by a reviewer on #409, posted after that PR merged.

Issue #327, opened 2026-08-07
Post-mortem: placeholder git identity mis-attributed ~703 commits across 23 repos


rtorcato:
Roll-up / post-mortem. Individual tracking issues are filed in each affected repo and link here.

## What happened

In the dotfiles repo, `~/.gitconfig` is a symlink to a **tracked** file, `dotfiles/gitconfig/.gitconfig`. Its committed `[user]` block held a deliberate placeholder:

```
email = richardtorcato@example.com
```

Because the live git config *was* a versioned file, two things followed:

1. Running `git config --global user.email <real address>` wrote into the repo, showing up as an unexpected modification.
2. Any `git checkout` / `switch` / `pull` that restored that file silently reset the identity back to the placeholder.

So the identity oscillated, and commits made while the placeholder was in place were authored as an address that resolves to no account. It is **intermittent**, which is exactly why it survived months unnoticed — plenty of commits in the same repos are correctly attributed.

A related failure mode produced commits from `richardtorcato@Richards-Mini.matrix`: git's hostname-derived fallback, used when `user.email` is unset entirely.

## Scope

Fully paginated counts (placeholder-authored / total commits), 2026-08-07:

### GitHub

| Repo | Affected |
|---|---|
| browser-common | 88/156 |
| **repo-tooling** | **79/397** |
| homelab-nut | 68/137 |
| infra-x | 52/168 |
| api-common | 45/147 |
| shadcn-ui | 43/145 |
| swift-common | 34/69 |
| js-common | 29/147 |
| supabase-common | 29/81 |
| rtorcato (profile README) | 23/28 |
| db-common | 15/47 |
| vibe-start | 15/15 |
| shared-docs | 9/22 |
| resume-rtorcato | 6/27 |
| cf-common | 5/46 |
| react-common | 4/93 |
| db-x | 2/52 |

### GitLab

| Repo | Affected |
|---|---|
| ai-lab | 54/55 |
| infra-x.gi | 34/70 |
| infra-x | 34/90 |
| homelab-control-node | 33/55 |
| learn-investing | 1/42 |
| homelab-one | 1/84 |

**~703 commits across 23 repos.** `dotfiles` itself had a further 322 and has already been fixed.

## Why there is no cheap retroactive fix

The usual remedy is to add the stray address as a verified secondary email so the commits re-link automatically. That is impossible here: `example.com` is IANA-reserved, so the address can never receive mail and can never be verified.

[...]

PR #465, opened 2026-08-20
test(ci): verify the CLI against an empty HOME with no dotfiles

### LINKED ISSUE #460: Verify repo-tooling works for a public consumer with no dotfiles repo
🤖 *Filed by Claude (AI agent) on rtorcato's behalf, from a `repo-tooling` session on 2026-08-20. Posted under the owner's account; there is no separate GitHub account for AI agents.*

`@rtorcato/repo-tooling` is a public package. A stranger installing it into their own project has **no dotfiles repo at all** — certainly not the maintainer's private one — yet parts of the shipped surface were written in an environment where that repo is always present. Nothing here is known to be broken for them; the point is that **nothing verifies it isn't**, and the failure mode is silent.

## The durable fix

One CI job is worth more than any number of greps: run the CLI end to end against a scratch repo with `HOME` pointed at an **empty temp directory**, so no `~/.claude`, no stow symlinks, no maintainer dotfiles exist.

```
HOME=$(mktemp -d)  →  npm pack / install the tarball into a fresh git repo
                   →  repo-tooling setup
                   →  repo-tooling doctor
                   →  repo-tooling fix <each target>
```

Assert: no crash, no path outside the project or the fake `HOME` is written, and every message names a path that actually exists in that environment. That closes the whole class instead of chasing instances, and it fails loudly the next time one is introduced.

The greps below are seed evidence, not the scope.

### 1. The shipped skill references a command that only exists in the maintainer's private dotfiles

`skills/ai-issue-loop/SKILL.md:158`:

> (`/_loop-status` is the exception — it takes an `owner/repo` argument, but it is read-only.)

`/_loop-status` is defined in `commands/_loop-status.md` in the owner's **private** dotfiles repo. It ships in no package. A consumer who installs the skill (`fix claude-skills` → `~/.claude/skills/ai-issue-loop/SKILL.md`) reads a parenthetical about a command they cannot run and cannot look up.

`/loop` (lines 9, 849–853) is fine — that's a Claude Code built-in.

Sweep the shipped skills and `tooling/claude/*.md` for any other `/_`-prefixed command reference; those are private by convention.

### 2. A personal LAN domain is baked into a public check

`src/base/git-identity.ts:35`:

```ts
/**
 * Suffixes of a machine-derived hostname. `.local` is mDNS (macOS default),
 * `.matrix` is this user's LAN — both produced real commits in the #327 set.
 */
const HOSTNAME_SUFFIXES = ['.local', '.matrix', '.localhost', '.lan', '.home', '.internal']
```

The comment says outright that `.matrix` is one person's network. Low severity — an extra suffix only widens detection — but it is exactly the pattern worth finding, and the comment reads oddly to anyone else.

Worth noting the same file gets this **right** a few lines up, and that reasoning is the standard to apply everywhere else:

> Its fourth rule — email sourced from a *tracked* gitconfig — is deliberately not ported: that's a property of the dotfiles symlink layout, not of an arbitrary repo.

### 3. stow-symlink handling needs a no-symlink path confirmed

`src/cli/generators/claude-skills.ts:114-137` detects when a skill file is a stow symlink and deliberately writes *through* it, and `src/base/fixers.ts:404` then prints:

```
wrote through a symlink — commit <realFile>
```

This looks correctly conditional on `result.viaSymlink`, so a plain `~/.claude/skills/` should take the ordinary path — but that is an assumption, not a test. Confirm under the empty-`HOME` job that a first-time consumer gets a normal write and a message naming a path inside their own `HOME`.

### 4. An emitted workflow hard-depends on the maintainer's repo at `@main`

`src/cli/generators/docs-site.ts:391` writes into the **consumer's** repository:

```yaml
uses: rtorcato/repo-tooling/.github/workflows/docs-deploy.yml@main
```

Not a dotfiles issue, but the same class: a consumer's CI silently gains a dependency on a repo they don't control, unpinned, so any push to `main` there changes their build. At minimum this deserves a documented decision; pinning to a tag or SHA is the conventional answer.

### 5. Confirm the documented escape hatch actually holds

`AGENTS.md:87` and `README.md:190` both document `~/.claude/skills` as the default with a `--skills-dir` override. That is the right design. Verify the override is honoured on every path that touches `~/.claude`, and that nothing writes there when it is set.

## Out of scope

`@rtorcato/...` in preset matchers (`src/languages/js/checks.ts`, `src/base/checks.ts`) is **correct** — those assert the consuming repo extends *this* package's presets, and the package genuinely is scoped `@rtorcato`. Not to be "fixed".

## Related

- Repo convention already recorded: never bake org-specific values into emitted files or checks; derive them from the consuming repo.

rtorcato:
🤖 *Automated — implementer via ai-issue-loop. Written by an AI agent and posted under the owner's account; not a human message.*

Closes #460

## What this adds

`scripts/integration/no-dotfiles.mjs` plus a `integration-no-dotfiles` CI job that runs the CLI the way a public consumer does: `pnpm pack` the working tree, install the tarball into a fresh `git init`'d project, then run `setup`, `doctor` and `fix` with **`HOME` pointed at an empty temp dir** — no `~/.claude`, no stow symlinks, no maintainer dotfiles. The CLI is invoked through the *installed tarball's* bin, not `dist/`, so anything that only resolves inside this checkout fails the job.

Assertions:

- `setup --preset library` exits 0 and creates no `~/.claude`.
- `doctor --json` emits a parseable report and reports `Claude skills` as `optional-missing` rather than crashing; creates no `~/.claude`.
- `fix --yes --json` (**the full matrix**, 15 targets on this scaffold) writes only inside the project, and leaves `~/.claude` alone because `claude-skills` is `explicitOnly`.
- `fix claude-skills --yes --json` with nothing to resolve refuses with `error: "no-skills-dir"` (#411) and creates no `~/.claude`.
- `--skills-dir` is honoured, and `~/.claude` stays absent while it is set.
- With a plain `~/.claude/skills` present, the skill is written as a **real file, not a symlink**, with no `wrote through a symlink` notice, and every reported path exists inside the fake `HOME` — seed finding 3 confirmed rather than assumed.
- The machine's real `HOME` gains no entries and its own installed `ai-issue-loop` skill is byte-identical afterwards.

Nothing was left out of the `fix` matrix: `fix --yes` runs every non-`explicitOnly` target in one pass and `claude-skills` gets four dedicated runs on top. Ubuntu, one Node major (22) — this exercises path and `HOME` handling, not the runtime, which the existing `integration` job already matrixes across 22/24. The job is wired into the `release` gate alongside `integration` and `integration-swift`. Locally the whole script takes well under a minute after the install.

## Seed findings

**Fixed — 1.** `skills/ai-issue-loop/SKILL.md` referenced `/_loop-status`, which exists only in the maintainer's private dotfiles and ships in no package. Dropped the parenthetical; the surrounding rule ("reads against other repos are fine, writes are not") stands without it. **Sweep result:** that was the *only* `/_`-prefixed reference in `skills/` and `tooling/claude/` — and in every tracked `.md` outside `CHANGELOG.md`. `/loop` is a Claude Code built-in and was left alone. The edit is one line and does not touch the region #459 is editing.

**Fixed — 2.** `src/base/git-identity.ts` — reworded the comment that called `.matrix` "this user's LAN". **I kept the suffix in the array**: removing it narrows detection for a case that produced real commits in the #327 set, and the array is a heuristic where an extra entry only widens coverage. If you'd rather it go, that's a one-line follow-up. `tests/base/git-identity.test.ts` still uses `Richards-Mini.matrix` as a fixture, which seemed fine to leave — it's a realistic example, not a shipped value.

**Confirmed — 3.** The stow-symlink handling is correctly conditional on `viaSymlink`; the new job proves the plain path produces an ordinary write and no symlink message.

**Confirmed with a caveat — 5.** `--skills-dir` is honoured on the only path that *writes* to `~/.claude` (`fix claude-skills` → `resolveSkillsDir`), and nothing is written there when it is set. **But `doctor` has no `--skills-dir` option at all**: `src/base/checks.ts:513` calls `claudeSkillStatus()` with no argument, so the function's `explicit` parameter is dead and doctor always reports against `~/.claude/skills` regardless. Read-only, so it's a reporting asymmetry rather than a stray write — a consumer using `--skills-dir` gets a permanent `optional-missing` for a skill they have installed. Adding the flag to `doctor` is a CLI-surface change, so I left it for you.

## For your decision — not changed

`src/cli/generators/docs-site.ts:391` emits `uses: rtorcato/repo-tooling/.github/workflows/docs-deploy.yml@main` into the **consumer's** CI. That is a real, unpinned cross-repo dependency: any push to `main` here changes their build, and it's a supply-chain surface they don't control. Pinning to a tag or a SHA is the conventional answer, but it's a maintenance decision (a SHA needs Dependabot or manual bumps to stay current), so the code is untouched.

## Out of scope, per the issue

`@rtorcato/...` in the preset matchers (`src/languages/js/checks.ts`, `src/base/checks.ts`) is correct and was not touched.

## Verification

`biome check src scripts`, `tsc --noEmit`, `vitest run` (1006 tests, 62 files) all pass. `node scripts/integration/no-dotfiles.mjs` passes locally on macOS; CI will be its first Linux run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -265,6 +265,36 @@ jobs:
             node scripts/integration/preset-lifecycle.mjs "$preset"
           done
 
+  # Integration: the CLI as a public consumer sees it — HOME pointed at an empty
+  # temp dir, so no ~/.claude, no stow symlinks and none of the maintainer's
+  # dotfiles exist (#460). Catches the silent class where shipped surface assumes
+  # a machine that only the maintainer has. One Node major: this exercises path
+  # and HOME handling, not the runtime, which the job above already matrixes.
+  integration-no-dotfiles:
+    runs-on: ubuntu-latest
+    needs: [check-skip]
+    if: needs.check-skip.outputs.should-skip != 'true'
+    steps:
+      - name: 📦 Checkout repository
+        uses: actions/checkout@v7
+
+      - name: 📦 Setup Node.js
+        uses: actions/setup-node@v7
+        with:
+          node-version: "22"
+
+      - name: 📦 Setup pnpm
+        uses: pnpm/action-setup@v6
+
+      - name: 📦 Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: 🏗️ Build CLI
+        run: pnpm build-cli
+
+      - name: 🏠 setup + doctor + fix against an empty HOME
+        run: node scripts/integration/no-dotfiles.mjs
+
   # swift-library can't ride the job above: ubuntu-latest ships no Swift
   # toolchain, and the scaffold's own CI targets Apple platforms anyway (#312).
   # macos-latest over a swift:* container because macOS runners are free on
@@ -372,6 +402,7 @@ jobs:
         typecheck,
         test,
         integration,
+        integration-no-dotfiles,
         integration-swift,
         build,
         varcheck,
@@ -386,6 +417,7 @@ jobs:
       needs.typecheck.result == 'success' &&
       needs.test.result == 'success' &&
       needs.integration.result == 'success' &&
+      needs.integration-no-dotfiles.result == 'success' &&
       needs.integration-swift.result == 'success' &&
       needs.build.result == 'success' &&
       needs.varcheck.result == 'success'
--- a/scripts/integration/no-dotfiles.mjs
+++ b/scripts/integration/no-dotfiles.mjs
@@ -1,0 +1,216 @@
+#!/usr/bin/env node
+// Integration test: run the CLI the way a *public consumer with no dotfiles
+// repo* does — `HOME` pointed at an empty temp dir, so there is no `~/.claude`,
+// no stow symlinks and none of the maintainer's config (#460).
+//
+// The CLI runs from the installed tarball's bin, not from dist/, so anything
+// that only resolves inside this checkout fails here.
+//
+// Asserts: nothing crashes, every reported write lands inside the project or the
+// fake HOME, the machine's real HOME is untouched, and `--skills-dir` keeps
+// `~/.claude` unused.
+//
+// Usage: node scripts/integration/no-dotfiles.mjs
+// Requires: network access to the npm registry (`pnpm pack` + install).
+
+import { execFileSync, spawnSync } from 'node:child_process'
+import fs from 'node:fs'
+import os from 'node:os'
+import path from 'node:path'
+
+const REPO = process.cwd()
+const REAL_HOME = os.homedir()
+
+const projectDir = fs.mkdtempSync(path.join(os.tmpdir(), 'rt-nodots-project-'))
+const fakeHome = fs.mkdtempSync(path.join(os.tmpdir(), 'rt-nodots-home-'))
+const packDir = fs.mkdtempSync(path.join(os.tmpdir(), 'rt-nodots-pack-'))
+const BIN = path.join(projectDir, 'node_modules', '.bin', 'repo-tooling')
+
+let failed = false
+
+function fail(msg) {
+	console.error(`\n❌ ${msg}`)
+	failed = true
+}
+
+function assert(cond, msg) {
+	if (cond) console.log(`  ✅ ${msg}`)
+	else fail(msg)
+}
+
+function run(cmd, args, cwd = REPO) {
+	console.log(`\n$ ${cmd} ${args.join(' ')}${cwd !== REPO ? `  (cwd=${cwd})` : ''}`)
+	execFileSync(cmd, args, { stdio: 'inherit', cwd })
+}
+
+/**
+ * Run the CLI under the fake HOME. A non-zero exit is a legitimate outcome here
+ * (doctor findings, the `--skills-dir` refusal), so the status comes back for the
+ * assertions to judge instead of throwing and losing the captured output.
+ */
+function cli(...args) {
+	const argv = [...args, '--directory', projectDir]
+	console.log(`\n$ repo-tooling ${argv.join(' ')}   (HOME=${fakeHome})`)
+	const res = spawnSync(BIN, argv, {
+		cwd: projectDir,
+		encoding: 'utf8',
+		env: { ...process.env, HOME: fakeHome, USERPROFILE: fakeHome },
+	})
+	if (res.error) throw res.error
+	if (res.stdout) console.log(res.stdout)
+	if (res.stderr) console.error(res.stderr)
+	return { status: res.status ?? 1, stdout: res.stdout ?? '', stderr: res.stderr ?? '' }
+}
+
+function parseJson(raw, what) {
+	try {
+		return JSON.parse(raw)
+	} catch {
+		fail(`${what} did not emit parseable JSON:\n${raw}`)
+		return {}
+	}
+}
+
+/** True when `file` resolves to somewhere under `dir`. Both must exist. */
+function contains(dir, file) {
+	if (!fs.existsSync(file)) return false
+	const rel = path.relative(fs.realpathSync(dir), fs.realpathSync(file))
+	return rel !== '' && !rel.startsWith('..') && !path.isAbsolute(rel)
+}
+
+// In-repo fixers report project-relative paths; the one that writes user-global
+// state reports absolute. Resolve both against the project so `contains` can
+// judge them the same way — a relative `../escape` still lands outside.
+const filesWritten = (payload) =>
+	(payload.actions ?? [])
+		.flatMap((a) => a.filesWritten ?? [])
+		.map((f) => (path.isAbsolute(f) ? f : path.join(projectDir, f)))
+
+// 1. Pack the working tree and install it into a bare project, the way a
+//    consumer installs from npm. This step runs under the real HOME on purpose:
+//    the package manager's store is not what this test is about.
+run('pnpm', ['pack', '--pack-destination', packDir])
+const tgz = fs.readdirSync(packDir).find((f) => f.endsWith('.tgz'))
+if (!tgz) {
+	console.error('\n❌ pnpm pack produced no tarball')
+	process.exit(1)
+}
+
+fs.writeFileSync(
+	path.join(projectDir, 'package.json'),
+	`${JSON.stringify(
+		{
+			name: 'no-dotfiles-probe',
+			version: '0.0.0',
+			private: true,
+			type: 'module',
+			devDependencies: { '@rtorcato/repo-tooling': `file:${path.join(packDir, tgz)}` },
+		},
+		null,
+		2
+	)}\n`
+)
+run('git', ['init', '-q'], projectDir)
+// --ignore-scripts: no husky yet, and nothing here needs a consumer's postinstall.
+run('pnpm', ['install', '--ignore-scripts'], projectDir)
+if (!fs.existsSync(BIN)) {
+	console.error(`\n❌ tarball installed but ${BIN} is missing`)
+	process.exit(1)
+}
+
+// Snapshot the real HOME so any escape from the fake one is caught at the end.
+const homeBefore = fs.readdirSync(REAL_HOME).sort().join('\n')
+const realSkill = path.join(REAL_HOME, '.claude', 'skills', 'ai-issue-loop', 'SKILL.md')
+const realSkillBefore = fs.existsSync(realSkill) ? fs.readFileSync(realSkill, 'utf8') : null
+const noUserClaude = () => !fs.existsSync(path.join(fakeHome, '.claude'))
+
+// 2. setup — the first thing a consumer runs.
+console.log('\n── setup ──')
+const setup = cli('setup', '--preset', 'library', '--skip-install')
+assert(setup.status === 0, 'setup exits 0 with an empty HOME')
+assert(fs.existsSync(path.join(projectDir, 'biome.json')), 'setup scaffolded into the project')
+assert(noUserClaude(), 'setup created no ~/.claude')
+
+// 3. doctor — read-only; must report the absent skills dir rather than crash.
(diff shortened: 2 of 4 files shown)


rtorcato (commented):

**1 low / 1 info — no criticals, no highs, no mediums.**

**Shell construction — clean.** Every subprocess call in `scripts/integration/no-dotfiles.mjs` uses `execFileSync`/`spawnSync` with array args (`run('pnpm', ['pack', ...])`, `run('git', ['init', '-q'], projectDir)`, `cli()` → `spawnSync(BIN, argv, {...})`). No `shell: true`, no string-interpolated command. A space or metacharacter in a `mkdtemp`-generated path cannot break command parsing here.

**Sandbox containment — verified.** `projectDir`, `fakeHome`, `packDir` are all created via `fs.mkdtempSync(path.join(os.tmpdir(), ...))` at the top of the script. The CLI-under-test is only ever invoked through `cli()`, which passes `HOME`/`USERPROFILE` as a per-child-process `env` override to `spawnSync` — it never mutates `process.env` of the runner script itself, so `REAL_HOME = os.homedir()` captured at the top stays the genuine value for the later diff check. The one step that intentionally runs under the real `HOME` (`pnpm pack` in step 1) is commented as deliberate and only writes to `packDir`, a temp dir the script owns.

**Destructive ops — safe.** The only `fs.rmSync({ recursive: true, force: true })` call targets `[packDir, projectDir, fakeHome]` — all three are the exact `mkdtempSync` results from the top of the file, never an env var, CLI arg, or anything that could be empty/attacker-influenced. Cleanup is also skipped on failure (dirs left for debugging), so there's no destructive path taken on the error branch either.

**Supply chain — clean.** `pnpm pack` packs the local working tree to `packDir`; the scratch project's `package.json` points its only dependency at that tarball via `file:...`, so installing it never reaches the registry for the package under test. `pnpm install --ignore-scripts` runs with `cwd: projectDir` (via the `run()` helper's `cwd` param), so no lockfile or registry config in the actual repo is touched.

**CI job permissions — not fully verifiable from the diff.** The new `integration-no-dotfiles` job declares no `permissions:` or `secrets:` block of its own, and structurally mirrors the existing `integration` job in the same file (checkout → setup-node → setup-pnpm → install → build → run script), so it doesn't appear to request anything beyond whatever the workflow already grants other jobs. The workflow-level `permissions:` block itself isn't part of this diff, so I can't confirm the inherited scope from the diff alone — flagging as unverified rather than asserting either way.

**Info-level nit:** the "real HOME gained no entries" check (`fs.readdirSync(REAL_HOME).sort().join('\n')`) is a shallow, non-recursive top-level listing — a write into an *already-existing* nested directory under `REAL_HOME` wouldn't change that diff. The script does back this with a targeted, content-level check of the one concrete risk path (`~/.claude/skills/ai-issue-loop/SKILL.md` byte-identical), so the actual scenario the issue cares about is covered; the generic directory-listing check is just weaker defense-in-depth than its comment implies.

### Before merging

Nothing that should block. The CI job-permissions point above is worth a human glance at the top of `ci.yml` since it wasn't in the diff, but the new job introduces no new permission/secret request itself.

rtorcato (commented):

Walked every assertion in `scripts/integration/no-dotfiles.mjs` against what would actually make it fail:

- `setup` — checks exit 0, a real scaffolded file (`biome.json`), and `noUserClaude()`. Would catch a regression in any of the three.
- `doctor --json` — parses real JSON (falls back to `{}` on parse failure rather than throwing, so a malformed report degrades into a failed `Array.isArray` assertion instead of a crash — good), checks a non-empty `results` array and the specific `Claude skills: optional-missing` status.
- `fix --yes` (bare matrix) — the script never hardcodes a target list, it just invokes `fix --yes` and lets the CLI's own default scope decide, so it naturally tracks the matrix as it evolves rather than testing a stale subset. `strayed` computation resolves both relative and absolute reported paths against `projectDir` via `realpathSync`, so an escape to an arbitrary absolute path would be caught, not just a path-prefix string check.
- `fix claude-skills` refusal — checks `status !== 0` *and* the JSON body names `no-skills-dir`; not just an exit-code check.
- `--skills-dir` and the plain-`~/.claude/skills` path — both check real file existence, non-symlink via `lstatSync`, absence of the "wrote through a symlink" string, and that every reported path both exists and resolves inside the fake HOME. The `written.length > 0` assertion guards against the loop below it silently iterating zero times.

No vacuous assertions found — no empty-match globs, no swallowed try/catch, no assert inside a callback that's never invoked.

**HOME redirection**: the script spawns the installed bin directly via `spawnSync(BIN, argv, { env: { ...process.env, HOME: fakeHome, USERPROFILE: fakeHome } })` — not through a shell, so `PATH` lookup isn't a factor. The job runs on `ubuntu-latest` only, and Node's `os.homedir()` on POSIX consults `process.env.HOME` before falling back to a native `getpwuid` lookup, so redirecting `HOME` genuinely redirects anything in the CLI that resolves the home dir via `os.homedir()`. This would not hold unmodified on Windows, but the job doesn't run there, and `USERPROFILE` is set defensively anyway.

**Real HOME safety**: the pre/post verification (`homeBefore`, `realSkillBefore`) only calls `readdirSync`/`existsSync`/`readFileSync` — no writes anywhere near `REAL_HOME`. The `pnpm pack`/`git init`/`pnpm install` steps are deliberately run under the real `HOME` (package manager store), which is called out in a comment and is a reasonable, low-risk choice.

**Temp dirs / cleanup**: three `mkdtempSync` dirs. On success, all three are `rmSync`'d. On assertion failure, the script deliberately skips cleanup and prints the paths for debugging — reasonable given the CI runner is destroyed after the job either way. If an earlier `run()`/`cli()` call throws (e.g., `pnpm pack` fails outright) the process exits via an uncaught exception, which also leaves the dirs — same non-issue on an ephemeral runner, though it does mean local reruns of the script by a developer could accumulate `/tmp` entries over repeated failures. Minor, not worth blocking on.

**Two riders**: `git-identity.ts` diff only touches the block comment above `HOSTNAME_SUFFIXES` — the array literal itself (`['.local', '.matrix', '.localhost', '.lan', '.home', '.internal']`) is byte-identical. `SKILL.md` only removes the parenthetical sentence about `/_loop-status`; no other line in that paragraph changed. Neither is a behavior change.

One nit: `fix --yes` step both calls `fail(...)` with the detailed stray-file list and then `assert(strayed.length === 0, ...)` immediately after, which double-reports the same failure if it ever fires. Harmless, not worth a round-trip.

CI wiring (`ci.yml`) adds the job to both the `release` job's `needs:` list and its success-condition boolean, matching the pattern of the `integration`/`integration-swift` entries already present in the same file.

### Before merging

Nothing.

rtorcato:

## What to do

**Merge it**, after a glance at the two notes below. Both reviewers passed, `mergeStateStatus` is `CLEAN`.

`ai-notes` is deliberately left on — not because anything blocks, but because `security-expert` raised one question it could not answer from inside a diff and one real caveat, and both are answered here rather than left for you to re-derive.

## Resolving the reviewer's open question — CI job permissions

`security-expert` flagged that it could not confirm what `permissions:` the new `integration-no-dotfiles` job inherits, because the workflow-level block is not in the diff. Checked against the branch:

```console
$ git show origin/ai-460-no-dotfiles-audit:.github/workflows/ci.yml | grep -n 'permissions:'
424:    permissions:
```

**`ci.yml` has no workflow-level `permissions:` block at all.** The single occurrence is on the `release` job (line 424: `contents: write`, `issues: write`, `pull-requests: write`, `id-token: write`), which is a different job. So `integration-no-dotfiles` runs on the repository's default token permissions, declares none of its own, and requests no secrets — identical to the existing `integration` job beside it.

**Nothing to do here.** The concern is resolved, not deferred.

While confirming that, the PR's claim that the job is wired into the release gate also checks out — `needs.integration-no-dotfiles.result == 'success'` sits in the `release` job's condition, so a failure here actually blocks a publish rather than just going red in the list.

## The remaining note

The "real `HOME` gained no entries" assertion is a **shallow top-level directory listing**, so a change nested deeper inside the developer's real home would not be caught. `security-expert` noted this itself and also noted why it is acceptable: the one concrete risk file — the machine's own installed `ai-issue-loop/SKILL.md` — gets a targeted byte-identical check on top. The realistic scenario is covered; the general claim is broader than the assertion behind it.

Worth knowing, not worth blocking.

This repo has no `release` environment with `required_reviewers` (the probe returns `0`), so merging `main` fires semantic-release with nobody in the loop and only Dependabot PRs merge unattended here. That gate is #429, still `holding` — it needs you to name who the reviewers should be.

The squash subject is `test(ci):`, so **merging this cuts no release.**

## Two things in the PR body that need your decision

Neither is a defect in this PR; the implementer correctly left both alone.

1. **`doctor` has no `--skills-dir` option.** `src/base/checks.ts:513` calls `claudeSkillStatus()` with no argument, so that function's `explicit` parameter is dead and `doctor` always reports against `~/.claude/skills`. A consumer using `--skills-dir` gets a permanent `optional-missing` for a skill they have installed. Read-only, so it is a reporting asymmetry rather than a stray write — but it is a genuine gap in exactly the "works without the maintainer's dotfiles" property #460 is about. Adding the flag is a CLI-surface change.

2. **`src/cli/generators/docs-site.ts:391`** emits `uses: rtorcato/repo-tooling/.github/workflows/docs-deploy.yml@main` into the *consumer's* CI — unpinned, so any push to `main` here changes their build. Pinning to a tag or SHA is conventional but needs a bump story.

Also worth noting: the implementer **declined** one of the issue's own seed fixes with a better reason than the issue gave. On `.matrix` in `HOSTNAME_SUFFIXES` it reworded the "this user's LAN" comment but kept the suffix, because removing it narrows detection for a case that produced real commits in the #327 set, and the array is a heuristic where an extra entry only widens coverage. That is the right call.

Issue #461, opened 2026-08-20
setup --preset is all-or-nothing and silently skips the wizard on a fresh project


rtorcato:
## What happened

Ran the documented non-interactive path on an empty directory:

```bash
npx @rtorcato/repo-tooling@3.15.0 setup --preset library -d . --skip-install
```

It scaffolded ~28 files — semantic-release, commitizen, husky, commitlint, knip,
size-limit, codecov, Dependabot, CodeQL, AGENTS.md/CLAUDE.md/Cursor/Copilot rules —
in an interactive terminal, with no confirmation, no file list shown up front, and
no way to decline any of it.

For a brand-new project that's the moment the answers matter most, and it's the one
moment the tool doesn't ask.

## Why it's structural, not just a bad preset choice

`src/cli/commands/setup-presets.ts:22` — `BASE` hardcodes five opinionated tools:

```ts
const BASE: Omit<ProjectConfig, 'projectName' | 'projectType' | 'typescript' | 'bundler'> = {
	// ...
	gitHooks: true,
	commitLint: true,
	semanticRelease: false,
	securityAutomation: true,
	badges: true,
	aiSetup: true,
}
```

Every `case` in `buildPresetConfig` spreads `BASE` and only ever *adds* to it.
Nothing subtracts. So there is no preset — and no flag — that yields a small repo.
`--preset` isn't just opinionated, it's structurally incapable of producing less.

## Proposal

1. **Make `--preset` TTY-aware.** In an interactive terminal without `--yes`,
   print the resolved file list (`computeFileList` already produces it) and let the
   user deselect before anything is written. Non-TTY — i.e. CI, the case `--preset`
   was actually added for — keeps today's one-shot behavior unchanged, so this
   isn't a breaking change to the documented workflow.

2. **Add a `minimal` preset.** tsconfig + biome + vitest, nothing else. Built
   without spreading `BASE`, the way `swift-library` already does.

Together: `--preset` stays the CI escape hatch it was designed to be, and a human
on a fresh directory can no longer end up with a release pipeline they never asked for.

## Also noticed on the same run

Not filing separately unless you want them split out:

- The scaffold's own output fails its own `biome check` — `biome.json`, `knip.json`,
  `tsup.config.ts`, `vitest.config.ts` and `.claude/settings.json` all needed
  `check:fix` (expanded arrays, spaces vs tabs). A fresh scaffold should be
  format-clean out of the box.
[...]

PR #466, opened 2026-08-21
fix(setup): use inquirer's `select` type so the wizard can run at all

### LINKED ISSUE #463: Interactive `setup` wizard is completely broken: inquirer v14 removed type: 'list'
## Severity

The interactive wizard — the headline command in the README — **cannot complete a
single prompt** in the published package. `npx @rtorcato/repo-tooling@latest setup`
crashes on its first question. `--preset` is currently the only working path into
`setup`.

## Reproduction

```
$ npx -y @rtorcato/repo-tooling@latest setup

🛠️  Welcome to JS Tooling Setup!

Setting up tooling in: /Users/rtorcato/Desktop/untitled folder

❌ Setup failed: UnknownPromptTypeError: Prompt type "list" is not registered.
Available prompt types: checkbox, confirm, editor, expand, input, number,
password, rawlist, search, select
    at PromptsRunner.prepareQuestion (inquirer/dist/ui/prompt.js:156:19)
    at PromptsRunner.fetchAnswer (inquirer/dist/ui/prompt.js:202:37)
    at PromptsRunner.run (inquirer/dist/ui/prompt.js:118:47)
    at async promptForLanguage (@rtorcato/repo-tooling/dist/cli/commands/setup.js:111:26)
    at async promptForConfig (@rtorcato/repo-tooling/dist/cli/commands/setup.js:142:22)
    at async Command.setupProject (@rtorcato/repo-tooling/dist/cli/commands/setup.js:59:24)
```

## Cause

`package.json:237` pins `"inquirer": "^14.0.2"`. Inquirer removed the `list` prompt
type in v10 — it was renamed to `select`. `confirm`, `input`, `checkbox` and
`rawlist` all survived the rename, which is why nothing else breaks and why this
slipped through.

`src/cli/commands/setup.ts` still uses `type: 'list'` in 11 places:

```
195, 276, 294, 321, 333, 353, 366, 391, 436, 459
```

Line 195 is inside `promptForLanguage`, the very first prompt — so every run dies
immediately, regardless of what the user would have answered.

## Fix

Mechanical rename of `type: 'list'` → `type: 'select'` at those 11 sites. The
choice shape (`{ name, value }`) is unchanged between the two, so nothing else
needs to move.

## The part worth fixing beyond the rename

`tests/cli/commands/setup.test.ts` is green against a wizard that cannot run at
all. Whatever it does with inquirer, it isn't exercising real prompt dispatch — so
a major-version bump of a prompt library silently shipped a dead headline command.

A smoke test that drives `setup` end-to-end against the real inquirer (piping
scripted answers, or asserting `promptForConfig` resolves) would have caught this
at the dependency bump. Without one, the same class of break returns on the next
inquirer major.

## Related

- #461 — `setup --preset` is all-or-nothing and skips the wizard. Worth noting the
  two interact badly: #461 argues a human on a fresh project should land in the
  wizard rather than a preset, and this issue is why they currently can't.

Environment: macOS, node via npx, `@rtorcato/repo-tooling@3.15.0`.

rtorcato:
🤖 *Written by Claude (AI agent) and posted under the owner's account. There is no separate GitHub account for AI agents, so this appears under @rtorcato's avatar.*

Closes #463

## The fix

`type: 'list'` → `type: 'select'`. **10 sites**, not the 11 the issue estimated — all in `src/cli/commands/setup.ts`, and `grep -rn "type: 'list'" src/` now returns nothing.

The `{ name, value }` choice shape is identical between the two types, so nothing else moved.

**No other prompt type is affected.** The only inquirer types this codebase uses are `list`, `confirm` and `input`; v14 registers `confirm` and `input` unchanged. The other `type:` values that turn up in a grep (`boolean`, `string`, `object`, `json`, `module`, `autogenerated`) belong to config schemas, not inquirer.

## Why it shipped, and the part that stops it recurring

Every test in `tests/cli/commands/setup.test.ts` does this:

```ts
const spy = vi.spyOn(inquirer, 'prompt')
```

which replaces the dispatcher wholesale. No prompt type is ever resolved, so 33 tests passed against a wizard that could not ask a single question.

The new test checks the questions against the **installed inquirer's own registry** rather than a list written down in the test:

```ts
const registered = new Set(Object.keys(inquirer.createPromptModule().prompts))
```

so it tracks the dependency instead of drifting from it. Two guards against it passing for the wrong reason:

- `expect(registered.has('select')).toBe(true)` — an empty registry would otherwise wave everything through.
- `expect(asked.length).toBeGreaterThan(5)` — zero collected questions would otherwise satisfy the assertion vacuously.

## Confirmed it can actually fail

Reverted a site and ran it — twice, deliberately, in two different `inquirer.prompt` calls, since collecting only from the first call would still catch the language prompt while missing the nine questions in the JS array:

```
# reverted promptForLanguage (call 0)
AssertionError: expected [ 'language: list' ] to deeply equal []

# reverted projectType, inside the JS question array (call 1)
AssertionError: expected [ 'projectType: list' ] to deeply equal []
```

Both restored. The failure names the offending question, so the next person gets told which one rather than just that something is wrong.

## Scope

Deliberately untouched: `src/cli/commands/setup-presets.ts` and anything to do with TTY-awareness or a `minimal` preset. That is #461, which is held behind this PR precisely because routing people into the wizard is pointless while the wizard crashes.

One thing worth flagging about this repo's own hygiene rather than this diff: `pnpm lint` covers `src scripts` only, so `tests/` is not formatted by the pre-commit hook and ~20 test files are currently unformatted relative to `biome check`. Running `biome check --write tests` touches all of them. Not fixed here — it would bury a 43-line fix under a few hundred lines of reformatting — but it is a real gap, and the pre-commit hook will not catch a badly formatted test.

## Verification

`vitest run` — 996 passed, 18 skipped, 62 files. `biome lint --config-path=biome.json src scripts` clean. `tsc --noEmit --project src/cli/tsconfig.json` clean.

### DIFF
--- a/src/cli/commands/setup.ts
+++ b/src/cli/commands/setup.ts
@@ -192,7 +192,7 @@ async function promptForLanguage(targetDir: string): Promise<LanguageModule> {
 	const detected = await detectLanguage(targetDir)
 	const { language } = await inquirer.prompt([
 		{
-			type: 'list',
+			type: 'select',
 			name: 'language',
 			message: '🗣️  What is the primary language of this repo?',
 			choices: Object.values(LANGUAGES).map((module) => ({
@@ -273,7 +273,7 @@ async function promptForConfig(targetDir: string): Promise<ProjectConfig | null>
 			validate: (input: string) => input.trim().length > 0 || 'Project name is required',
 		},
 		{
-			type: 'list',
+			type: 'select',
 			name: 'projectType',
 			message: '🏗️  What type of project are you building?',
 			choices: [
@@ -291,7 +291,7 @@ async function promptForConfig(targetDir: string): Promise<ProjectConfig | null>
 			default: true,
 		},
 		{
-			type: 'list',
+			type: 'select',
 			name: 'tsConfig',
 			message: '⚙️  Which TypeScript configuration?',
 			choices: (answers: any) => {
@@ -318,7 +318,7 @@ async function promptForConfig(targetDir: string): Promise<ProjectConfig | null>
 			when: (answers: any) => answers.useTypeScript,
 		},
 		{
-			type: 'list',
+			type: 'select',
 			name: 'lintingTool',
 			message: '🔍 Which linting/formatting tool?',
 			choices: [
@@ -330,7 +330,7 @@ async function promptForConfig(targetDir: string): Promise<ProjectConfig | null>
 			default: 'biome',
 		},
 		{
-			type: 'list',
+			type: 'select',
 			name: 'eslintConfig',
 			message: '🔧 Which ESLint configuration?',
 			choices: (answers: any) => {
@@ -350,7 +350,7 @@ async function promptForConfig(targetDir: string): Promise<ProjectConfig | null>
 			when: (answers: any) => answers.lintingTool !== 'none',
 		},
 		{
-			type: 'list',
+			type: 'select',
 			name: 'testingFramework',
 			message: '🧪 Which testing framework?',
 			choices: [
@@ -363,7 +363,7 @@ async function promptForConfig(targetDir: string): Promise<ProjectConfig | null>
 			default: 'vitest',
 		},
 		{
-			type: 'list',
+			type: 'select',
 			name: 'testEnvironment',
 			message: '🌍 Test environment?',
 			choices: [
@@ -388,7 +388,7 @@ async function promptForConfig(targetDir: string): Promise<ProjectConfig | null>
 			when: (answers: any) => answers.gitHooks,
 		},
 		{
-			type: 'list',
+			type: 'select',
 			name: 'releaseTool',
 			message: '🚀 Automated release tool?',
 			choices: [
@@ -433,7 +433,7 @@ async function promptForConfig(targetDir: string): Promise<ProjectConfig | null>
 			default: true,
 		},
 		{
-			type: 'list',
+			type: 'select',
 			name: 'orchestrator',
 			message: '🚀 Monorepo task orchestrator?',
 			choices: [
@@ -456,7 +456,7 @@ async function promptForConfig(targetDir: string): Promise<ProjectConfig | null>
 				answers.projectType === 'nextjs-app',
 		},
 		{
-			type: 'list',
+			type: 'select',
 			name: 'bundler',
 			message: '📦 Which bundler/build tool?',
 			choices: (answers: any) => {
--- a/tests/cli/commands/setup.test.ts
+++ b/tests/cli/commands/setup.test.ts
@@ -498,4 +498,37 @@ describe('setup language prompt', () => {
 		const lock = await fs.readJson(join(dir, '.repo-tooling.json'))
 		expect(lock.config.language).toBe('js')
 	})
+
+	// #463: every test above spies on `inquirer.prompt`, which replaces the
+	// dispatcher wholesale — so they all passed against a wizard that could not
+	// ask a single question. inquirer v10 renamed `list` to `select`, v14 shipped
+	// in this repo, and `npx repo-tooling setup` died on its first prompt with
+	// `UnknownPromptTypeError` while this suite stayed green.
+	//
+	// The fix is to check the questions against the *installed* inquirer's own
+	// registry rather than a list written down here, so the next rename fails at
+	// the dependency bump instead of in a user's terminal.
+	it('only uses prompt types the installed inquirer registers', async () => {
+		const registered = new Set(Object.keys(inquirer.createPromptModule().prompts))
+		// Guard the oracle itself: an empty registry would pass everything.
+		expect(registered.has('select')).toBe(true)
+
+		const dir = newTmpDir()
+		const spy = mockPrompt({ language: 'js' }, JS_ANSWERS)
+		const logSpy = vi.spyOn(console, 'log').mockImplementation(() => {})
+		try {
+			await setupProject({ directory: dir, skipInstall: true })
+		} finally {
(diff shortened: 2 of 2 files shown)


rtorcato (commented):

Diff is a mechanical `type: 'list'` → `type: 'select'` rename across 10 inquirer question objects in `src/cli/commands/setup.ts`, plus one new test in `tests/cli/commands/setup.test.ts`. Confirmed against the diff (not just the PR prose):

- **Scaffolding behavior unchanged.** Every hunk touches only the `type:` field; the `{ name, value }` choice arrays, `default`, `when`, and `validate` clauses are untouched in the diff. Same answer keys resolve to the same values, so `setup` writes the same config for the same input as before — this doesn't silently re-route what gets scaffolded.
- **No untrusted-input-to-shell/filename path in this diff.** Nothing here touches how a typed project name or answer reaches a filename or a shell/exec call; the change is confined to prompt-type strings.
- **New test doesn't introduce new fs/env surface.** It reuses the same `setupProject({ directory: dir, ... })` / temp-dir pattern already used by the passing tests above it in the same file — no new write target, no new dependency, no network or env mutation added.
- **No dependency/supply-chain change.** `inquirer` is already a pinned dependency (`^14.0.2` per the linked issue); nothing in `package.json` or lockfile is touched by this diff.
- **Issue-vs-diff count matches.** Issue #463 estimated 11 sites; diff shows 10 renamed hunks, consistent with the PR's claim that a grep for `type: 'list'` now returns nothing.

No injection, secret-leak, unsafe shell/SQL, or supply-chain concerns in this change.

### Before merging

Nothing.

rtorcato (commented):

**Rename completeness/correctness** — Diff shows exactly 10 hunks, each changing only the `type: 'list'` → `type: 'select'` line; every surrounding `name`, `message`, `choices`, `default`, `when`, `validate` field is untouched context. That matches the issue's own enumerated line list (195, 276, 294, 321, 333, 353, 366, 391, 436, 459 — which is 10 numbers, despite the issue's prose saying "11 places"). So the PR's "10, not 11" claim is correct; the discrepancy originates in the issue's miscount, not in this fix. No unrelated site was missed or altered.

**Would the new test fail on a revert?** Yes. `spy.mock.calls.flatMap(([questions]) => questions)` iterates every recorded call, not just the first — `promptForLanguage` calls `inquirer.prompt` once (call 0, the single-question array) and `promptForConfig` calls it once with the ~10-question array (call 1), so both are collected. That matches the PR's description of reverting one site in each call and getting a targeted `AssertionError` naming the offending question. Reasoned independently from the diff, this holds.

**Oracle soundness** — `registered = new Set(Object.keys(inquirer.createPromptModule().prompts))` reflects the same internal registry inquirer itself consults to throw `UnknownPromptTypeError`, so it's a reasonable source of truth rather than a hand-maintained list. The `registered.has('select') === true` guard catches the case where that registry comes back empty or renamed (an empty set would otherwise cause `!registered.has(q.type)` to flag every real question, which actually fails loudly rather than passing vacuously — but the explicit guard makes the intent unambiguous and fails fast on the oracle itself). `asked.length > 5` guards against the empty-`asked` case, which is the real vacuous-pass risk (if no questions were collected, `.toEqual([])` would trivially hold). Both guards are doing real work, not decorative.

**Flakiness/order-dependence** — The test reuses the same `vi.spyOn(inquirer, 'prompt')` + `mockPrompt` pattern already used by the other 33 tests in this file, so it isn't introducing a new mocking approach. The `console.log` spy is restored in a `finally`, so it can't leak state on assertion failure either.

No defects found.

### Before merging
Nothing.

rtorcato:

## What to do

**Merge it.** Both reviewers returned `### Before merging: Nothing.`, every check is green, and `mergeStateStatus` is `CLEAN`.

⏳ This unblocks **#461**, which is deliberately held behind it — re-add `ai-ready` there once this lands.

This repo has no `release` environment with `required_reviewers` (the probe returns `0`), so merging `main` fires semantic-release and publishes to npm with nobody in the loop. Only Dependabot PRs merge unattended here. That gate is #429, still `holding` — it needs you to name who the reviewers should be.

The squash subject is `fix(setup):`, so **merging this cuts a patch release** — which is the point: the broken wizard is in the published package right now.

## What the reviewers actually checked

Both were briefed on the specific risk rather than sent in cold, because the diff is trivial and the test is the substance.

`code-reviewer` verified the two claims that matter and reasoned them out independently rather than taking the PR's word:

- **The rename is complete and inert.** Ten hunks, each changing only the `type:` line, with every surrounding `name`, `message`, `choices`, `default`, `when` and `validate` field untouched as context.
- **The test genuinely fails on a revert.** `spy.mock.calls.flatMap(...)` iterates every recorded call, so both `promptForLanguage`'s single-question array and `promptForConfig`'s ~10-question array are collected — not just the first.

It also resolved the count discrepancy: the issue's prose says "11 places" but its own enumerated line list (195, 276, 294, 321, 333, 353, 366, 391, 436, 459) contains **10 numbers**. The miscount is in the issue, not the fix.

`security-expert` returned `Nothing.` — proportionate for a prompt-type rename.

## One thing this PR raises but does not fix

`pnpm lint` covers `src scripts` only, so `tests/` is never formatted by the pre-commit hook and roughly 20 test files are currently drifted from `biome check`. Running `biome check --write tests` touches all of them.

It was left out of this PR on purpose — a few hundred lines of reformatting would bury a 43-line fix — but it is a real gap: a badly formatted test cannot be caught by the hook that exists to catch exactly that.

Issue #468, opened 2026-08-21
Every @biomejs/biome bump fails the build: nothing updates the $schema pin to match


rtorcato:
🤖 *Filed by Claude (AI agent) on rtorcato's behalf, from a `repo-tooling` session. Posted under the owner's account; there is no separate GitHub account for AI agents.*

## What happens

Every Dependabot bump of `@biomejs/biome` fails the `build` job, because nothing updates the `$schema` URL in `biome.json` to match.

Observed on #455 (`dev-minor`, 18 packages), where every other check passed:

```
❌ dogfood: doctor found 1 genuine finding(s) on this repo:

   [drift] Biome — biome.json targets Biome 2.5.5 but @biomejs/biome 2.5.8 is installed
```

That is #435's check working correctly — `biome.json` pins `.../2.5.5/schema.json` while the bump moved the package to 2.5.8. Dependabot updates the dependency; it has no idea a config file elsewhere restates the version as a URL.

## Why it needs fixing rather than fixing-up

The repair is already one command — `checkBiomeSchema`'s own hint says so:

```
hint: 'Run `npx @rtorcato/repo-tooling fix biome` to rewrite the $schema URL'
```

The problem is *who runs it*. On a Dependabot PR nobody does, so the branch sits red until a human checks it out, runs `fix biome`, commits and pushes — and **pushing to a Dependabot branch makes Dependabot stop rebasing it**, so the cheap fix costs you the bot's management of that PR.

This recurs on every Biome release. It is not a one-off on #455.

## Scope — narrower than it first looks

Worth stating plainly so nobody over-builds this: since #462, `@biomejs/biome` is recognised as a `peerDependency` and its bumps **never auto-merge anyway** — the gate stands down and hands them to a human. Verified on this same run:

```
##[notice]@biomejs/biome ships to consumers — leaving this PR for a human
```

So this does not block the unattended path; that path was already closed for Biome. What it blocks is the **human** path: they cannot merge a red PR either, and the redness is a config file the tooling can rewrite itself.

## Options

| Option | Notes |
|---|---|
| **Stop pinning a version in `$schema`** | If Biome accepts an unversioned or `latest` schema URL, the drift cannot occur and #435's check becomes redundant for this repo. Needs checking against what Biome actually resolves — a stale-but-pinned schema is at least deterministic. |
[...]

PR #469, opened 2026-08-21
fix(biome): mark the shipped preset non-root so a whole-repo check can run

### LINKED ISSUE #467: Lint scope is narrowed to src scripts because the shipped Biome preset blocks a whole-repo check
🤖 *Filed by Claude (AI agent) on rtorcato's behalf, from a `repo-tooling` session. Posted under the owner's account; there is no separate GitHub account for AI agents.*

Raised while working #463 (PR #466) and deliberately not fixed there — a few hundred lines of reformatting would have buried a 43-line bug fix.

## The symptom

Every lint script in this repo is scoped to `src scripts`:

```json
"lint":      "pnpm exec biome lint --config-path=biome.json src scripts",
"format":    "pnpm exec biome format --config-path=biome.json src scripts",
"check":     "pnpm exec biome check --config-path=biome.json src scripts",
"check:fix": "pnpm exec biome check --config-path=biome.json --fix src scripts",
```

Both gates use them, so neither catches anything else:

- `.husky/pre-commit` runs `pnpm run check`
- the CI `lint` job (`.github/workflows/ci.yml:97`) runs `pnpm lint`

**`tests/` is 62 files that nothing checks.** 20 are currently drifted — 19 formatting, and one real rule violation (`lint/style/useTemplate`). Running `biome check --write tests` touches all 20 in one go.

## The cause — it is a workaround, not an oversight

`biome check .` does not merely find problems, it **refuses to run**:

```console
$ biome check --config-path=biome.json .
/…/repo-tooling/tooling/biome/biome.json configuration ━━━━━━━━━━━━━━━━━━

  × Found a nested root configuration, but there's already a root configuration.

  i The other configuration was found in /…/repo-tooling.

  × Biome exited because the configuration resulted in errors.
```

`tooling/biome/biome.json` — **the preset this package ships to consumers** — is itself a root config, and Biome 2.x rejects a nested root. So the `src scripts` narrowing is what someone reached for to get a green gate, and it silently traded away every file outside those two directories.

## Likely fix, verified

Adding `"root": false` to `tooling/biome/biome.json` clears the error. Probed locally and reverted:

| | Before | After `root: false` |
|---|---|---|
| `biome check .` | configuration error, 0 files | **Checked 225 files** |
| `biome check src scripts` | Checked 73 files, clean | Checked 73 files, clean |

So `extends` still resolves for consumers — the narrow path is unaffected — and the whole repo becomes checkable.

That is the *enabling* change. What it exposes is a second, larger piece of work: **36 errors and 62 warnings** across the 225 files, concentrated as

- `tests/` — 19 format + 1 `lint/style/useTemplate`
- `apps/` — mostly CSS: `lint/complexity/noImportantStyles` ×12, `lint/style/noDescendingSpecificity`
- `tooling/`, `.claude/` — a file each

Worth staging: land `root: false` plus the scope widening and the mechanical `--write` pass separately from any rule violation that needs a judgement call (the `!important` usages in `apps/` may well be deliberate and want a targeted suppression rather than a rewrite).

## Why this matters beyond tidiness — the repo does not dogfood what it ships

`src/cli/generators/package-json.ts:93-96` and `src/languages/js/fixers.ts:41-44` emit **whole-repo** scripts into every consumer:

```ts
scripts['lint']      = 'biome lint .'
scripts['format']    = 'biome format .'
scripts['check']     = 'biome check .'
scripts['check:fix'] = 'biome check --fix .'
```

So `repo-tooling` hands consumers `biome check .` while running `biome check src scripts` on itself — and the reason it cannot follow its own standard is a file it publishes. That is the part worth fixing regardless of how the drifted files are handled.

**Consumer impact needs confirming rather than assuming.** A consumer gets the preset at `node_modules/@rtorcato/repo-tooling/tooling/biome/biome.json`, and Biome ignores `node_modules`, so `biome check .` in a normal consumer repo should be fine. Worth verifying — the new `integration-no-dotfiles` job (#465) scaffolds a project and could assert the emitted `check` script actually passes on fresh output, which would catch this class for consumers the way nothing currently catches it here.

## Also surfaced

Running the whole-repo check printed a Biome version notice: the shipped preset pins `$schema` **2.5.0** while the root `biome.json` pins **2.5.5**. #435 shipped a `doctor` check for exactly this drift — worth confirming whether it covers the *shipped preset's own* pin, or only a consuming repo's.

## Related

- #466 / #463 — where this was noticed
- #435 — `doctor` detects Biome `$schema` drift from the installed CLI
- #465 — the empty-`HOME` integration job that could assert emitted scripts pass

rtorcato:
🤖 *Written by Claude (AI agent) on rtorcato's behalf, via `ai-issue-loop`. Posted under the owner's account; there is no separate GitHub account for AI agents.*

Closes #467.

## What this does

`tooling/biome/biome.json` — the preset this package publishes — was itself a **root** config. Biome 2.x refuses to run when it finds a nested root inside a project that already has one, so `biome check .` did not merely report problems, it exited on a configuration error before checking a single file. The lint scripts were narrowed to `src scripts` to get a green gate, which silently left everything else unchecked.

Adding `"root": false` to the preset clears it, and the scripts widen to the whole repo:

```diff
-"lint":      "pnpm exec biome lint  --config-path=biome.json src scripts",
-"format":    "pnpm exec biome format --config-path=biome.json src scripts",
-"check":     "pnpm exec biome check --config-path=biome.json src scripts",
-"check:fix": "pnpm exec biome check --config-path=biome.json --fix src scripts",
+"lint":      "pnpm exec biome lint .",
+"format":    "pnpm exec biome format .",
+"check":     "pnpm exec biome check .",
+"check:fix": "pnpm exec biome check --fix .",
```

That is now character-for-character what `src/cli/generators/package-json.ts:93-96` and `src/languages/js/fixers.ts:41-44` emit to consumers, which was the point of the issue — the repo did not dogfood its own scripts. `--config-path=biome.json` was redundant (Biome discovers `./biome.json` from the cwd anyway); dropping it verified identical: 228 files either way. **That local verification missed a CI failure — see the Fix round section at the bottom.** The flag stays dropped; it was proven not to be the cause.

`apps/docs/biome.json` already carried `"root": false`, so this is an existing pattern in the repo, not a new one.

## Verification, both directions

| | before | after |
|---|---|---|
| `biome check .` | configuration error, **0 files** | **228 files**, exit 0 |
| `biome check --config-path=biome.json src scripts` | 75 files, clean | 75 files, clean |
| preset copied to a bare project root (what `copy biome` does) | n/a | loads and runs, no error |

That last row mattered enough to test directly: `copy biome` writes the preset as a **consumer's own root `biome.json`**, so `root: false` had to be proven harmless with no parent root above it. It is — Biome does not require a root config to exist.

Full suite: `biome check .` / `lint .` / `format .` all exit 0, `tsc --noEmit` clean, **996 tests pass** (61 files, 1 skipped), `build-cli` clean. CI's lint job (`pnpm lint`) and the `pre-push` `pnpm verify` both pass.

## Why `fix:` and not `chore:`

This edits a **published** artifact (`tooling/biome/biome.json`, exported as `@rtorcato/repo-tooling/biome`), and it repairs a real defect in it rather than tidying the repo. Any consumer that has the preset *inside* their project tree rather than resolved from `node_modules` — a workspace link, a `file:` dependency, a vendored copy — hits exactly the error this repo hit, and cannot run a whole-repo check. `chore:` would ship that fix to nobody. `fix:` cuts a patch, which is the right blast radius. Flagging it explicitly since the squash subject decides the release.

## The three commits

Kept separate so the reformatting does not bury the substance:

1. `badcf3b` **fix(biome)** — `root: false` + the widened scripts + the `AGENTS.md` line that documented the old narrow command. 3 files.
2. `580a302` **style** — the mechanical `biome format --write .` pass. **34 files touched**, 26 of them under `tests/`. No logic changes: the whitespace-insensitive diff is nothing but double→single quotes, dropped semicolons and re-wrapping. These files were written to Prettier defaults and had simply never been checked.
3. `eebf346` **fix(lint)** — the two actual rule violations (below).

38 files total, +521 / −562.

Note that the format pass also reformats several *shipped* preset files (`tooling/changesets/config.json`, `tooling/nx/nx.json`, `tooling/oxlint/oxlintrc.json`, `tooling/release-please/*.json`, `tooling/commitlint/commitlint.mjs`, `tooling/rolldown/rolldown.config.mjs`) from tabs/double-quotes to the preset's own style. Semantically identical — the tests that read them use `fs.readJson`/import rather than byte comparison, and asset hashes are recorded per-consumer at copy time, not against a committed baseline. It does mean published bytes change.

## Rule violations fixed

The issue predicted "19 format + 1 `useTemplate`". The real whole-repo run found **one error the issue did not mention**:

- **`lint/a11y/noSvgWithoutTitle`** in `favicon.svg` — the only hard error left once formatting was done. Fixed by adding `<title>repo-tooling</title>` rather than suppressing: it changes no rendering, and it is genuinely correct if the mark is ever inlined into a document.
- **`lint/style/useTemplate`** in `tests/cli/e2e.test.ts:67` — `stderr + ''` → `String(stderr)`.

## For your decision — the 62 CSS warnings

**No suppression, no rule-disable and no `files.includes` exclusion was needed, because these are warnings and warnings do not fail `biome check`.** Verified directly: `biome lint` over the offending CSS alone exits **0**; only the a11y error above ever produced a non-zero exit. So CI is green with these left fully visible, which seemed strictly better than hiding them.

That is the part I want a maintainer to confirm rather than decide myself. The counts are also **much larger than the issue estimated** — it said `noImportantStyles ×12` and `noDescendingSpecificity ×1`; the actual run reports:

| rule | count |
|---|---|
| `lint/complexity/noImportantStyles` | **46** warnings |
| `lint/style/noDescendingSpecificity` | **16** warnings |

by file:

| file | warnings |
|---|---|
| `tooling/docusaurus/theme.css` | 29 |
| `apps/docs/src/css/theme.css` | 29 |
| `apps/docs/src/css/custom.css` | 2 |
| `apps/docs/src/pages/index.module.css` | 2 |

The two `theme.css` files are **byte-identical** (`cmp` confirms) — `tooling/docusaurus/theme.css` is shipped and `apps/docs/src/css/theme.css` mirrors it — so it is really 31 distinct warnings counted twice, plus 2 more.

I deliberately did not touch these. At 46 sites this is far past the "couple of near-identical suppressions" threshold: `!important` in a Docusaurus theme is overriding vendor CSS you do not control, so it is very likely deliberate, and 46 `biome-ignore` comments would be noise. Whether to leave them as warnings (current state), disable the two rules for `**/*.css` in the preset, or actually rewrite the specificity is a maintainer call — and since one of the files is **shipped**, changing it affects consumers.

## The `$schema` notice (#468 territory — not fixed here)

Asked to check whether this change affects its visibility: **it makes it more visible.** Measured:

- old narrow scope → notice **not printed**
- `biome check .` → notice **printed**, as the single remaining `info`

The reason is specific: the notice is not emitted by `extends` resolution. It fires because `tooling/biome/biome.json` is now **in scope as a file being checked**, so Biome deserializes it as a config and reports that its pinned `2.5.0` does not match the CLI's `2.5.5`. Confirmed by checking that one file alone (notice appears) versus `biome check src` (silent).

It is an `info`, so it does **not** fail the build — `pnpm run check` exits 0. But it now prints on every run, so #468 is somewhat more worth doing after this lands. Left alone as instructed.

Worth noting for #435's scope: the same notice appears when the preset is copied into a bare consumer project, so the drift is consumer-visible too, not just internal.

---

## Fix round — the CI lint failure

CI's `lint` job failed on `pnpm exec biome lint .`:

```
× Failed to resolve the configuration from @rtorcato/repo-tooling/biome
  Caused by: Could not resolve @rtorcato/repo-tooling/biome: module not found
```

Fixed in `b8bf1ef` — a **one-line change to `apps/docs/biome.json`**.

### It was not `--config-path`, and restoring it would not have helped

The obvious suspect was the dropped `--config-path=biome.json`. It is not the cause. I reproduced the failure locally and tested that hypothesis directly — see the controls below: **with the flag restored, the failure is identical.** So the flag stays dropped and the scripts stay character-for-character identical to what the package emits. **This PR introduces no divergence from the emitted scripts** — the gap #467 was filed about is still closed.

### Actual root cause

Widening the scope from `src scripts` to `.` made Biome descend into `apps/docs` for the first time and load its nested `apps/docs/biome.json`. That file extends `@rtorcato/repo-tooling/biome`, which is unresolvable from there:

- The nearest `package.json` walking up from `apps/docs/` is `apps/docs/package.json`, named `@rtorcato/docs` — so Node-style **self-reference does not apply**. (It *does* apply at the repo root, where the nearest `package.json` is `@rtorcato/repo-tooling` itself.)
- `apps/docs` neither declares nor is linked to the tooling package: `apps/docs/node_modules/@rtorcato/` contains only `shared-docs`.
- The CI `lint` job never runs `pnpm install`. It restores `~/.pnpm-store` + **root `node_modules` only** from `actions/cache` (`.github/workflows/ci.yml:63-66`), so `apps/docs/node_modules` does not even exist there.

This was a **latent breakage in `apps/docs/biome.json`, pre-existing on `main`** — the narrow `src scripts` scope simply never reached it. Widening the scope is what exposed it, which is precisely what #467 was for.

**Why it passed locally:** the agent worktree symlinks `apps/docs/node_modules` into the main checkout (`.claude/settings.json` `worktree.symlinkDirectories`). That symlink is what made resolution succeed on my machine and nowhere else. Local success was the bug, not the evidence.

### The fix

```diff
 // apps/docs/biome.json
-  "extends": ["@rtorcato/repo-tooling/biome"],
+  "extends": ["../../tooling/biome/biome.json"],
```

A relative path needs no module resolution at all. The **root `biome.json` deliberately keeps** `extends: ["@rtorcato/repo-tooling/biome"]` — that is the shape a consumer writes, it is the dogfooding surface that matters, and it resolves via self-reference against the root `package.json` `exports` with **zero `node_modules` present** (proved below). `apps/docs` is not a consumer; the package specifier only ever "worked" there by accident.

### Proof against the real failure mode, not just locally

"Passes locally" is what produced this bug, so I reproduced CI's tree shape instead: export **only git-tracked files** (`git ls-files | tar`) into a clean directory, `git init`, and **no `node_modules` anywhere** — root or `apps/docs`. Then run the same Biome 2.5.5 binary.

Controls, all in that node_modules-free tree:

| # | tree state | command | result |
|---|---|---|---|
| A | before fix | `biome lint .` | ✗ **exact CI error reproduced**, byte-for-byte |
| B | before fix | `biome lint --config-path=biome.json .` | ✗ **same failure** — refutes the flag as the cause |
| C | before fix | `biome lint --config-path=biome.json src scripts` | ✓ 75 files — why `main` is green: never reaches `apps/docs` |
| D | before fix, `apps/docs/biome.json` deleted | `biome lint .` | ✓ 4 files — isolates the culprit to that one file |
| E | **after fix** | `biome lint .` | ✓ **228 files, exit 0, zero resolution errors** |
| F | **after fix** | `biome check .` | ✓ **228 files, exit 0** |

Rows A and E are the before/after on an identical tree with the one-line change as the only difference. Row D confirms the root config alone resolves fine with no `node_modules` at all — i.e. self-reference genuinely works, so the root config needs no change.

Also re-run in the worktree: `biome check .` → 228 files, exit 0; `tests/cli/generators/linting.test.ts` + `tests/cli/commands/doctor.test.ts` → **142 tests pass**. No test asserts on `apps/docs/biome.json`; `src/cli/generators/linting.ts` and doctor still emit and expect `@rtorcato/repo-tooling/biome` for consumers, unchanged.

### Bearing on the two reviewer notes

Left alone as instructed, but one now has a data point: the question about **`root: false` under a consumer whose monorepo has its own root config above it** is the same resolution/discovery machinery this fix exercised. Control D shows Biome is content with a root config that has no parent root; controls A/B show a *nested* config is loaded as soon as the scope reaches it, and that its `extends` resolves relative to **its own** directory, not the run root. That is the mechanism a maintainer would want to reason about for the nested-consumer case, but it does not answer it — still a maintainer call. The second note (copied assets reporting as drift) is untouched by this fix.


rtorcato (commented):

Checked every "shipped" file the format pass touches for a value change hidden in the reformat: `tooling/changesets/config.json`, `tooling/nx/nx.json`, `tooling/oxlint/oxlintrc.json`, both `tooling/release-please/*.json` files, and `tooling/rolldown/rolldown.config.mjs` are all pure tab→2-space (or quote-style) reformats with every key/value unchanged. `tooling/commitlint/commitlint.mjs` is a quote-style + trailing-semicolon change only, no rule values touched. `tooling/biome/biome.json` gets the same reindent plus the one intentional new key, `\"root\": false\" — no other field moved or changed. No supply-chain tampering hidden in the reformat.

The lint script change (`biome lint --config-path=biome.json src scripts` → `biome lint .`) is a strict widening — whole repo instead of two directories — so it cannot check fewer files for `src/` than before; no gate weakening. `favicon.svg` only gained a plain `<title>repo-tooling</title>` text node, no embedded script or foreign content. The `root: false` change on the shipped preset is a real behavior change worth a human's attention (see below), but nothing here is exploitable and no secrets or unsafe shell construction were introduced.

### Before merging
Nothing.

rtorcato (commented):

The two mechanical checks confirmed clean: `String(stderr)` and `stderr + ''` are equivalent for every type in play here (string, Buffer, null all coerce through the same primitive/`toString()` path), and the `AGENTS.md` line now matches the actual `check` script. Dropping `--config-path=biome.json` is plausible since Biome discovers `./biome.json` from cwd anyway, but nothing in this diff touches `.husky/pre-commit` or the CI lint job, so that can't be fully confirmed from the diff alone.

`root: false` was verified only against a bare consumer root with no parent config above it. It was not tested against a consumer nested inside a *larger* monorepo that has its own real root `biome.json` further up the tree — that's the case where `root: false` lets Biome walk past the copied preset to an outer root and change resolution behavior, the opposite of what `root: true` used to guarantee. That's a real gap in the verification claim, not a proven bug.

The style pass also reformats several *shipped* files (`tooling/changesets/config.json`, `tooling/nx/nx.json`, `tooling/oxlint/oxlintrc.json`, `tooling/release-please/*`, `commitlint.mjs`, `rolldown.config.mjs`). If the copied-asset drift check in `doctor` compares content/hash rather than semantics, every consumer who already copied these will show as stale/drifted on their next `doctor` run after upgrading — a real, if purely cosmetic, consequence of bundling `style:` reformatting into a `fix:` release that the PR doesn't address.

### Before merging
Confirm (a) `root: false` is safe for a consumer nested inside a larger monorepo with its own root config above, and (b) whether reformatting these shipped config files will surface as drift/staleness for existing consumers on their next `doctor` run.

rtorcato:

**Merge it, after weighing two open questions.** Both reviewers re-passed against the fix, `CLEAN`, all checks green.

The `lint` failure is resolved. `security-expert` confirmed `../../tooling/biome/biome.json` resolves inside the repo with no module resolution at all, and that `apps/docs` is still genuinely linted rather than silently resolving to nothing.

Two questions `code-reviewer` leaves for you — https://github.com/rtorcato/repo-tooling/pull/469#pullrequestreview-4994421398

- `root: false` is untested against a consumer nested inside a larger monorepo that has its own root config above it.
- Reformatting shipped config files may make existing consumers' copied assets report as drift on their next `doctor` run.

The second is the one with reach: it affects every repo that already ran `copy`, not just this one. `fix(biome):` cuts a patch.

rtorcato (commented):

The diagnosis holds up and the fix is real, not just CI-green: the reproduction exports only git-tracked files into a clean tree with zero `node_modules` anywhere, matching CI's actual restore shape, and control D (deleting `apps/docs/biome.json`) isolates the file as the sole cause while control B (restoring `--config-path`) rules out the obvious-looking suspect. That's causal evidence, not incidental passing.

The relative-path `extends` in `apps/docs/biome.json` is correctly scoped to this repo — `apps/docs` is a workspace member, not an external consumer, and nothing in `src/cli/generators/linting.ts` or its tests changed, so generated output for real consumers is untouched (still `@rtorcato/repo-tooling/biome`). Keeping the root config's package-specifier `extends` asymmetric from `apps/docs`'s relative one is justified: the root resolves via Node self-reference against `@rtorcato/repo-tooling`'s own `exports`, which is exactly the consumer-facing shape being dogfooded, while `apps/docs`'s nearest `package.json` is `@rtorcato/docs` so self-reference never applied there regardless of this PR.

The rest of the diff — `root: false` on the preset, the widened scripts, the 34-file mechanical format pass, the `favicon.svg` `<title>`, and the `useTemplate` fix — is consistent with #467 and free of logic changes; the reformatted shipped presets (`nx.json`, `oxlintrc.json`, etc.) are whitespace/quote-only.

### Before merging
`root: false` remains untested against a consumer nested inside a larger monorepo with its own root config above it. Reformatting shipped config files (indent/quote style) may make existing consumers' copied assets report as drift on their next `doctor` run.

rtorcato (commented):

The fix commit (`b8bf1ef`) is exactly the one line claimed: `apps/docs/biome.json` `extends` changes from the package specifier `@rtorcato/repo-tooling/biome` to the relative path `../../tooling/biome/biome.json`. From `apps/docs/`, that resolves to the repo-root `tooling/biome/biome.json` — stays inside the repo, no escape, no ambiguity, and needs no module resolution (which is precisely what was broken in CI, since `apps/docs/node_modules` doesn't exist there). The root `biome.json` is untouched and still extends via the package specifier, which is the correct shape since the root does resolve via self-reference. `files.includes` in `apps/docs/biome.json` is reformatted only (still excludes `.docusaurus`, `build`, `node_modules`, `docs/changelog.md`) — no value drift.

The lint-scope widening in `package.json` (`src scripts` → `.` for lint/format/check/check:fix) is unchanged by this fix commit and still strictly increases coverage. `tooling/biome/biome.json`'s formatting diff is still tabs→spaces only, `root: false` unchanged from the prior round. My previous pass on the 34-file format sweep and shipped-preset reformats (changesets, nx, oxlint, release-please, commitlint, rolldown) stands — nothing in that set changed in this diff.

No new files, scripts, or resolution paths were introduced beyond the one line described.

### Before merging
Nothing.

PR #470, opened 2026-08-21
feat(setup): review --preset in a terminal and add a minimal preset

### LINKED ISSUE #461: setup --preset is all-or-nothing and silently skips the wizard on a fresh project

rtorcato:
🤖 *Automated — implementer via ai-issue-loop. Written by an AI agent and posted under the owner's account; not a human message.*

Closes #461

## What changed

**1. TTY-aware `--preset`.** When `process.stdin.isTTY` and `process.stdout.isTTY` are both true and `--yes` is absent, `setup --preset` now prints the file list `computeFileList` resolved and opens a `checkbox` prompt (a type in the installed inquirer v14 registry — not `list`, per #463) letting the user drop git hooks, commitlint, semantic-release, Dependabot/CodeQL, badges and the AI agent files before anything is written.

Piped or redirected either direction is treated as CI and keeps today's exact one-shot behaviour. `--dry-run` and the new `-y, --yes` also bypass the review.

**2. A `minimal` preset.** tsconfig + Biome + Vitest, nothing else — built without spreading `BASE`, the way `swift-library` already is.

It uses `projectType: 'node-api'` rather than `'library'`. `library` is the only project type that adds surface of its own: `publint` / `attw` / `size-limit` scripts, the `@arethetypeswrong/cli` + `size-limit` deps, `.size-limit.json`, and a `dist/`-rooted `exports` map that `bundler: 'none'` never builds. `node-api` adds nothing, so it is the honest "plain Node package" shape. Flagging the naming in case you would rather I add a neutral project type instead.

**3. One root-cause fix picked up on the way.** `generateConfigs` only called `generateGitConfigs` when `gitHooks || commitLint`, but that function also writes `.gitignore`, which has nothing to do with either. So *every* `minimal` run — and every deselection of git hooks in the new prompt — would have scaffolded a repo with no `.gitignore`. The call is now unconditional (it already gates husky and commitlint internally). This also fixes the pre-existing case where the wizard's "no to hooks, no to commitlint" answers produced the same hole. `.gitignore` moved to the unconditional baseline in `computeFileList` to match.

## For your decision

**Is proposal 1 a breaking change — `feat:` or `feat!:`?**

I committed it as `feat:` and did not decide this. The argument either way:

- **Not breaking:** the non-TTY path is byte-identical, and non-TTY is the documented reason `--preset` exists. Every scripted/CI invocation is unaffected.
- **Breaking:** an interactive `--preset` run without a flag now stops and asks, where before it wrote and exited. That is a behaviour change to a documented path, and anyone running `--preset` by hand in a terminal (or from a tool that keeps a TTY attached) will now block on a prompt.

Retitle to `feat!:` before merging if you read it the second way. Note this decides whether a release goes out.

## Tests

`tests/cli/commands/setup.test.ts`, +224 lines:

- `minimal` flags, its file list, and its scaffold on disk — including that `package.json` has no `publint`/`attw`/`size-limit`/`release`/`prepare` scripts, no `exports`, and none of husky / lint-staged / commitlint / semantic-release / size-limit / attw in `devDependencies`.
- Non-TTY `--preset` calls `inquirer.prompt` **zero** times and still writes `.husky/pre-commit`, `AGENTS.md` and `.github/dependabot.yml` — this fails if the non-TTY path ever starts prompting.
- The TTY path asserts on the *questions handed to inquirer*: exactly one question, `type: 'checkbox'`, the type is present in the installed `inquirer.createPromptModule().prompts` registry, and the six choices are all there and pre-checked.
- Unchecking features writes none of their files and records the `false` in `.repo-tooling.json`.
- The file list is printed before the prompt.
- `--yes` and `--dry-run` each skip the review in a faked TTY.

TTY is faked by `Object.defineProperty` on the real `process.stdin`/`process.stdout`, restored in `finally`, so the tests exercise the real detection rather than a seam.

## Verification

- `biome lint --config-path=biome.json src scripts` — clean
- `biome check` on every file touched — clean (`check --write` was **not** run on `tests/`; that is #467)
- `vitest run` — 1005 passed, 18 skipped, 61 files

## Left alone deliberately

- The scaffold's own output failing `biome check` — #467.
- The `biome.json` `$schema` version pin vs the installed Biome — #468.
- `minimal` is **not** wired into `scripts/integration/preset-lifecycle.mjs`, so the CI preset loop does not run a real install/build/verify against it. Its config is a strict subset of presets that are covered, and the unit tests assert the emitted `package.json` and file set, but say the word and I will add it.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -14,7 +14,7 @@ Every command supports `--json` and a non-interactive mode. Combine with `--yes`
 
 | Command | Non-interactive | JSON output | Use case |
 |---|---|---|---|
-| `setup --preset <name>` | ✅ | `--dry-run` only | Scaffold a new project. Presets: `library`, `web-app`, `node-api`, `nextjs-app`, `react-app`, `swift-library`. |
+| `setup --preset <name>` | ✅ (non-TTY, or with `--yes`) | `--dry-run` only | Scaffold a new project. Presets: `minimal`, `library`, `web-app`, `node-api`, `nextjs-app`, `react-app`, `swift-library`. In an interactive terminal it first prints the file list and lets you deselect; pass `--yes` for the one-shot behaviour. |
 | `setup --config <path>` | ✅ | `--dry-run` only | Scaffold with a full `ProjectConfig` JSON file. See `setup --config-schema`. |
 | `setup --config-schema` | ✅ | ✅ (JSON Schema) | Print the JSON Schema for `ProjectConfig`. Use to validate configs before scaffolding. |
 | `setup --dry-run` | ✅ | ✅ | Print resolved config + file list without writing. Pair with `--preset` or `--config`. |
--- a/README.md
+++ b/README.md
@@ -35,9 +35,15 @@ npx @rtorcato/repo-tooling setup
 Non-interactive — scaffold from a named preset in one shot (CI-friendly):
 
 ```bash
-npx @rtorcato/repo-tooling setup --preset library -d ./my-lib --skip-install
-# presets: library | web-app | node-api | nextjs-app | react-app | swift-library
+npx @rtorcato/repo-tooling setup --preset library -d ./my-lib --skip-install --yes
+# presets: minimal | library | web-app | node-api | nextjs-app | react-app | swift-library
 ```
+
+Run without `--yes` in a terminal and the preset first prints every file it would
+write and lets you uncheck the parts you don't want (git hooks, commitlint,
+releases, Dependabot/CodeQL, badges, AI agent rules). Piped or redirected — CI —
+it writes everything in one shot, exactly as before. `minimal` is tsconfig +
+Biome + Vitest and nothing else.
 
 `swift-library` scaffolds a SwiftPM package (manifest, sources, tests, SwiftLint,
 Periphery, macOS CI) instead of an npm one — see the
--- a/apps/docs/docs/guides/cli.md
+++ b/apps/docs/docs/guides/cli.md
@@ -13,10 +13,16 @@ npx @rtorcato/repo-tooling setup -d ./my-app  # specific directory
 npx @rtorcato/repo-tooling setup --skip-install  # skip npm/pnpm install
 ```
 
-Add `--preset <name>` to skip the prompts entirely: `library`, `web-app`,
-`node-api`, `nextjs-app`, `react-app`, or `swift-library`. The Swift preset
-scaffolds a SwiftPM package rather than an npm one — see the
-[Swift guide](./swift.md).
+Add `--preset <name>` to skip the wizard: `minimal`, `library`, `web-app`,
+`node-api`, `nextjs-app`, `react-app`, or `swift-library`. `minimal` is tsconfig
++ Biome + Vitest and nothing else. The Swift preset scaffolds a SwiftPM package
+rather than an npm one — see the [Swift guide](./swift.md).
+
+In a terminal, `--preset` prints the file list it resolved and lets you uncheck
+the extras (git hooks, commitlint, releases, Dependabot/CodeQL, badges, AI agent
+rules) before anything is written. Pass `--yes` to write it all in one shot. When
+stdin or stdout is piped or redirected — CI — the review is skipped
+automatically.
 
 ## copy \<name\>
 
--- a/apps/docs/docs/guides/for-ai-agents.md
+++ b/apps/docs/docs/guides/for-ai-agents.md
@@ -14,10 +14,10 @@ All commands take `-d <path>` for the target directory (defaults to `process.cwd
 ### `setup` — scaffold a new project
 
 ```bash
-# Sane defaults per project type (no prompts)
-npx @rtorcato/repo-tooling setup --preset library -d ./my-lib --skip-install
-npx @rtorcato/repo-tooling setup --preset react-app -d ./my-app --skip-install
-npx @rtorcato/repo-tooling setup --preset nextjs-app -d ./my-site --skip-install
+# Sane defaults per project type. --yes guarantees no prompts even from a terminal.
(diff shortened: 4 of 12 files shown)


rtorcato (commented):

No injection surface here: the project name only ever passes through `path.basename(path.resolve(...))` before landing in filenames/content, there's no `child_process` or template-string shell construction anywhere in this diff, and no dependency is added (inquirer/chalk are already installed). The new `--yes`/TTY-review logic is a pure config-object mutation feeding the existing writers, so nothing new can be interpolated unsafely.

On consistency: the `.gitignore` fix is the one genuinely security-relevant change — it moves `.gitignore` off the `gitHooks` gate so `minimal` (and any deselection of git hooks) no longer scaffolds a repo that would commit `node_modules`/secrets by default. `minimal` dropping `securityAutomation` (Dependabot/CodeQL) is called out explicitly in the PR body, README, docs, and SKILL.md, so it's not a silent baseline loss. I didn't find a new case where partial deselection produces a broken reference (e.g. a badge pointing at a workflow that was skipped) introduced by this diff specifically — that combination, if it exists, predates this PR via the classic wizard path.

Reviewable-features checkbox is opt-out (all pre-checked) and TTY-gated correctly (piped stdin/stdout keeps the exact one-shot CI behavior, verified by the added tests). No secrets, no unsafe shell, no supply-chain changes.

### Before merging

Nothing.

rtorcato (commented):

Verified the three focus areas from diff. (1) Non-TTY path: `resolveConfig` returns `buildPresetConfig(...)` directly when `options.dryRun || options.yes || !isInteractiveTerminal()`, before any prompt — the "does not prompt when stdin/stdout are not a terminal" test spies on `inquirer.prompt` and asserts zero calls while confirming `.husky/pre-commit`, `AGENTS.md`, and `.github/dependabot.yml` still land. (2) `minimal` in `setup-presets.ts` is a standalone literal object, not a spread of `BASE`, and its own test asserts none of husky/commitlint/semantic-release/size-limit/attw show up in files or `devDependencies`. (3) The new prompt uses `type: 'checkbox'`, and the test asserts that type against `inquirer.createPromptModule().prompts` (the actual v14 registry), not just a mock call count — so it would have caught the `list` regression from #463.

The `.gitignore`-unconditional fix in `generators/index.ts` is a real, correctly-scoped bug fix: it only changes output for configs where both `gitHooks` and `commitLint` are false (new `minimal` case, or the wizard's "no/no" path), and doesn't touch any existing full-preset's non-TTY output.

### Before merging
The `## For your decision` semver question (is the interactive prompt a `feat:` or `feat!:`) is a real call for a human — non-TTY/CI behavior is unchanged, but a human running `--preset` at a terminal without `--yes` now blocks on a prompt instead of writing immediately, which is arguably a breaking behavior change to a documented path.

rtorcato:

**Merge it, after answering one question.** Both reviewers passed, `CLEAN`, all checks green.

`code-reviewer` left the semver call for you: non-TTY and CI behaviour is unchanged, but an interactive `--preset` without `--yes` now blocks on a prompt instead of writing immediately. That decides `feat:` vs `feat!:`, and the squash subject decides the release. Retitle before merging if you want the major — https://github.com/rtorcato/repo-tooling/pull/470#pullrequestreview-4994347945

No unattended merge here because this repo has no `release` environment with `required_reviewers` (#429).

PR #471, opened 2026-08-21
feat(security): auto-merge github-actions patch and minor bumps

### LINKED ISSUE #452: Dependabot automerge: github-actions bumps lost auto-merge, and was fix: the right semver?

Follow-up to #423 / PR #447 (`bb32f69`), which narrowed the scaffolded `dependabot-automerge.yml` to auto-merge only the `dev-minor` group. `code-reviewer` raised two questions on that PR that are human calls; it merged before they were answered, so filing them rather than letting them evaporate.

## 1. `github-actions` bumps lost auto-merge as a side effect

The goal was to stop auto-merging **production dependencies** — packages that ship to consumers of the published package. But the gate is now `dependency-group == 'dev-minor'`, and `github-actions` updates are not in that group, so they no longer auto-merge either.

A CI action bump is not a runtime dependency and does not reach any consumer of the published package, so this is wider than the stated goal. The practical cost is that routine `actions/checkout` and `actions/setup-node` bumps now wait for a human across every scaffolded repo — which is the kind of friction that ends with people rubber-stamping without reading, weakening the gate that was just added.

Options:

| Option | Notes |
|---|---|
| Add `github-actions` to the auto-merge condition | Restores the previous convenience for CI actions while keeping production deps gated. Matches the issue's stated intent most closely. |
| Leave as-is | Defensible on the grounds that a compromised action runs in CI with repo credentials, which is its own supply-chain surface — arguably *more* sensitive than a dev dependency, not less. |
| Give actions their own group and decide separately | Most explicit; more config surface. |

Genuinely arguable in both directions — the second column is not a strawman.

## 2. `fix:` may be the wrong semver

PR #447 landed as `fix(security):`, so it cut a **patch**. But it removes behaviour consumers currently rely on: a repo that takes the update stops auto-merging production bumps it merges today.

Whether that is a patch (a security fix, and arguably the old behaviour was the bug) or a breaking change (`feat!:` — scaffolded output changes in a way that alters what merges without a human) is a judgement call. It is already released, so this is not "fix the tag" — it is about whether the next similar change should be typed differently, and whether anything needs saying in the release notes for the version that shipped it.

## Not a defect in #447

Both are consequences of the option pinned before implementation ("dev-only", chosen because it is the only one that fails closed), not mistakes in the code. The implementation itself was reviewed clean by both agents: the group pairing matches what the config generator emits, the gate fails closed on an empty or unexpected `dependency-group`, and the majors exclusion survived.

## Related

- #423 — the parent issue
- PR #447 — `bb32f69`, where both questions were raised as `ai-notes`

rtorcato:
🤖 *Written by Claude (AI agent) under @rtorcato's account, from the repo-tooling `ai-issue-loop` session. Not a human message — there is no separate GitHub account for AI agents.*

Implements the owner's decision recorded on #452 (option 1): restore auto-merge for `github-actions` Dependabot bumps.

## What changed

`DEPENDABOT_AUTOMERGE_WORKFLOW` now matches CI action bumps on `package-ecosystem` as an **alternative to the `dev-minor` group**, inside the same parenthesised alternation:

```yaml
if: |
  steps.gate.outputs.safe == 'true' &&
  (steps.metadata.outputs.dependency-group == 'dev-minor' ||
  steps.metadata.outputs.package-ecosystem == 'github-actions') &&
  (steps.metadata.outputs.update-type == 'version-update:semver-patch' ||
  steps.metadata.outputs.update-type == 'version-update:semver-minor')
```

Action bumps are ungrouped, so they report an empty `dependency-group` — matching on the ecosystem is what lets them through without loosening the group gate for anything else.

Every other exclusion survives:

- **Majors still wait for a human.** The `update-type` clause is a separate `&&` term, so it applies to the ecosystem arm too. `actions/checkout@v7 → v8` does not auto-merge.
- **#462's manifest gate still runs and composes correctly.** I ran it rather than assumed it: an action bump reports names like `actions/checkout`, which appear in no `package.json`, so `ships` never matches and it resolves `safe=true`. The empty-`NAMES` fail-closed branch is untouched and still refuses a PR that reports nothing.
- **`production-minor` still excluded** — asserted by `expect(content).not.toMatch(/'production-minor'/)`.

Also:

- **`doctor`** (`checkDependabot`) now asserts the `github-actions` arm as well, so a repo that took the #423/#458 pair is told it is behind. Both existing greps (`dependency-group == 'dev-minor'`, `steps.gate.outputs.safe == 'true'`) survive the change verbatim — re-read, not assumed.
- **Dogfooded.** `.github/workflows/dependabot-automerge.yml` was regenerated from the built generator (not hand-edited), so it is byte-identical to what `fix dependabot` emits. `pnpm dogfood` is clean: 60 checks, 6 accepted.
- **`.github/dependabot.yml`** also regenerated. It carried a stale pre-#458 comment ("the only tier that auto-merges on green") that this change makes doubly wrong; the diff there is comment-only, no config change.
- **Docs** — the auto-merge table, the "excluded for the same reason as production bumps" statement (now false), and the recorded counter-argument.

## Tests

`tests/cli/generators/security.test.ts` now asserts the merge condition **whole** rather than clause by clause — a partial match would pass on an expression that quietly dropped one exclusion. Plus a test that the ecosystem arm sits inside the group alternation (majors excluded), and a gate test that a CI action bump resolves `safe=true`.

Mutation-checked both required properties before pushing:

| Mutation | Result |
|---|---|
| Remove the `update-type` (majors) clause | 3 tests fail |
| Remove `steps.gate.outputs.safe == 'true' &&` | 3 tests fail |

`biome lint`, `tsc --noEmit`, `vitest run` (1017 tests) all green.

## Semver

Typed **`feat(security):`**, deliberately — a minor, not the `fix` that #447 used.

The owner's recorded reasoning for `fix` on #447 was that auto-merging production deps was the bug, so narrowing it is a fix. That argument does not transfer: this goes the other direction. It **widens** what merges unattended in every repo that runs `fix dependabot`, and a patch release is too quiet a signal for "more things now merge without a human". I recorded the symmetry in the guide: narrowing ships as `fix`, widening as `feat`.

## For your decision

Two things I did not decide silently:

1. **`feat` vs `fix` is genuinely arguable.** You could read this as fixing #447's overreach — the issue itself calls the actions exclusion "wider than the stated goal", i.e. a defect in the old behaviour, which is exactly the `fix` argument. If you prefer `fix(security):`, say so and I will amend before this merges; nothing else changes.

2. **`cooldown` does not actually cover CI actions.** The decision comment lists the 7-day `cooldown` as one of the three mitigations, but `cooldown` is declared **per `updates:` entry** and the canonical `github-actions` entry has none — only the npm entry does. So an action bump can open the day it is published, and the real mitigations are the majors exclusion and required status checks. I documented that gap honestly rather than repeating the claim, and left the config alone because adding `cooldown` to the `github-actions` entry is outside the agreed scope. It is a three-line change to `dependabotConfig()` if you want it; worth a follow-up issue either way.

Closes #452

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -25,8 +25,11 @@ updates:
         update-types:
           - minor
           - patch
-      # The only tier that auto-merges on green — dev tooling never reaches a
-      # consumer of the published package.
+      # The tier that can auto-merge on green. "development" is Dependabot's
+      # classification, not a safety property: a package listed in both
+      # devDependencies and peerDependencies lands here, and peer deps are part
+      # of a published package's contract (#458). The auto-merge workflow
+      # re-checks every name against the manifests rather than trusting this.
       dev-minor:
         dependency-type: development
         update-types:
--- a/.github/workflows/dependabot-automerge.yml
+++ b/.github/workflows/dependabot-automerge.yml
@@ -58,10 +58,17 @@ jobs:
       # Belt and braces: the dev-minor group is already declared minor+patch in
       # dependabot.yml, but this workflow is the security gate and shouldn't
       # trust a config file a consumer repo can edit independently.
-      - name: Auto-merge dev-dependency patch and minor updates
+      #
+      # github-actions bumps are ungrouped, so they report an empty
+      # dependency-group and are matched by ecosystem instead (#452). They reach
+      # no consumer of the published package, and their squash subject is
+      # "ci(deps): …", which cuts no release. Majors still fall through to a
+      # human — the update-type clause below applies to them too.
+      - name: Auto-merge dev-dependency and CI action patch and minor updates
         if: |
           steps.gate.outputs.safe == 'true' &&
-          steps.metadata.outputs.dependency-group == 'dev-minor' &&
+          (steps.metadata.outputs.dependency-group == 'dev-minor' ||
+          steps.metadata.outputs.package-ecosystem == 'github-actions') &&
           (steps.metadata.outputs.update-type == 'version-update:semver-patch' ||
           steps.metadata.outputs.update-type == 'version-update:semver-minor')
         run: gh pr merge --auto --squash "$PR_URL"
--- a/apps/docs/docs/guides/dependabot-strategy.md
+++ b/apps/docs/docs/guides/dependabot-strategy.md
@@ -28,18 +28,20 @@ Dependabot opens roughly **3–4 PRs per cycle** instead of one per package:
 | `production-minor` | runtime `dependencies` | minor, patch | ❌ manual |
 | `dev-minor` | `devDependencies` | minor, patch | ✅ on green, if nothing in it ships |
 | `major-updates` | all packages | major | ❌ manual |
-| `github-actions` | workflow actions | all | ❌ manual |
+| `github-actions` | workflow actions | minor, patch | ✅ on green |
+| `github-actions` | workflow actions | major | ❌ manual |
 
-## 2. Auto-merge — dev dependencies only
+## 2. Auto-merge — dev dependencies and CI actions
 
-Patch and minor updates **of `devDependencies`** merge themselves once CI is
-green — no human in the loop. Implemented by
+Patch and minor updates **of `devDependencies`** and **of GitHub Actions** merge
+themselves once CI is green — no human in the loop. Implemented by
 `.github/workflows/dependabot-automerge.yml` using `dependabot/fetch-metadata` +
-`gh pr merge --auto --squash`, gated on **both** `dependency-group == 'dev-minor'`
-and `version-update:semver-patch` / `version-update:semver-minor`.
+`gh pr merge --auto --squash`, gated on the consumer-facing check below, **either**
+`dependency-group == 'dev-minor'` **or** `package-ecosystem == 'github-actions'`,
+**and** `version-update:semver-patch` / `version-update:semver-minor`.
 
-Everything else waits for a human, and the gate **fails closed**: a PR outside a
-group reports an empty `dependency-group` and so never matches.
+Everything else waits for a human, and the gate **fails closed**: an npm PR
+outside a group reports an empty `dependency-group` and so never matches.
 
 > **The group name is not the gate.** `dependency-type: development` is
 > Dependabot's classification, and it files a package listed in *both*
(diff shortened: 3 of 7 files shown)


rtorcato (commented):

The gate composes correctly: `safe == 'true' && (dev-minor || github-actions) && (patch || minor)`. The `update-type` term is a separate `&&` clause outside the OR group, so it applies to both arms uniformly — an `actions/checkout` major still falls through to a human, and no `||` placement widens past that. The manifest gate from #462 is untouched and verified rather than assumed: a new test runs the actual gate script with action names (`actions/checkout,actions/setup-node`) and confirms `safe=true`, while the empty-names fail-closed branch is unaffected. `production-minor` exclusion is still asserted unchanged.

The generator (`src/cli/generators/security.ts`) and dogfooded workflow (`.github/workflows/dependabot-automerge.yml`) hunks are byte-identical in the diff. `checkDependabot` in `src/base/checks.ts` gains a new drift assertion for the `github-actions` arm, and the diff context shows the prior `dev-minor`/`safe` checks are untouched (no hunk touches those lines). Tests assert the whole `if:` expression as one string rather than clause-by-clause, so removing either the majors exclusion or the `safe` gate breaks the match — that's a real mutation-resistant test, not just a substring grep.

No defects found. This is a well-scoped, well-tested restoration of previously-authorized behavior.

### Before merging
Nothing.

rtorcato (commented):

Confirmed the widened arm is exactly what #452 authorized and no wider. The new condition adds `steps.metadata.outputs.package-ecosystem == 'github-actions'` strictly as an alternative *inside* the same parenthesised group-match term — `(dev-minor || github-actions) && (patch || minor)` — so the `update-type` clause still applies to the new arm with no operator-precedence escape hatch; a major action bump cannot short-circuit into the auto-merge path. The gate script itself (`steps.gate.outputs.safe`, the manifest/`NAMES` fail-closed check from #462) is untouched by this diff, and the new test confirms an action name like `actions/checkout` matches no `package.json` entry so it resolves `safe=true` without weakening the empty-`NAMES` fail-closed branch. No new permissions, secrets, or PR-content execution were introduced — the diff only extends an existing `if:` condition, a doc/comment set, and a `doctor` drift check.

The threat model is stated honestly: the docs explicitly flag that a compromised action runs in CI holding `GITHUB_TOKEN`, name the majors-exclusion and required-status-checks as the real mitigations, and correctly note that `cooldown` does **not** apply to this arm since it's declared per `updates:` entry and the `github-actions` entry has none — this corrects rather than repeats the prior (now-inaccurate) claim that actions were "excluded for the same reason" as production bumps.

### Before merging
Nothing.

Issue #473, opened 2026-08-21
Pass 4 never clears ai-ready when claiming an issue, so merged work can be re-implemented


rtorcato:
🤖 *Filed by Claude (AI agent) on rtorcato's behalf. Posted under the owner's account; there is no separate GitHub account for AI agents.*

`skills/ai-issue-loop/SKILL.md` ships to npm, so this is published behaviour.

## The bug

Pass 4 claims an issue by **adding** `ai-wip` and never removing `ai-ready`:

```bash
gh issue edit <N> --add-label ai-wip
```

So a claimed issue carries both. That is harmless while `ai-wip` is present — Pass 4's eligibility query filters on it:

```
| select([.labels[].name] | index("ai-wip") == null)
```

It stops being harmless the moment `ai-wip` clears for any reason other than the PR merging and closing the issue:

- **Pass 2's benign-stall path** clears `ai-wip` *without* `ai-blocked` by design, so the issue can be picked straight back up. With `ai-ready` still attached, it is — correctly.
- **An implementer that hands back** runs `--add-label ai-blocked --remove-label ai-wip`, and `ai-blocked` filters it out. Fine.
- **A PR that uses `Refs #N` rather than `Closes #N`**, or a partial implementation, leaves the issue open with a PR already up. `ai-wip` gets cleared, `ai-ready` remains, and the next tick spawns an agent to re-implement work already sitting in an open PR.

The last case is not hypothetical. It came up four times in one session:

- **#458** — PR #462 was open and closing it; the issue still carried `ai-ready`. Caught by hand, or the next tick would have re-implemented merged work.
- **#467, #461, #452** — all three sat as `["ai-ready", "ai-wip"]` with PRs open, cleared manually.

#429 hit the same shape earlier via `Refs` instead of `Closes` and had to be moved to `holding`.

## Why the labels should be exclusive

The skill's own state diagram says so:

```
issue: ai-ready ─pickup─> ai-wip ─> PR opened, labelled ai-review
```

That is a transition, not an accumulation. The Labels table defines `ai-ready` as *"Eligible for an agent"* and `ai-wip` as *"Claimed"* — an issue cannot honestly be both, and the list view showing both makes the queue unreadable at a glance.

## Fix

Pass 4's claim becomes:

```bash
gh issue edit <N> --add-label ai-wip --remove-label ai-ready
```

Then anything that legitimately returns an issue to the queue has to re-add `ai-ready` explicitly, which is the correct and visible behaviour:

[...]

PR #474, opened 2026-08-21
fix(skills): cap ai-issue-loop agent comment length

### LINKED ISSUE #472: ai-issue-loop leaves 40-line comments on every PR, including clean ones

`skills/ai-issue-loop/SKILL.md` ships to npm, so this is published behaviour, not a local wart.

## Measured

Every PR the loop touches collects ~6KB of agent prose across three comments, most of which changes nothing a reader would do.

| Comment | Size |
|---|---|
| Pass 1 decision comments | 2360–4033 chars, 30–45 lines |
| Reviewer comments | 1652–4303 chars |

The clearest case: **#466's decision comment ran 2501 chars to say "both reviewers found nothing, merge it."** #464's was 2394 for the same verdict.

The skill already makes this exact argument about `ai-notes` — *"on every PR is the failure mode — it trains the reader to ignore it, which is worse than not having it"* — and then produces a 40-line comment on every PR regardless of outcome.

## Three causes

**1. One comment shape is applied to every outcome.** Pass 4's decline shape (`## To lift this hold`, an options table, a `<details>` block) was designed for *"no agent should ever start this issue"* — a genuinely hard handoff. Pass 1 then says the ready-handoff *"gets the decision comment"* too, so a PR where both reviewers returned `Nothing.` receives the full ceremony.

**2. The skill already bans restating, and it is ignored.** Pass 1 says:

> Pointing at the notes beats restating them — the reviewer already wrote a `### Before merging` section, so link its comment rather than paraphrasing it into drift.

In practice every decision comment restated the findings, the package tables and the check lists. #455's listed the same 13 peer dependencies for the third time on that PR.

**3. Reviewer bodies have no length rule at all.** The prompt ends *"Say nothing else and return a one-line summary"*, but that governs the **return value**, not the comment body. So reviewers narrate their whole process above `### Before merging`, which is the only part read at merge time.

## Proposed rules

| Outcome | Comment |
|---|---|
| Clean + ready | **None.** `ai-ok-code, ai-ok-sec` + assigned + no `ai-review` already says it. |
| `ai-notes` | ≤10 lines. Link the reviewer's `### Before merging`; never restate it. |
| `ai-changes` / red / blocked | ≤10 lines, leading with the action, then the specific cause. |
| Reviewer body | `### Before merging` plus **at most 3 short paragraphs** above it. |

The first row is the significant one and it is pure subtraction. Pass 1 justifies the decision comment as *"a label records that something is wrong, never what"* — which simply does not apply when nothing is wrong. Dropping it removes the majority of the volume without losing any signal.

Worth stating the reasoning in the skill itself, since it is the same argument it already makes for `ai-notes`: a 40-line comment on every PR trains the reader to skip all of them, including the one that matters.

## Already tested

Both halves were tried by hand before filing:

- Reviewers on #469 and #470 were briefed with *"at most 3 short paragraphs above the closing section"* and produced usable reviews.
- #469's send-back decision comment came to **901 chars** — down from the 2360–4033 range — while still naming the failing check, quoting the error, giving the cause, and pointing at the reviewer notes rather than repeating them.

So the rules are known to work; this issue is about writing them into the skill.

## Scope

`skills/ai-issue-loop/SKILL.md` only. Pass 1's decision-comment section, Pass 4's decline shape (which should say it is for declines specifically, not every handoff), and the two reviewer prompt templates.

The marker-upsert mechanism itself is fine and should not change.

## Related

- #473 — Pass 4 never clears `ai-ready` when claiming an issue. Same file; queue after this one.

rtorcato:
🤖 *Written by Claude (AI agent) via ai-issue-loop, under the owner's account. There is no separate GitHub account for AI agents, so this PR appears under @rtorcato's avatar; it is not a human-authored message.*

Closes #472

`SKILL.md`: **938 → 963 lines** (+31/-6). The rules are stated once and kept terse rather than repeated at each site.

## What changed

The budget is stated **once**, beside the `🤖` header rule that already governs every agent comment — a small table plus two sentences reusing the skill's own `ai-notes` argument (a comment on every PR trains the reader to skip all of them). The four sites then just point at it:

- **Pass 1, ready handoff** — posts **no comment at all**. Sharpened the existing "Take no other action" line: `ai-ok-code, ai-ok-sec` + assigned + no `ai-review` is the whole message. A comment records what a label cannot; a clean PR has nothing to record.
- **Pass 1, send-back** — the existing "comment why" now reads "≤10 lines, leading with what must change, then the failing check and its error."
- **Both reviewer prompts** — the cap is folded into the *existing* paragraph about `### Before merging` being the only part read at merge time, rather than added as a new rule beside it. The closing line now says explicitly that "say nothing else" governs the reply to the orchestrator, not the comment body.
- **Pass 4, decline** — marked the one exception to the budget, so its shape stops leaking onto PR handoffs.

## Scope correction — please read

**Three of the four things #472 describes are not in this file.** The `<!-- ai-issue-loop:decision -->` marker-upsert, the `## To lift this hold` + options-table + `<details>` shape, and the sentence *"Pointing at the notes beats restating them"* appear only in `~/dotfiles/dotfiles/claude/.claude/skills/ai-issue-loop/SKILL.md` — the local copy that actually drives the loop. No open PR adds them here. The two copies have diverged.

So this PR fixes the **shipped** skill: the reviewer-prompt caps (cause 3 in the issue, and the one that measurably produced the 1652–4303 char reviews) plus the clean-PR rule written in so the decision comment cannot be reintroduced unbudgeted. Fixing the *observed* Pass 1 comments needs the same edit in the dotfiles copy, which is outside this repo. Worth a follow-up issue on the divergence itself.

## Commit type

`fix` rather than `docs`: this file ships to npm and is installed as the skill consumers run, so the change has to reach them via a release.

## Verification

- `./node_modules/.bin/vitest run` — 996 passed, 18 skipped (62 files)
- `./node_modules/.bin/tsc --noEmit --project src/cli/tsconfig.json` — clean

No `biome --write` pass was run (#467 is in flight on the same kind of change); the diff is `skills/ai-issue-loop/SKILL.md` only.

### DIFF
--- a/skills/ai-issue-loop/SKILL.md
+++ b/skills/ai-issue-loop/SKILL.md
@@ -52,6 +52,20 @@ header names the agent *and* says why it is wearing a human's face:
 
 `🤖 *Automated — <which agent> via ai-issue-loop. Posted under the owner's account by an agent; not a human message. There is no separate GitHub account for AI agents, so this appears under @<owner>'s avatar.*`
 
+**Comment budget: ≤10 lines, and a clean outcome gets no comment at all.** A
+40-line comment on every PR trains the reader to skip all of them, including the
+one that matters — the same failure mode as `ai-notes` on every PR, below. Link
+the reviewer's `### Before merging` rather than restating it; a paraphrase is
+drift with a second copy to maintain.
+
+| Outcome | Comment |
+|---|---|
+| Clean and ready | **None.** `ai-ok-code, ai-ok-sec` + assigned + no `ai-review` already says it. |
+| `ai-notes` | ≤10 lines; link the reviewer's `### Before merging`. |
+| `ai-changes`, CI red, `ai-blocked` | ≤10 lines, action first, then the specific cause. |
+| Reviewer verdict | `### Before merging` plus at most 3 short paragraphs above it. |
+| Declining an issue | The one exception — a hard handoff needs its reasoning; see Pass 4. |
+
 ## Labels
 
 | Label | On | Meaning |
@@ -250,7 +264,10 @@ It lands in the user's *Assigned to you* view, and the labels then read as state
 than noise — `ai-ok-code, ai-ok-sec` with no `ai-review` means **waiting on you**. Both
 halves matter: Pass 3 only ever *adds* the `ai-ok-*` labels, so without the removal a
 finished PR keeps wearing `ai-review` forever and looks mid-review. Idempotent, so
-re-running a tick is harmless. Take no other action — do not merge.
+re-running a tick is harmless. Take no other action — do not merge, and **post no
+comment**: nothing is wrong, so those three labels are the whole message. A
+comment is how the loop records what a label cannot; a clean PR has nothing to
+record.
 
 **Never strip `ai-notes` here.** It is the whole point of the handoff: it has to
 survive to the moment of merging, which is the moment it is for. A ready PR reads
@@ -281,7 +298,8 @@ a required check, so nothing in the check list looked wrong either.
 
 Diff-scoped reviewers cannot catch this — they never see CI. So when a
 both-passed PR is not `CLEAN`, do not assign it as ready. Send it back, and
-**comment why**: the reviewers passed it, so the fix-round implementer would
+**comment why** — ≤10 lines, leading with what must change, then the failing
+check and its error. The reviewers passed it, so the fix-round implementer would
 otherwise read the comments and find no instruction to act on.
 
 ```bash
@@ -506,7 +524,10 @@ Reviewer prompt template:
 >
 > That section is what a human reads at merge time, so put anything you would
 > want them to know there rather than leaving it in the prose above — a finding
-> buried mid-paragraph does not survive the handoff. The bar is a finding that
+> buried mid-paragraph does not survive the handoff. For the same reason, **cap
+> the body at that section plus at most 3 short paragraphs above it**: no process
+> narration, no restating the diff, no listing what you checked and found fine.
+> The bar is a finding that
 > **changes what a human would do**: a semver implication, a deliberate
 > omission, a follow-up that must be filed. Not observations, not praise, not
 > restating the diff. Writing `Nothing.` is a real verdict and the common one —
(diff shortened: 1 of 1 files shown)


rtorcato (commented):

Pure comment-length policy change on `SKILL.md`; no code, no merge-gate logic touched. Checked the safety-relevant bits specifically: the `🤖 *Automated …*` header requirement is untouched, the new "clean gets no comment" rule stays scoped to the clean/ready case only — `ai-notes`, `ai-changes`, CI-red, and `ai-blocked` all still get a required ≤10-line comment stating the reason, per the new table. The `ai-notes`-suppresses-auto-merge rule, the release gate, and the Dependabot-only unattended-merge rule are all untouched. Reviewer prompts still require the verdict label and the closing `### Before merging` section — the 3-paragraph cap applies only to the prose above it, and the "say nothing else" line now explicitly scopes to the orchestrator reply, not the comment body, removing an ambiguity rather than adding one.

The PR body itself flags the honest limitation: three of #472's four fixes (marker-upsert shape, `## To lift this hold` table, "pointing beats restating") live only in the dotfiles copy of this skill and aren't touched here, with a stated follow-up needed. That's a scope note, not a defect in this diff.

### Before merging
Nothing.

rtorcato (commented):

Growth (+31/-6, 938→963) is mostly one legitimate addition: a single table+paragraph stating the budget once, which every other site then points at. The four in-place edits (Pass 1 ready/send-back, both reviewer templates, Pass 4 decline) mostly *extend* existing sentences rather than bolt on new paragraphs, and all five checklist items pass — clean-ready PRs explicitly get **no comment** (pure subtraction, the biggest win), Pass 4's ceremony is now marked as the one exception instead of leaking onto handoffs, both reviewer templates (general + Dependabot) get the 3-paragraph cap, and marker/header/label mechanics are untouched.

One instance of the exact failure mode the issue names: the new top paragraph adds "Link the reviewer's `### Before merging` rather than restating it; a paraphrase is drift with a second copy to maintain" as a second, differently-worded copy of the pre-existing "Pointing at the notes beats restating them" rule instead of sharpening that sentence in place. Minor, not functional, but it's the ironic case of stating "don't add a rule beside an existing one" while doing exactly that.

### Before merging
Consider replacing the new "Link the reviewer's ### Before merging rather than restating it" sentence with a pointer to the existing "Pointing at the notes beats restating them" line, per the issue's own instruction to sharpen rather than duplicate.

rtorcato:

**Merge it.** Both reviewers passed, `CLEAN`, all checks green.

`code-reviewer` left one nit worth knowing rather than acting on: the new "Link the reviewer's `### Before merging` rather than restating it" sentence is a second, differently-worded copy of the pre-existing "Pointing at the notes beats restating them" rule — i.e. the PR states *don't add a rule beside an existing one* while doing exactly that. Not functional. https://github.com/rtorcato/repo-tooling/pull/474#pullrequestreview-4994419764

**Cheapest fix is to let it ride.** #473 edits this same file next and is queued behind this PR; the dedup can go in there rather than costing a re-review cycle here.

No unattended merge — this repo has no `release` environment with `required_reviewers` (#429).

PR #476, opened 2026-08-21
fix(skills): clear ai-ready when ai-issue-loop claims an issue

### LINKED ISSUE #473: Pass 4 never clears ai-ready when claiming an issue, so merged work can be re-implemented

rtorcato:
🤖 *Written by Claude (AI agent) on rtorcato's behalf. Posted under the owner's account; there is no separate GitHub account for AI agents.*

Pass 4 claimed an issue with `gh issue edit <N> --add-label ai-wip` and never removed `ai-ready`, so a claimed issue carried both. That is harmless only while `ai-wip` is set, because Pass 4's eligibility query filters on it. The moment `ai-wip` clears for any reason other than the PR closing the issue — a `Refs #N` body, a partial implementation, a benign stall — the leftover `ai-ready` returns the issue to the queue and the next tick re-implements work already sitting in an open PR. Seen on #458, #467, #461 and #452 in one session; #429 hit the `Refs` shape earlier.

The claim now reads `--add-label ai-wip --remove-label ai-ready`, matching the skill's own state diagram (`ai-ready ─pickup─> ai-wip` is a transition, not an accumulation). The diagram was already right and needed no change.

## The audit

Making `ai-ready` exclusive makes it load-bearing, so every path that clears `ai-wip` now decides explicitly whether the issue returns to the queue. Getting that backwards — an issue silently leaving the queue forever — would be worse than the original bug, so each transition is now stated rather than implied:

| Transition | Decision |
|---|---|
| Pass 2 benign stall (`ai-wip` cleared, no `ai-blocked`) | **Re-queue.** Now `--add-label ai-ready --remove-label ai-wip`; the comment must say it was re-queued. Clearing `ai-wip` alone would drop it out silently. |
| Pass 2 merged-PR cleanup | **Do not re-queue.** Normally a no-op on an issue already closed by `Closes #N`. A `Refs #N` PR leaves it open — assign it (`--add-assignee @me`) so a landed issue is owned rather than unowned. This is the #429 case. |
| Pass 2 stall reaping | **Do not re-queue.** Confirmed as-is: `ai-blocked` filters it out and a human re-adds `ai-ready` when unblocking. |
| Pass 3 ping-pong stop | **Do not re-queue.** Same reasoning. |
| Implementer give-up | **Do not re-queue.** Same reasoning. |

The three `ai-blocked` exits are covered by one sentence in Pass 2 rather than three copies, and the Labels table now defines `ai-ready` as cleared on pickup, `ai-wip` as never riding alongside it, and `ai-blocked` as human-only re-queue.

## Size

963 → 980 lines (+17; +28/-11). Kept deliberately small given #474 just landed on keeping this file terse — the changes sharpen existing sentences rather than adding paragraphs beside them.

## Verification

`vitest run` — 1008 passed, 18 skipped, 61 files. The `src/base/labels.ts` test asserting the `gh label create` block is unaffected; that block is unchanged, only the prose Labels table moved.

Closes #473

### DIFF
--- a/skills/ai-issue-loop/SKILL.md
+++ b/skills/ai-issue-loop/SKILL.md
@@ -70,9 +70,9 @@ drift with a second copy to maintain.
 
 | Label | On | Meaning |
 |---|---|---|
-| `ai-ready` | issue | Eligible for an agent. The hard gate. |
-| `ai-wip` | issue | Claimed; a worktree exists. |
-| `ai-blocked` | issue | Agent gave up; needs a human. |
+| `ai-ready` | issue | Eligible for an agent. The hard gate; **cleared on pickup**. |
+| `ai-wip` | issue | Claimed; a worktree exists. Never rides alongside `ai-ready`. |
+| `ai-blocked` | issue | Agent gave up; needs a human. Only a human re-adds `ai-ready`. |
 | `ai-review` | PR | Awaiting agent review. |
 | `ai-reviewing-code` | PR | `code-reviewer` claimed and running. Cleared with its verdict. |
 | `ai-reviewing-sec` | PR | `security-expert` claimed and running. Cleared with its verdict. |
@@ -405,8 +405,12 @@ gh issue edit <N> --remove-label ai-wip 2>/dev/null
 A closed-unmerged PR is the exception: there is no squash to find, so skip the
 confirmation and remove — the work was abandoned deliberately.
 
-The issue itself closes from the PR body's `Closes #N`. This pass is what frees
-concurrency slots, so it must run before Pass 4.
+The issue itself closes from the PR body's `Closes #N`, so that edit is normally a
+no-op on a closed issue. A PR that said only `Refs #N` leaves it **open** — add
+`--add-assignee @me` in that case. The work has landed, so it must not go back in
+the queue; pickup already dropped `ai-ready`, and assigning it is what stops a
+merged issue sitting unowned instead (#429 had to be moved to `holding` by hand).
+This pass is what frees concurrency slots, so it must run before Pass 4.
 
 **Then reap the stalled.** Nothing can time out an agent: the Agent tool takes no
 timeout, and an agent whose session died leaves its labels behind with no process
@@ -432,8 +436,10 @@ work must never be reaped out from under itself.
 The **no PR exists** condition on the first row is what makes reaping safe. An
 agent that got as far as opening a PR has handed off to the label state machine
 and is no longer the thing being waited on; only a run that produced nothing is
-presumed dead. The reaped issue keeps its worktree removed, so a re-labelled
-`ai-ready` starts clean.
+presumed dead. Reaping deliberately does **not** restore `ai-ready` — `ai-blocked`
+means a human decides when the issue re-enters the queue, and the removed worktree
+means their re-label starts clean. The other two `ai-blocked` exits, Pass 3's
+ping-pong stop and an implementer handing back, leave it off for the same reason.
 
 **Every `ai-blocked` must say why, and land in front of a human.** So reaping always
 does three things together — label, assign, comment — and the comment opens with
@@ -448,9 +454,12 @@ puts it in the statusline and fires a notification with a sound.
 **Reaping is not always the right call — say so when it isn't.** The rule assumes a
 dead agent, but a stale `ai-wip` can also come from a run that was cancelled
 deliberately, in which case the work is fine and only the claim is stale. If you know
-the cause and it is benign, clear `ai-wip` **without** `ai-blocked` so Pass 4 can pick
-it straight back up, and say in the comment that you deviated and why. `ai-blocked`
-means *a human must look*; do not spend it on a claim you already understand.
+the cause and it is benign, **return it to the queue** — `gh issue edit <N> --add-label
+ai-ready --remove-label ai-wip`, no `ai-blocked` — so Pass 4 picks it straight back up,
+and say in the comment that you re-queued it, that you deviated, and why. Re-adding
+`ai-ready` is not optional: pickup cleared it, so clearing `ai-wip` alone drops the
+issue out of the queue silently, which is the worse failure. `ai-blocked` means *a
+human must look*; do not spend it on a claim you already understand.
 
 ### Pass 3 — review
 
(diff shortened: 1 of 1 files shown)


rtorcato (commented):

Diff-scoped review against the stated concerns. The pickup claim changes from add-only (`--add-label ai-wip`) to an atomic add+remove (`--add-label ai-wip --remove-label ai-ready`) in one `gh issue edit` call, so the claim is no less atomic than before — no new double-claim window is introduced. The eligibility query's `ai-wip == null` filter and the author-association backstop (`OWNER`/`MEMBER`/`COLLABORATOR`) are untouched by this diff.

The label is now exclusive rather than accumulating, and every path that clears `ai-wip` is audited: the `ai-blocked` exits (stall reaping, Pass 3 ping-pong stop, implementer give-up) explicitly do **not** re-add `ai-ready`, only a human does that; the one automatic re-queue path (Pass 2's benign-stall reap) only ever fires on an issue that still carries stale `ai-wip` from an already-claimed, in-flight attempt — never on an issue a human pulled `ai-ready` from directly, since such an issue would never have picked up `ai-wip` in the first place. No path treats "was once claimed" as a substitute for the live label.

### Before merging

Nothing.

rtorcato (commented):

Audit checks out for four of five transitions: the benign-stall path now re-adds `ai-ready` with an inline command and a comment note (fixing the highest-risk case as required), stall reaping and the two other `ai-blocked` exits are confirmed unchanged and the prose now says why in one place, and the Labels table/diagram already matched and needed no edit.

The merged-PR cleanup (case 3, the actual #429 case) is where the fix is incomplete. The prose at line 408-413 now says a `Refs #N` PR should get `--add-assignee @me` instead of re-entering the queue, but the actual command block just above it (line 402, `gh issue edit <N> --remove-label ai-wip 2>/dev/null`) was not updated — it's unconditional, with no branch that detects "issue still open" and applies the assignee. Every other transition in this file pairs its decision with an exact `gh` invocation; this one is prose-only, so an agent following the code block literally reproduces the original #429 gap this PR claims to close.

The +17 line growth is earned — it's audit reasoning the issue explicitly asked for, not padding, and mostly reworded existing sentences rather than adding beside them.

### Before merging
Add the missing command for the `Refs #N` branch in the merged-PR cleanup section (e.g. check `gh issue view <N> --json state` after removing `ai-wip` and run `gh issue edit <N> --add-assignee @me` if still open), so the fix for #429 is actually executable and not just described.

rtorcato:
🤖 *Automated — fix-round implementer via ai-issue-loop. Posted under the owner's account by an agent; not a human message. There is no separate GitHub account for AI agents, so this appears under @rtorcato's avatar.*

Addressed the `code-reviewer` finding on the merged-PR cleanup. The `Refs #N` branch is now a command in the block, not prose: after `--remove-label ai-wip`, an `if [ "$(gh issue view <N> --json state -q .state)" = OPEN ]` guard runs `gh issue edit <N> --add-assignee @me`. It no-ops on the ordinary `Closes #N` case, where the issue is already CLOSED by then. Reworded the paragraph below to point at that check rather than describe a missing one.

`skills/ai-issue-loop/SKILL.md` is now **984 lines** (+4 net). `tests/base/labels.test.ts` and `tests/cli/generators/claude-skills.test.ts` pass (30 tests).

rtorcato (commented):

The four previously-approved transitions are unchanged in this diff (labels table wording, Pass 4's claim now dropping `ai-ready`, the benign-stall re-queue, and reaping's non-restore) — still fine, no re-derivation needed.

The fix closes the gap. The added block runs `gh issue view <N> --json state -q .state` after `--remove-label ai-wip` and only assigns `@me` when the result is `OPEN` — which is exactly the `Refs #N` case, since a `Closes #N` issue is already `CLOSED` by then, so the ordinary path is a correct no-op. The command block and the reworded prose now agree, matching the pairing style used by every other transition in the file. On a `gh issue view` failure the substitution comes back empty, the comparison is false, and the block just skips the assign — that fails toward "no assignee" rather than toward a merged issue silently re-entering the queue, so it degrades in the safe direction even without `2>/dev/null` on that line (the one stylistic inconsistency worth a nit, not a block). Growth is the minimum four lines needed for the check itself — no padding.

### Before merging
Nothing.

rtorcato (commented):

Diff changed since last pass: yes — commit 1c0ccae turns the previously prose-only `Refs #N` case into an actual command (the `gh issue view ... -q .state` guard + `--add-assignee @me`), which is what `code-reviewer` flagged. That block only sets an assignee; it does not touch `ai-ready`/`ai-wip`, so it cannot return the issue to the agent-eligible pool — the pickup atomicity and authorization gate audited last round are unaffected.

Failure behaviour: if `gh issue view` errors, its stdout capture is empty (stderr isn't part of the substitution), so `[ "" = OPEN ]` is false and the block is a no-op — this fails safe (skips the assignment), not open.

Nothing else new beyond that block and the adjoining reworded paragraph.

### Before merging

Nothing.

Issue #330, opened 2026-08-07
biome peer range ^2.0.0 is wider than the shipped preset supports (needs ^2.5.0)


rtorcato:
Found while migrating `js-common` off the old package name (rtorcato/js-common#132).

`package.json` declares:

```json
"peerDependencies": { "@biomejs/biome": "^2.0.0" }
```

But the shipped preset `tooling/biome/biome.json` uses `linter.rules.preset`:

```json
"linter": { "rules": { "preset": "recommended", ... } }
```

`preset` is a Biome **2.5** key. On 2.4.x and earlier the equivalent is
`recommended: true`, and 2.4.9 rejects the config outright rather than
degrading:

```
× Found an unknown key `preset`.
i Known keys: recommended, a11y, complexity, correctness, nursery,
  performance, security, style, suspicious
```

The `$schema` on the same file already points at `2.5.0`, so the preset is
self-consistently a 2.5 artifact — only the peer range disagrees.

## Impact

Any consumer resolving Biome in `2.0`–`2.4` gets a hard configuration error
on every `biome check`, with no indication that the peer range is the cause.
The range says the pairing is supported; it isn't. `js-common` had
`^2.3.0`, resolved 2.4.9, and its lint step failed the moment it moved onto
the new package.

## Suggested fix

Narrow the peer range to `^2.5.0`. That turns a confusing config error into
an install-time peer warning that names the real problem.

Worth a look at whether the other presets have the same drift — the peer
ranges appear to predate several of the configs they gate, so this may not
be limited to Biome.

## Possible doctor check

This is the shape of thing doctor could catch on this repo itself: a peer
range whose floor the repo's own devDependency has already moved past is a
reliable signal that the shipped config has outrun what it claims to
support. Happy to file that separately if it's wanted.

rtorcato:
Building this surfaced a problem with the check as I specified it above.

I implemented the proposed rule — every peer range compared against its
devDependency floor — and measured it against this repo. **24 findings**,
nearly all of them correct practice:

```
prettier                          peer ^3.0.0,  dev ^3.9.6
@typescript-eslint/eslint-plugin  peer ^8.0.0,  dev ^8.65.0
rollup                            peer ^4.0.0,  dev ^4.62.0
tslib                             peer ^2.0.0,  dev ^2.8.0
...
```

Declaring support for a whole major while developing against the newest
release of it is exactly how a peer range is *supposed* to be written. So my
[...]

PR #477, opened 2026-08-21
fix(biome): stop pinning a version in the $schema URL

### LINKED ISSUE #468: Every @biomejs/biome bump fails the build: nothing updates the $schema pin to match

rtorcato:
🤖 *Automated — implementer via `ai-issue-loop`. Written by an AI agent and posted under the owner's account; not a human message. There is no separate GitHub account for AI agents, so this appears under @rtorcato's avatar.*

Closes #468.

Implements the owner's decision recorded on the issue (2026-08-21): **option 1, stop pinning a version in `$schema`**.

## Why `latest` and not dropping the key

Both work — verified against the installed 2.5.5:

| `$schema` | version notice | runs |
|---|---|---|
| pinned `2.5.0` | 2 lines | ✓ |
| `latest` | **none** | ✓ |
| omitted | none | ✓ |

`latest` wins because the key exists for exactly one reason: editors resolve it for autocomplete and validation of `biome.json`. Omitting it removes the failure class *and* that benefit; `latest` removes only the failure class. It costs nothing — Biome does not fetch the URL, it only parses a version out of it, and there is now nothing to parse.

## What moved

Both stale pins, as flagged on the issue:

- `biome.json` (root) — was 2.5.5
- `tooling/biome/biome.json` — the **shipped preset**, was 2.5.0
- `apps/docs/biome.json` — was 2.5.5
- `src/cli/generators/linting.ts` — the generator that emits `$schema` for scaffolded repos, now a constant
- `apps/docs/docs/guides/cli.md` — the `fix biome --diff` sample output

`fix biome` routes through `generateBiomeConfig`, so the fixer writes the unpinned URL with no separate change.

`resolveBiomeSchemaVersion` and `installedBiomeVersion` had no callers left once the emitted URL became a constant, so they were deleted along with `BIOME_SCHEMA_FALLBACK` and a now-unused JSON helper. Net **-40 lines**.

## For your decision

You asked whether #435's `$schema` drift check in `checkBiome` (`src/languages/js/checks.ts`) should be **removed** or **repurposed**, since the drift it looks for can no longer occur in a repo scaffolded by this version.

**I repurposed it, and I think that is right — but this is the part to overrule if you disagree.**

The check now flags **any surviving version pin**, whatever version it names, rather than comparing against the installed CLI:

```
biome.json pins Biome 2.4.16 in $schema — the next @biomejs/biome bump past it breaks every Biome run
→ Run `npx @rtorcato/repo-tooling fix biome` to drop the version from the $schema URL
```

Three reasons:

1. **Every existing consumer repo still has a pin.** They keep hitting this failure on every Biome bump until someone runs `fix biome`. Unpinning the generator fixes new repos and does nothing for them. This is the only thing that tells them.
2. **The old condition was too narrow anyway.** It only fired when the pin disagreed with the installed CLI — so a repo pinned to exactly what it has installed stayed green while sitting one Dependabot PR away from red. The pin that matches today is the one that breaks tomorrow. That is a widening, not just a re-aim, and it is why the old "matches the installed CLI" test now asserts `drift`.
3. **It costs almost nothing and gets simpler.** It no longer resolves `node_modules`, so it also works on a fresh clone before install, and it goes quiet permanently once a repo is migrated.

Case for **removing** it instead, so you have it: a check that only ever fires during a one-time migration is arguably a to-do disguised as a check, and it will sit in `doctor` forever after the last repo migrates. If you prefer that, deleting the loop in `checkBiome` and reverting it to a plain `checkFile` call is a small follow-up — the file-shape matching is untouched by this PR.

Note this changes doctor's verdict on repos that were previously green: any repo with a pinned `$schema` now reports Biome drift. That is the intent, and `fix biome` resolves it in one command.

**`checkConfigSchemaVersions` (#330) is deliberately untouched.** It degrades correctly on its own — `schemaUrlVersion` returns `null` for `latest`, so the config is skipped and it reports "no versioned config schemas to compare". It is generic over tools, so it keeps its value the moment a second entry lands in `SCHEMA_CONFIGS`.

## Commit type

`fix`, so this cuts a **patch**. It repairs a failure class rather than adding a capability, and the doctor change re-aims an existing check rather than introducing one. Flagging it explicitly because it edits a published file (`tooling/biome/biome.json`) and changes the output every scaffolded repo gets — if you read the new doctor verdict as significant enough to want a minor, say so and I will amend to `feat`.

## Verification

All run against the worktree's binaries, no repo-wide `--write` pass:

- `biome lint .` — clean (62 pre-existing warnings, all `apps/docs` CSS, unchanged by this PR)
- `vitest run` — **1009 passed, 18 skipped, 0 failed** (61 files)

Confirmed independently that `latest` is accepted with no version notice, running the installed 2.5.5 against a scratch config.

Tests updated: `tests/cli/commands/doctor.test.ts` (pinned-URL drift, including the new matches-installed case, plus the unpinned-is-ok case) and `tests/cli/generators/linting.test.ts` (generator emits `latest`; shipped preset is unpinned — the second stale pin now has a test guarding it).

Also noted on the issue: #455 was hand-fixed to 2.5.8 to unblock it. That hand fix becomes unnecessary once this lands, though it is harmless to leave.

### DIFF
--- a/apps/docs/biome.json
+++ b/apps/docs/biome.json
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://biomejs.dev/schemas/2.5.5/schema.json",
+  "$schema": "https://biomejs.dev/schemas/latest/schema.json",
   "root": false,
   "extends": ["../../tooling/biome/biome.json"],
   "files": {
--- a/apps/docs/docs/guides/cli.md
+++ b/apps/docs/docs/guides/cli.md
@@ -146,7 +146,7 @@ $ npx @rtorcato/repo-tooling fix biome --diff
     -  "rules": { "noConsole": "error" }
     -}
     +{
-    +  "$schema": "https://biomejs.dev/schemas/2.5.0/schema.json",
+    +  "$schema": "https://biomejs.dev/schemas/latest/schema.json",
     +  "extends": ["@rtorcato/repo-tooling/biome"],
     +  …
     +}
--- a/biome.json
+++ b/biome.json
@@ -1,4 +1,4 @@
 {
-  "$schema": "https://biomejs.dev/schemas/2.5.8/schema.json",
+  "$schema": "https://biomejs.dev/schemas/latest/schema.json",
   "extends": ["@rtorcato/repo-tooling/biome"]
 }
--- a/src/cli/generators/linting.ts
+++ b/src/cli/generators/linting.ts
@@ -44,61 +44,17 @@ export const BIOME_CONFIG = 'biome.json'
 export const BIOME_LEGACY_CONFIG = 'biome.jsonc'
 
 /**
- * Used when nothing in the target repo says which Biome will read the config —
- * a bare directory mid-`setup`, before `pnpm install` has run. Matches the
- * `$schema` the shipped preset carries.
- */
-const BIOME_SCHEMA_FALLBACK = '2.5.0'
-
-async function readJsonOrNull(filepath: string): Promise<Record<string, unknown> | null> {
-	try {
-		return (await fs.readJson(filepath)) as Record<string, unknown>
-	} catch {
-		return null
-	}
-}
-
-/**
- * The `@biomejs/biome` actually installed under `targetDir`, or null when
- * node_modules isn't populated. This is the binary that will parse the config
- * and emit the schema-mismatch warning, so it is the only version worth
- * comparing a written `$schema` against — a declared range like `^2.5.0` says
- * nothing about which 2.x resolved (#424).
- */
-export async function installedBiomeVersion(targetDir: string): Promise<string | null> {
-	const installed = await readJsonOrNull(
-		path.join(targetDir, 'node_modules', '@biomejs', 'biome', 'package.json')
-	)
-	const match =
-		typeof installed?.version === 'string' ? /(\d+\.\d+\.\d+)/.exec(installed.version) : null
-	return match?.[1] ?? null
-}
-
-/**
- * The Biome version the scaffolded `$schema` URL should name (#363).
+ * Unversioned on purpose (#468). A `$schema` naming an exact version is only
+ * ever right until the next `@biomejs/biome` bump, at which point Biome prints
+ * "The configuration schema version does not match the CLI version" on every
+ * invocation and doctor reports drift — so every Dependabot bump needed a hand
+ * edit before it could go green. Biome accepts `latest` and says nothing about
+ * it (verified against 2.5.5), which removes the failure class outright.
  *
- * A hardcoded version drifts the moment the consumer's Biome moves, and Biome
- * reports the mismatch on *every* invocation ("The configuration schema version
- * does not match the CLI version 2.5.7") — noise on a file the consumer never
- * wrote. The installed package wins because it is literally the binary that
- * will parse the file and emit that warning; the declared range is the next
- * best statement of intent when node_modules isn't populated yet.
+ * `latest` rather than dropping the key: editors still resolve a real schema
+ * for autocomplete and validation, which is the only reason the key is there.
  */
-export async function resolveBiomeSchemaVersion(targetDir: string): Promise<string> {
-	const installed = await installedBiomeVersion(targetDir)
-	if (installed) return installed
-
-	const pkg = await readJsonOrNull(path.join(targetDir, 'package.json'))
-	const deps = {
-		...((pkg?.dependencies as Record<string, string> | undefined) ?? {}),
-		...((pkg?.devDependencies as Record<string, string> | undefined) ?? {}),
-	}
-
-	// A range ("^2.5.0") carries the x.y.z we need just as an exact version does.
-	const declared = deps['@biomejs/biome']
-	const match = typeof declared === 'string' ? /(\d+\.\d+\.\d+)/.exec(declared) : null
-	return match?.[1] ?? BIOME_SCHEMA_FALLBACK
-}
+const BIOME_SCHEMA_URL = 'https://biomejs.dev/schemas/latest/schema.json'
 
 /**
  * The thin pointer config — the same shape this repo dogfoods. `fix biome` used
@@ -111,9 +67,8 @@ export async function generateBiomeConfig(targetDir: string): Promise<string> {
 	// file globs via `files.includes`; emitting the old 1.x `include`/`ignore`
 	// keys here forced consumers to run `biome migrate` before `biome check`
 	// would run at all.
-	const version = await resolveBiomeSchemaVersion(targetDir)
 	const biomeConfig = {
-		$schema: `https://biomejs.dev/schemas/${version}/schema.json`,
+		$schema: BIOME_SCHEMA_URL,
 		extends: ['@rtorcato/repo-tooling/biome'],
 	}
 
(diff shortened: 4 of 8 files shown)


rtorcato (commented):

The dead-code deletion is internally consistent: `installedBiomeVersion` and `resolveBiomeSchemaVersion` lose their only caller in the same diff (the `checks.ts` import is removed alongside), and the unexported `readJsonOrNull`/`BIOME_SCHEMA_FALLBACK` were never reachable from outside the module. The one thing the diff itself can't settle is whether either exported function was re-exported from the package's public entry point — a `tsc --noEmit` pass is consistent with "no", but doesn't prove it, since deleting a re-exported symbol is a compile error only if something still imports it *by that name from that path*, not if a barrel just stops listing it.

The repurposed `checkBiome` check is correct as written: it gates on `url?.includes('biomejs.dev')` before parsing a version, so a non-Biome `$schema` or a missing key both fall through to no drift, and `latest`/absent-version URLs parse to `null` and are silently treated as ok — matching the new "unpinned is fine" test. The widening argument (installed-matching pin was one bump from red) holds and is exactly what the added "matches installed CLI" test now asserts as drift. The doctor blast-radius claim checks out too: the hint text matches the actual fix command, and `generateBiomeConfig` now writes the constant unpinned URL with no separate fixer change, which the new "reports ok for the unpinned $schema" test exercises directly. `checkConfigSchemaVersions`/`schemaUrlVersion` aren't touched by this diff, so the #330 untouched claim is plausible but not independently checkable from the diff alone.

The `fix` vs `feat` semver question and repurpose-vs-remove are both flagged by the author as open calls, not something this review is resolving.

### Before merging
Confirm `installedBiomeVersion`/`resolveBiomeSchemaVersion` weren't re-exported from a package-level barrel (e.g. `src/index.ts`) before treating their removal as a non-breaking internal cleanup — if they were part of the public API, this is a breaking change regardless of internal callers.

rtorcato (commented):

No secrets, injection, or unsafe shell construction in this diff — it's a JSON `$schema` URL string, a generator constant, and a doctor check re-aim, not a new dependency or exec path. Verified `https://biomejs.dev/schemas/latest/schema.json` resolves and returns a real Biome Configuration schema (fetched it directly), so the PR's premise holds and editor autocomplete/validation will keep working — the concern about it 404ing does not apply.

`installedBiomeVersion` and `resolveBiomeSchemaVersion` have no remaining callers in the diff (the import in `checks.ts` is removed alongside them), and nothing security-relevant depended on knowing the exact installed version — it only fed the `$schema` string. The repurposed `checkBiome` in `src/languages/js/checks.ts` is confirmed strictly more likely to fire: the new test `reports drift for a pinned $schema even when it matches the installed CLI` shows a previously-green case (pin == installed version) now correctly reports drift, so this is a widening, not a weakening.

One trade-off the PR states only its upside on: unlike a pinned version (fixed by the installed package, moves only on a visible Dependabot bump), `latest` is a live remote document with no lockfile-equivalent — the schema Biome's own docs serve there can change at any time with zero corresponding diff in any consumer repo. That's a reasonable trade for editor autocomplete, but it's not risk-free the way "costs nothing" implies; a future upstream schema change could make editors flag a previously-valid `biome.json` with no local change to point to.

### Before merging
Consider stating in the PR description that `latest` trades a visible, Dependabot-tracked version bump for a silent, unversioned upstream document — not a strictly free upgrade over pinning. Non-blocking.

rtorcato:
🤖 *Automated — `ai-issue-loop` orchestrator. Posted under the owner's account by an agent; not a human message.*

**Resolving `code-reviewer`'s open item: the deletion is not breaking.** It could not confirm this diff-scoped; checked outside that scope.

`package.json` has **no root `.` export**, and `main`, `module` and `types` are all null. The only entry points are the ~30 `./tooling/*` preset subpaths and the `repo-tooling` bin. None resolves into `src/languages/js/checks.ts` or `src/base/`.

`dist/languages/**/*.js` does ship via `files`, but an `exports` map without a matching subpath makes deep imports unreachable — Node refuses them. And the only files referencing `installedBiomeVersion` / `resolveBiomeSchemaVersion` are `src/cli/generators/linting.ts` and `src/languages/js/checks.ts`, both inside this PR.

So they were internal-only. No consumer could have imported them.

rtorcato:

**Sent back: `DIRTY`.** Both reviewers passed, but #455 merged (`3f5f1bb`) and conflicts on one line.

#455 carried a hand-fix pinning `biome.json` `$schema` to **2.5.8**; this PR rewrites that line to `latest`. Resolution is to keep `latest` — that hand-fix existed only to unblock #455 and is exactly what this PR makes unnecessary.

Rebase only, no logic change. Both reviews cleared as stale and will be redone.

Two open questions still stand for the human afterwards: whether `https://biomejs.dev/schemas/latest/schema.json` actually resolves (editor autocomplete is the whole reason to keep the key), and `fix` vs `feat` given doctor's verdict changes for existing repos.

Issue #478, opened 2026-08-21
ai-notes conflates merge-blockers with follow-up work, and follow-ups die on merge


rtorcato:
🤖 *Filed by Claude (AI agent) on rtorcato's behalf. Posted under the owner's account; there is no separate GitHub account for AI agents.*

`skills/ai-issue-loop/SKILL.md` ships to npm, so this is published behaviour.

## The defect is in the label's own definition

> The bar is a finding that **changes what a human would do**: a semver implication, a deliberate omission, **a follow-up that must be filed**.

It says *must be filed* — and the process never files anything. The follow-up stays as prose in a PR comment, and a merged PR's comments are archaeology. Nobody re-reads them.

Worse, those three examples are not one kind of thing. `ai-notes` collapses them, so the reader cannot tell from the label whether they must act before merging or may act later.

## Three kinds, currently indistinguishable

Real examples from one session:

| Finding | Actually is | Today |
|---|---|---|
| #470 — is this `feat:` or `feat!:`? | **Decides the merge.** The squash subject cuts the release. | `ai-notes` |
| #477 — `latest` trades a Dependabot-visible bump for a silent upstream document | **Context.** Changes nothing you do, worth knowing. | `ai-notes` |
| #469 — `root: false` untested against a consumer nested in a larger monorepo | **Follow-up work.** A real task this PR does not do. | `ai-notes` |
| #469 — reformatting shipped configs may make consumers' copied assets report drift | **Follow-up work**, and consumer-facing. | `ai-notes` |
| #474 — the new rule duplicates an existing one | **Follow-up work**, trivial. | `ai-notes` |

Three of five are follow-ups. On merge, all three evaporate.

## Proposal

| Kind | Blocks? | Where it lives |
|---|---|---|
| Decides the merge | Effectively yes — must be answered first | PR comment + `ai-notes`, as today |
| **Follow-up work** | **No** | **A filed issue**, labelled `ai-suggested`. The PR comment becomes one line and a link. |
| Context / trade-off | No | One line in the PR comment |

**Filing is the alternative to blocking, not a precondition for it.** The current setup feels like it should block precisely because the finding has nowhere else to go — give it an issue and that pressure disappears.

Mechanically this is cheap: the reviewer already has `gh`, and GitHub cross-links the issue and PR automatically, so the trail survives the merge in both directions.

## `ai-suggested` must be a backlog, not a queue

The obvious failure mode is agents filling the tracker with speculative issues.

[...]

PR #482, opened 2026-08-21
fix(skills): cut reviewer comment boilerplate in ai-issue-loop

### LINKED ISSUE #479: Reviewer comments: drop the header boilerplate, negative findings, and confirmations of the PR's own claims

Follow-on to #472/#474. That capped reviewer bodies at "3 short paragraphs", which was the wrong instrument — a paragraph has no length, so comments obeying the cap are still too long. Three concrete rules instead.

## 1. Drop the header boilerplate

Every agent comment currently opens with some form of:

> 🤖 *Automated — `security-expert` via ai-issue-loop. Posted under the owner's account; not a human review. There is no separate GitHub account for AI agents, so this appears under @rtorcato's avatar.*

**Keep the attribution, drop the explanation.** Naming the agent is load-bearing — it is why the header exists, so a detailed security review is not mistaken for a human's. Explaining *why* it wears the owner's avatar is read once and then never again, on every comment forever.

```
```

**5 sites** mandate the long form (lines 53, 441, 509, 672, 861 in the shipped copy).

## 2. Do not report what you checked and found clean

Negative findings are the single biggest source of length. From an actual review:

> No secrets, injection, or unsafe shell construction in this diff — it's a JSON `$schema` URL string, a generator constant, and a doctor check re-aim, not a new dependency or exec path.

That is a paragraph saying *nothing happened*. `### Before merging: Nothing.` already carries it.

The cause is prompt shape: a brief that says "check X, Y and Z" invites "I checked X, Y and Z." The prompt should ask for the **conclusion**, not the walk.

## 3. Do not confirm the PR's own claims

This is the rule that would cut the most, and nothing currently states it.

> `installedBiomeVersion` and `resolveBiomeSchemaVersion` have no remaining callers in the diff […] The repurposed `checkBiome` is confirmed strictly more likely to fire […]

Both are true, and both were already asserted in the PR body. **Speak only where the PR is wrong, or silent.** A review that agrees with the PR should be nearly empty — that is what a pass label is for.

Verification still happens; it just does not get narrated. The reviewer that quietly confirmed four claims and flagged the fifth writes one sentence, not four paragraphs.

## 4. Budget in characters, not paragraphs

**≤600 characters above `### Before merging`.** Checkable; "short paragraph" is not.

## Worked example

The review that prompted this ran ~1,400 characters across three paragraphs, of which one carried new information. Under these rules:

> 🤖 *Automated review — `security-expert` via ai-issue-loop.*
>
> `latest` trades a Dependabot-visible version bump for a live upstream document with no lockfile equivalent — an upstream schema change could make editors flag a valid `biome.json` with no local diff to point at. The PR states only the upside.
>
> ### Before merging
> Consider saying in the PR description that `latest` is not a strictly free upgrade over pinning. Non-blocking.

Same signal. About a third the length, and the part that matters is no longer buried behind two paragraphs of "nothing wrong here".

## Scope

`skills/ai-issue-loop/SKILL.md` — both reviewer prompt templates, the header spec, and the four other header mandate sites. No behaviour, labels or merge conditions change.

⏳ **This cannot reach the running skill on its own** — see the fork issue filed alongside this one. The installed copy at `~/.claude/skills/` has diverged from what the package ships, so a fix here does not propagate until that is reconciled.

## Related

- #472 / #474 — the first pass at comment length
- #478 — `ai-notes` conflating merge-blockers with follow-up work

rtorcato:

The "3 short paragraphs" cap from #474 was the wrong instrument — a paragraph has no length, so comments obeying it were still long. Three enforceable rules replace it.

## What changed

**1. Short header, at all 5 mandate sites.** Attribution stays; the explanation of why the comment wears the owner's avatar goes.

```
```

Sites updated: the header spec at the top of the file, Pass 2 stall reaping, Pass 3 ping-pong, the reviewer prompt template, and the implementer prompt. The Dependabot prompt already back-referenced `🤖 *Automated review — …*`, so it inherits it. The prose introducing the requirement argued the reasoning across two paragraphs; it is one clause now.

**2. No negative findings.** The reviewer prompts enumerate what to check, which invited "I checked X, Y and Z". Each enumeration is now explicitly *"the checklist to run, not an outline to write up"*, and the budget paragraph states it directly.

**3. No confirming the PR's own claims.** New, and the rule that cuts most: *"Verify everything; narrate only where the PR is **wrong** or **silent**"* — verification still happens, it just is not narrated. Agreement is what the pass label is for.

**4. A character budget.** `≤600 characters above `### Before merging``, replacing the paragraph count in both the outcome table and both reviewer prompts.

## Line count

**984 → 981 lines** (51394 → 51009 characters). Shorter, as the issue required. Each rule replaced the prose it supersedes rather than being appended: the header's two-paragraph justification, `"no process narration, no restating the diff, no listing what you checked and found fine"`, `"Not observations, not praise, not restating the diff"`, and the decline section's restatement of the avatar rationale all came out.

## Unchanged

Labels, the state machine, merge conditions, the release-environment gate, the `### Before merging` / `Nothing.` convention, and the marker-upsert logic. Nothing from #478 attempted.

## Verification

`vitest run` — 1009 passed, 18 skipped, 61 files. Per the task brief I ran no `pnpm install` (the worktree's `node_modules` is a symlink) and no `biome --write` pass. The diff is `skills/ai-issue-loop/SKILL.md` and nothing else.

Note the issue's own caveat: the installed copy at `~/.claude/skills/` has diverged from what the package ships, so this does not reach the running skill until that is reconciled.

Closes #479

### DIFF
--- a/skills/ai-issue-loop/SKILL.md
+++ b/skills/ai-issue-loop/SKILL.md
@@ -43,14 +43,12 @@ real approvals.
 
 The same constraint makes everything an agent posts *look* hand-written by the
 owner. So **every comment any agent leaves — review, blocked, gave-up, declined —
-opens with a `🤖 *Automated …*` italic header line** naming which agent wrote it,
-followed by a blank line. Non-negotiable: a detailed security review under a
-human's avatar misrepresents who reviewed the code.
+opens with a `🤖 *Automated …*` italic header line naming which agent wrote it**,
+then a blank line. Name the agent and stop there: a detailed security review
+under a human's avatar misrepresents who reviewed the code, but *why* it wears
+that avatar is read once and then reread on every comment forever.
 
-Spell the reason out rather than assuming the reader knows the convention — the
-header names the agent *and* says why it is wearing a human's face:
-
-`🤖 *Automated — <which agent> via ai-issue-loop. Posted under the owner's account by an agent; not a human message. There is no separate GitHub account for AI agents, so this appears under @<owner>'s avatar.*`
+`🤖 *Automated — <which agent> via ai-issue-loop.*`
 
 **Comment budget: ≤10 lines, and a clean outcome gets no comment at all.** A
 40-line comment on every PR trains the reader to skip all of them, including the
@@ -63,7 +61,7 @@ drift with a second copy to maintain.
 | Clean and ready | **None.** `ai-ok-code, ai-ok-sec` + assigned + no `ai-review` already says it. |
 | `ai-notes` | ≤10 lines; link the reviewer's `### Before merging`. |
 | `ai-changes`, CI red, `ai-blocked` | ≤10 lines, action first, then the specific cause. |
-| Reviewer verdict | `### Before merging` plus at most 3 short paragraphs above it. |
+| Reviewer verdict | `### Before merging` plus ≤600 characters above it. |
 | Declining an issue | The one exception — a hard handoff needs its reasoning; see Pass 4. |
 
 ## Labels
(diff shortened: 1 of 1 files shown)


rtorcato (commented):

### Before merging
Nothing.

rtorcato (commented):

### Before merging
Nothing.

PR #483, opened 2026-08-21
feat(skills): tell a stale installed skill from a deliberate local fork

### LINKED ISSUE #480: fix claude-skills would silently overwrite the forked local ai-issue-loop skill

The `ai-issue-loop` skill that actually runs has **forked in both directions** from the one this package ships, and `fix claude-skills` would silently overwrite it.

## Evidence

| | shipped (`origin/main`) | running (`~/.claude/skills/…` → private dotfiles) |
|---|---|---|
| lines | 963 | **1084** |
| stamped version | *(stamped at install)* | **3.9.2** — npm is at 3.11.0 |
| `<!-- ai-issue-loop:decision -->` marker upsert | absent | **present** |
| `node_modules` rebuild after worktree removal | absent | **present** |
| `ai-reviewing-*` claim labels | **present** (9 refs) | absent |

Not "one is behind" — each has content the other lacks.

## Why it silently overwrites

`installSkill` (`src/cli/generators/claude-skills.ts:158`) guards on version only:

```ts
if (installedVersion && isNewerVersion(installedVersion, shipped.version)) {
    return { ...base, status: 'declined-downgrade' }
}
```

Installed `3.9.2` < shipped `3.11.0`, so this is not a downgrade and the write proceeds. The 1084-line file is replaced by the 963-line one, and the marker upsert and rebuild step are gone with no warning.

There is a content check for *copied assets* — #448, "tell a stale copied asset from a deliberate local fork" — but **skills do not go through it**. A skill is judged purely on the version stamp, and a fork that is merely older than the package looks identical to a stale copy.

Compounding it: the file is a **stow symlink into a private dotfiles repo**, so the overwrite lands in a second repo's working tree, where it shows up as an unexplained 121-line deletion.

## Consequences already visible

Fixes merged into the shipped skill do not reach the loop that is running. #474 capped comment length and merged; the running copy does not have it, so the comments it produces are unchanged. Anything landed for #476 or the comment-noise issue filed alongside this one has the same problem.

## What to decide

| Option | Notes |
|---|---|
| **Detect the fork and refuse** | Extend the #448 content-vs-version distinction to skills: if the installed content differs from the shipped content *and* is not simply an older stamp, refuse and print a diff. Safe by default, and the missing half of #448. |
| **Reconcile first, then keep in sync** | Port the marker upsert and the rebuild step upstream into the shipped skill, port `ai-reviewing-*` downstream, then let `fix claude-skills` own the file. Removes the fork rather than defending against it. |
| **Both** | Reconcile now; keep the guard so the next fork is caught rather than clobbered. |

The middle option needs a human — deciding which side of each divergence is correct is not something an agent should infer.

⏳ Worth resolving before the next `fix claude-skills` run, not after.

## Related

- #448 — the copied-asset fork detection this needs the equivalent of
- #474 — merged, and not in the running copy
- #478 — `ai-notes` semantics; same file, same propagation gap

rtorcato:

Closes #480

Option 1 only — detect the fork and refuse. Reconciling the two divergent copies of `SKILL.md` is a human decision and stays out of this PR; `skills/ai-issue-loop/SKILL.md` is untouched.

## What changed

`installClaudeSkill` guarded on the version stamp alone, so a fork that was merely *older* than the package was indistinguishable from a stale copy — the write proceeded and the local content was gone. Through a stow symlink that overwrite lands in a second repo's working tree.

This applies #448's mechanism rather than inventing a second one. Every install now stamps `repo-tooling-hash` — the sha256 of the pristine content — beside `repo-tooling-version`. The record lives in the file, not `.repo-tooling.json`, because skills are user-global and no one repo owns them.

| Installed content | `SkillContentState` | `fix claude-skills` |
|---|---|---|
| matches the recorded hash, or is verbatim what we ship now | `pristine` | updates |
| differs from the recorded hash | `modified` | `declined-fork` |
| no record at all (installed before this) | `unknown` | `declined-fork` |

The refusal is actionable — it names what diverged, the real path (`file → realFile` when it is a symlink, so the message points at the dotfiles checkout the write would have hit), a ready-to-run `diff` against the shipped file, and `--force-skills`.

`--yes` deliberately does **not** imply `--force-skills`. Unattended runs pass `--yes`, and this is the one overwrite that destroys work living outside the repo.

`doctor` reports a fork as `ok` and never as `needsInstall`, matching how #448 treats a modified copied asset: named once, never nagged as fixable. Pointing at a fix that would refuse is worse than saying nothing.

## Verification

1016 tests pass; `biome lint .` at the same 62 pre-existing warnings as `main`; `biome format .` and `tsc --noEmit` clean.

New tests cover: fork declines and the file is byte-identical afterwards; an unstamped copy declines; a stale-but-pristine copy still updates cleanly; `--force` overwrites; identical content is `up-to-date`; the hash survives a round trip through the stamps; and `needsInstall` stays false for a fork.

## For your decision

- **`unknown` refuses, and that changes the first run after upgrade.** Every copy installed before this has no hash, so the first `fix claude-skills` on an out-of-date skill will now refuse and ask for `--force-skills` once. I chose safety over convenience because your machine is exactly this case — a legacy stamp at `3.9.2` with 121 lines the shipped copy lacks. The alternative (treat "no record" as safe to overwrite, per #448's `unknown` → never drift) would leave the reported bug unfixed on precisely the install that reported it. It converges after one run.
- **The old `adopts an unstamped copy` test asserted the behaviour this issue calls the defect**, so it is replaced rather than kept.
- **Ordering:** a fork stamped *newer* than the package reports `declined-downgrade`, not `declined-fork`. Both refuse, so nothing is lost, but the message names the version rather than the divergence.

### DIFF
--- a/apps/docs/docs/guides/ai-issue-loop.md
+++ b/apps/docs/docs/guides/ai-issue-loop.md
@@ -23,7 +23,7 @@ the machine — which is why it is **opt-in**: a bare `fix` or `fix --yes` skips
 it and says so, and `doctor` reports it as *not configured* rather than as a
 finding against your repo.
 
-Three things follow from that:
+Four things follow from that:
 
 - **`--skills-dir <path>`** overrides the destination. It is required alongside
   `--yes` / `--json` when `~/.claude/skills` does not exist, since a prompt
@@ -37,6 +37,16 @@ Three things follow from that:
   `repo-tooling-version` stamp in its frontmatter. A repo pinned to an older
   release reports and skips rather than overwriting a newer skill — otherwise
   two repos on different versions would fight over it on every `fix`.
+- **The install refuses to overwrite a local fork.** Alongside the version, each
+  copy carries a `repo-tooling-hash` of the content we wrote. If the installed
+  file no longer matches that hash — or predates it, so nothing can be proven —
+  the install prints what diverged and stops, the same rule
+  [`fix copied-assets`](./cli.md) follows for copied presets. This is the case
+  the version stamp alone cannot see: a fork that is merely *older* than the
+  package looks exactly like a stale copy. Diff it against the shipped file the
+  message names, then pass **`--force-skills`** to take the shipped version.
+  `--yes` deliberately does *not* imply it: unattended runs pass `--yes`, and
+  this is the one overwrite that destroys work living outside the repo.
 
 Any agent that reads the [`skills`](https://www.npmjs.com/package/skills) CLI
 format can also take it straight from GitHub:
--- a/src/base/checks.ts
+++ b/src/base/checks.ts
@@ -544,6 +544,21 @@ export async function checkClaudeSkills(): Promise<CheckResult> {
 			hint,
 		}
 	}
+	// A local fork is `ok` for the same reason a modified copied asset is (#448):
+	// it is somebody's deliberate work, so it is named once and never nagged as
+	// fixable — pointing at a `fix` that would refuse is worse than saying nothing.
+	if (status.contentState && status.contentState !== 'pristine') {
+		const why =
+			status.contentState === 'modified'
+				? `has local changes since ${status.installedVersion}`
+				: 'carries no content record, so a fork cannot be told from a stale copy'
+		return {
+			check,
+			status: 'ok',
+			detail: `${SHIPPED_SKILL} skill at ${status.file} ${why}; this package ships ${status.shippedVersion} and will not overwrite it`,
+			hint: `Diff it against the shipped copy, then run \`npx @rtorcato/repo-tooling fix claude-skills --force-skills\` to take the shipped version`,
+		}
+	}
 	return {
 		check,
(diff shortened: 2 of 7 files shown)


rtorcato (commented):

### Before merging
Nothing.

rtorcato (commented):

Traced the hash round trip (`stampSkill` → `classifySkillContent` → `hashSkillContent`) and the `--force-skills`/`--yes` wiring through `cli/index.ts` → `fix.ts` → `fixers.ts`; both hold up under the failure modes flagged for review.

### Before merging

`doctor`'s fork hint (`src/base/checks.ts`) says "Diff it against the shipped copy" but never names that path — unlike the `fix` refusal in `fixers.ts`, which builds a runnable `diff "<realFile>" "<shippedFile>"`. Someone acting on `doctor` output alone has nothing to diff against. Minor; not blocking.

PR #489, opened 2026-08-21
feat(skills): route agent follow-ups to ai-suggested issues

### LINKED ISSUE #478: ai-notes conflates merge-blockers with follow-up work, and follow-ups die on merge

rtorcato:

Closes #478

`ai-notes` listed "a follow-up that must be filed" among its examples and then never filed anything. Verified against the file before changing it: that sentence was at `SKILL.md:90` and in the reviewer prompt at `SKILL.md:543`, and nothing anywhere in the skill created an issue.

## The split

| Kind | Blocks? | Now goes to |
|---|---|---|
| Decides the merge | Effectively yes | PR comment + `ai-notes`, unchanged |
| Follow-up work | No | An `ai-suggested` issue the reviewer files; comment keeps `Follow-up: #N` |
| Context / trade-off | No | One line in the comment body |

`Follow-up: #N` goes in the comment body **above** `### Before merging`, deliberately not inside it. That section's contents are what trigger `ai-notes`, so putting a follow-up there would make later work suppress Dependabot auto-merge. Keeping it out means the existing `ai-notes` trigger, the merge conditions, and the auto-merge gate are all untouched.

## `ai-suggested` is a backlog, not a queue

Added to Pass 4's eligibility `--jq` alongside the `holding` exclusion, with a one-line why. An agent can never pick up another agent's suggestion; promoting one is a human act. Also added to the Labels table and the `gh label create` block, matching the description already on the repo (`#c2e0c6`).

The bar is what `ai-notes` has today **plus** "is work someone would plausibly do" — stated once in the label definition and once in the reviewer prompt, matching how the file already carries that sentence in both places.

## Scope deviation — `src/base/labels.ts`

I was asked to touch `SKILL.md` only. That is not achievable together with a green test run: `src/base/labels.ts` is the canonical label table, and `tests/base/labels.test.ts:180` asserts the skill's bootstrap block equals it element-for-element. Adding `ai-suggested` to only one side fails that test. So the diff is two files, and the second is seven lines: the label spec, plus two comments that said "eleven labels" and now say twelve.

## Size

**981 → 1020 lines (+39).** It grew, against #482's brevity pass. The new rule has four obligatory touch points (label definition, bootstrap block, both reviewer prompts) and one of them needs a `gh issue create` block a fresh agent can copy. I replaced the "must be filed" prose at both sites rather than adding beside it, and the Dependabot prompt points at the generic one instead of restating it, but I could not get this under the old line count without dropping a rule.

## Out of scope, as instructed

The implementer's `## For your decision` sections were left alone. For what it's worth: **yes, they should follow**, and I'd do it as a separate issue. The argument is the same one this PR makes and the mechanism is already built after this lands — but the implementer writes into a PR body rather than a review comment, so the "where does the one-liner go without tripping `ai-notes`" question has a different answer there and deserves its own review.

## Checks

`vitest run` — 1017 passed, 18 skipped, 61 files. Pre-commit (`biome check`, `typecheck`) passed via the husky hook.

### DIFF
--- a/skills/ai-issue-loop/SKILL.md
+++ b/skills/ai-issue-loop/SKILL.md
@@ -60,6 +60,7 @@ drift with a second copy to maintain.
 |---|---|
 | Clean and ready | **None.** `ai-ok-code, ai-ok-sec` + assigned + no `ai-review` already says it. |
 | `ai-notes` | ≤10 lines; link the reviewer's `### Before merging`. |
+| Follow-up found | One line — `Follow-up: #<new>`. The issue carries the context. |
 | `ai-changes`, CI red, `ai-blocked` | ≤10 lines, action first, then the specific cause. |
 | Reviewer verdict | `### Before merging` plus ≤600 characters above it. |
 | Declining an issue | The one exception — a hard handoff needs its reasoning; see Pass 4. |
@@ -78,6 +79,7 @@ drift with a second copy to maintain.
 | `ai-ok-sec` | PR | `security-expert` passed. |
 | `ai-changes` | PR | A reviewer requested changes. Reviewers never apply it to a Dependabot PR. |
 | `ai-notes` | PR | Passed, but a reviewer left something to read before merging. |
+| `ai-suggested` | issue | Follow-up a reviewer filed. A triage queue, never auto-picked. |
 | `holding` | issue | A gate — closes on human judgement, never picked up. |
 
 **`ai-notes` is advisory and never blocks.** It rides *alongside* a pass label,
@@ -87,11 +89,18 @@ so `ai-notes` is the hold itself: it suppresses auto-merge and routes the PR to
 the human. It exists because a pass label currently means both "clean" and
 "I found something real but would not hold the PR over it", and those two are
 indistinguishable in the *Assigned to you* view where merges actually happen.
-The bar is a finding that **changes what a human would do**: a semver
-implication, a deliberate omission, a follow-up that must be filed. Not
+The bar is a finding that **changes what a human would do at merge time**: a
+semver implication, a deliberate omission, a question only they can answer. Not
 observations, not praise, not restating the diff. `ai-notes` on every PR is the
 failure mode — it trains the reader to ignore it, which is worse than not having
 it.
+
+**Follow-up work is an issue, not a note.** A finding that clears that bar *and*
+is work someone would plausibly do gets filed as its own issue labelled
+`ai-suggested`, by the reviewer that found it; the PR comment keeps one line and
+a link. It does **not** earn `ai-notes` — later work does not decide this merge.
+An observation is not a follow-up. Prose in a merged PR's comments is
+archaeology, which is how every follow-up left there so far has died on merge.
 
 First run in a repo, create any that are missing (`gh label create` is a no-op
 error if it exists — ignore that):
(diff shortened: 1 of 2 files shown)


rtorcato (commented):

Verified the `ai-suggested` exclusion is a real filter line in Pass 4's jq query, not just prose describing one (the #476 failure shape doesn't recur here). The +39 growth traces to real new content per hunk: the `gh issue create` block, the Dependabot arm pointing back at the generic prompt instead of restating it, and rationale paragraphs that match the file's existing convention for `holding`. Both reviewer prompts got the routing, and "a follow-up that must be filed" is gone from both copies of the bar definition. The "plausibly do" clause is stated in both the label definition and the reviewer prompt.

### Before merging
Nothing.

rtorcato (commented):

Verified against the diff, not the PR's description: the `ai-suggested` exclusion is a real `select([.labels[].name] | index("ai-suggested") == null)` line in Pass 4's `--jq` filter, not just prose — the failure mode flagged for a prior PR on this file doesn't repeat here. The `author_association` check on that same query, and the `ai-ready` gate it backstops, are untouched.

The new `gh issue create` has no `--repo` flag, so it can only write to the repo in cwd — bounded, and consistent with every other `gh` call in this skill (none pin `--repo`, all rely on the worktree). Label color/description match exactly between `labels.ts` and the bootstrap block. No merge condition, gate, or the `🤖` header changed.

### Before merging
Nothing.

PR #490, opened 2026-08-21
feat(doctor): honour --skills-dir when reporting the Claude skill

### LINKED ISSUE #485: doctor ignores --skills-dir, so a custom skills path always reports as missing

`doctor` always reports the Claude skill against `~/.claude/skills`, whatever `--skills-dir` a repo actually uses.

- `src/cli/generators/claude-skills.ts:283` — `claudeSkillStatus(name, explicit?)` takes an override
- `src/base/checks.ts:528` — calls `claudeSkillStatus()` with **no arguments**, so `explicit` is dead
- `src/cli/index.ts:358` — `--skills-dir` is defined on `fix` only; `doctor` (line 341) has `-d` and `--json` and nothing else

So a consumer who installs with `fix claude-skills --skills-dir /custom/path` then runs `doctor` gets a permanent `optional-missing` for a skill they have installed. Read-only, so no stray writes — but it is a false negative in exactly the setup the flag exists to support.

Found while verifying #460's empty-`HOME` work; the flag is honoured on every path that *writes*, and only the reporting path was missed.

## Scope

Add `--skills-dir` to the `doctor` command and thread it to `claudeSkillStatus`. Match how `fix` documents and validates it. Worth checking whether any other `doctor` check has the same shape — an option honoured on write and ignored on read.

rtorcato:

`--skills-dir` was defined on `fix` only, so `doctor` always resolved the skill against `~/.claude/skills`. `claudeSkillStatus(name, explicit?)` already took the override; `checks.ts` called it bare, so a repo installed anywhere else got a permanent false `optional-missing` for a skill it has.

## What changed

- `--skills-dir <path>` on `doctor`, threaded action → `runDoctor` → base suite → `checkClaudeSkills`. No module-level state.
- No `--yes`/`--json` guard, unlike `fix`: doctor never writes, so an unresolved directory is just reported as not installed rather than being an error.
- `fix` now passes its own `--skills-dir` to the `runDoctor` call it makes, so the diagnosis it acts on and the install it performs agree on one directory.

## The other half: is anything else honoured on write and ignored on read?

No — and there is a structural reason, not just an empty grep.

- **Every other CLI option that affects a write has no read-side counterpart.** `--force-skills` is overwrite policy (doctor reports a fork as `ok` with a hint; there is nothing to honour). `--preset`, `--config`, `--skip-install` are `setup`-only inputs that leave no state outside the repo. `-d/--directory` is already honoured on both sides.
- **`--skills-dir` was the only candidate by construction**: `checkClaudeSkills` is, as its own doc comment says, the only check that reports on state *outside* the repo being audited. Every other check derives everything it reads from `dir`, so there is no second path where a flag could go missing.
- The other optional overrides in the check modules (`exec?: GhExec` in `labels.ts`, `milestones.ts`, `github-settings.ts`; `exec?: GitExec` in `git-identity.ts`) are test seams with no CLI flag behind them. Being called bare is what they are for.

## Tests

`tests/cli/commands/doctor.test.ts` — `doctor` reports the skill `ok` when `--skills-dir` points at a directory containing it, and `optional-missing` when it does not. The pair is revert-sensitive on any machine: whichever way the real `~/.claude/skills` happens to look, dropping the threading fails one of the two.

Verified: `biome lint .` (62 warnings, identical to `origin/main`), `tsc --noEmit -p src/cli/tsconfig.json`, `vitest run` (1019 passed).

`feat` rather than `fix`: it closes a bug report, but what ships is a new flag on a public command.

Closes #485

### DIFF
--- a/src/base/checks.ts
+++ b/src/base/checks.ts
@@ -521,11 +521,16 @@ export async function checkAiSetup(dir: string): Promise<CheckResult> {
  * is not a defect in this repo, and either of those statuses would fail its CI
  * over a file CI has never seen. `optional-missing` is the honest verdict — an
  * opt-in workflow tool that isn't configured — and it leaves the exit code alone.
+ *
+ * `skillsDir` is doctor's `--skills-dir`. Without it this reported against
+ * `~/.claude/skills` whatever directory the repo actually installs into, so a
+ * consumer who passes the flag to `fix` saw a permanent false `optional-missing`
+ * for a skill they have (#485).
  */
-export async function checkClaudeSkills(): Promise<CheckResult> {
+export async function checkClaudeSkills(skillsDir?: string): Promise<CheckResult> {
 	const check = 'Claude skills'
 	const hint = `Run \`npx @rtorcato/repo-tooling fix claude-skills\` to install the ${SHIPPED_SKILL} skill (writes outside the repo; opt-in, so \`fix\` alone skips it)`
-	const status = await claudeSkillStatus()
+	const status = await claudeSkillStatus(SHIPPED_SKILL, skillsDir)
 	if (!status.installed) {
 		return {
 			check,
--- a/src/cli/commands/doctor.ts
+++ b/src/cli/commands/doctor.ts
@@ -82,6 +82,8 @@ export { evaluateNodeVersion }
 export interface DoctorOptions {
 	directory?: string
 	json?: boolean
+	/** `--skills-dir`, mirroring `fix` — the read side of the same flag (#485). */
+	skillsDir?: string
 }
 
 const PACKAGE = '@rtorcato/repo-tooling'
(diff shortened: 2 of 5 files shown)


rtorcato (commented):

Verified: no new write path (`checkClaudeSkills` only reaches the existing `claudeSkillStatus` read call, `--skills-dir` is a plain string threaded through, never shell-interpolated), and `fix`'s use of its own `--skills-dir` for its internal `runDoctor` call just aligns the pre-existing diagnose/install directories rather than adding a new capability. Fork-guard/`--force-skills` hash logic is untouched.

### Before merging
Nothing.

rtorcato (commented):

Checked the revert-sensitivity claim by hand: `checkClaudeSkills()` bare falls back to `resolveSkillsDir(undefined)`, which resolves to the real `~/.claude/skills` (or `null`) — identically for both tests in the same run. So a revert gives both tests the *same* actual status, but the tests assert opposite expected statuses (`ok` vs `optional-missing`), so exactly one fails regardless of what's on the machine. The claim holds.

`fix`'s own `runDoctor(targetDir, options.skillsDir)` call is a genuine fix, not scope creep: `claudeSkillStatus`/`checkClaudeSkills` are read-only (`fs.pathExists`/`fs.readFile` only, confirmed in `claude-skills.ts`), so no write path is reached via that call. The asymmetry (no `--yes`/`--json` guard on doctor) is correctly justified.

`feat` vs `fix` labeling is a fair call given the new public flag — not blocking.
### Before merging
Nothing.

Issue #486, opened 2026-08-21
root: false on the shipped Biome preset is untested inside a nested monorepo


rtorcato:
🤖 *Filed by Claude (AI agent) on rtorcato's behalf.*

Raised by `code-reviewer` on PR #469 and unresolved when it merged.

`tooling/biome/biome.json` gained `"root": false` so a whole-repo `biome check` can run here. It was verified two ways — that the preset still resolves via `extends`, and that it still works when `copy biome` writes it as a consumer's **own root** `biome.json` with no parent above it.

**Not verified: a consumer nested inside a larger monorepo**, where a root config *does* exist above them. That is the configuration `root: false` actually describes, and it is the one case nobody tested.

## Scope

Construct the nested case — a parent `biome.json` at a monorepo root, the preset in a package beneath it — and confirm the package's config still applies rather than deferring to the parent. A lint config that silently resolves to nothing is a regression that looks like a green build.

If it holds, a test pinning it is worth more than the answer; if it does not, that is a consumer-facing bug in a shipped file.

rtorcato:
🤖 *Automated — `ai-issue-loop`.*

**Investigated in https://github.com/rtorcato/repo-tooling/pull/491** — tests only, and deliberately `Refs` rather than `Closes`, so this issue stays open for the decision below.

The question this issue asked — does the nested case hold? — **yes**. A package under a monorepo root carrying the preset verbatim applies it correctly, from any cwd.

Testing it surfaced two things nobody asked about:

**1. `copy biome` into a standalone repo silently ignores the preset.** `root: false` means "my root is above me". With no root above, Biome discards the config and falls back to built-in defaults — no warning, exit 0. Consumers get stock Biome, not the preset. Not a security regression: `security-expert` checked, and stock is *stricter* on the observed rule. It is config integrity, not exposure.

**2. `root` does not inherit through `extends`.** The pointer we scaffold has no `root` key, so Biome reads it as a root config: inside a monorepo it errors *"Found a nested root configuration"* and nothing is linted at all.

So `root: false`'s only working effect is letting **this** repo run `biome check .` — #469's goal. For consumers it is neutral-to-harmful.

## The decision

| Option | Cost |
|---|---|
[...]

PR #491, opened 2026-08-21
test(biome): pin which config wins for the shipped preset's root: false

rtorcato:

Refs #486. Leaving the issue open: the nested case it asked about holds, but testing it turned up two that do not, and the fix for those is a trade-off — see **For your decision**.

## What I ran

Everything below is `@biomejs/biome` **2.5.5**, run against throwaway fixtures outside the repo. No reasoning about how Biome "should" resolve configs — only observed output.

The probe file is `export const v: any = "x"`. That matters: `noExplicitAny` is **on** under Biome's built-in defaults and explicitly **off** in our preset, so its *absence* proves the preset was applied. `noDebugger` — the obvious probe — is recommended in both, so it is reported either way and would have passed against a config that had silently evaporated. My first pass used `noDebugger` and reached the wrong conclusion twice before I switched.

## 1. The case the issue asked about: it holds

Parent `biome.json` at a monorepo root (`recommended`), `packages/nested/biome.json` = the shipped preset verbatim, `packages/sibling` with no config at all.

| Run | Result |
|---|---|
| `biome check .` from the monorepo root | no config error; both packages checked |
| `biome check packages/nested` | `noExplicitAny` **not** reported → the nested preset won |
| `biome check packages/sibling` | `noExplicitAny` **reported** → the parent really was in force |
| From inside `packages/nested` | same as from the root; violations reported, exit 1 |

Formatting confirms it independently: the preset's `quoteStyle: single` is applied under `packages/nested` and not under the sibling. So the answer to all three questions in the issue is: the nested config applies, the violation is reported, and it does not differ by cwd.

## 2. `copy biome` into a standalone repo: the preset is silently ignored

`root: false` means "I am nested; my root is above me". When there is no root above, **Biome discards the config and falls back to its built-in defaults — with no warning, no error, and exit 0 on `--verbose`.**

`copy biome` writes `tooling/biome/biome.json` verbatim to the consumer's `biome.json`, at their repo root. So today, every repo scaffolded that way runs on stock Biome:

```
# preset copied in verbatim (has "root": false), no parent above
$ biome check src/probe.ts
src/probe.ts:1:17 lint/suspicious/noExplicitAny     <- preset turns this OFF
  × Formatter would have printed:  export const v: any = "x";   <- preset wants 'x' and no semicolon
```

Strip the `root` key from that same file and both diagnostics disappear. This is exactly the "resolves to nothing and looks like a green build" failure the issue predicted — just in the layout nobody was worried about.

I hit this in this repo too, while trialling a fix: adding `"root": false` to our own `biome.json` made `biome check --write` reformat 400+ lines to Biome defaults. Our root config has no `root` key, so **we are not affected** — but that is why the preset's `root: false` looked verified. It was never being read.

## 3. `root` does not inherit through `extends`

`generateBiomeConfig` writes a thin pointer, `{ "$schema": ..., "extends": ["@rtorcato/repo-tooling/biome"] }`, with no `root` key. The preset's `root: false` does **not** come through it, so Biome reads the pointer as a root config. Inside a larger monorepo:

```
$ biome lint .          # from the monorepo root
× Found a nested root configuration, but there's already a root configuration.
× Biome exited because the configuration resulted in errors.
```

Nothing is linted at all. Loud, not silent — but a scaffolded package inside someone else's monorepo cannot be checked from that monorepo's root. Adding `"root": false` to the pointer by hand fixes it (verified), which is also the proof that `extends` does not carry it.

This means the preset's `root: false` does nothing for `extends` consumers — the shape we actually scaffold — and actively breaks `copy biome` consumers. Its only working effect is letting *this* repo run `biome check .`, which was #469's goal.

## What I did not change

I tried to fix #3 by emitting `"root": false` on the pointer and reverted it: it fixes the nested case and silently disables linting for every standalone consumer, which is the larger and much worse population. Two dead ends also ruled out, both still hitting `Found a nested root configuration`:

- `files.includes: ["**", "!tooling/biome/**"]` on the root config — `includes` filters files, not config discovery.
- `files.experimentalScannerIgnores` — not a key Biome 2.5.5 accepts under `files`.

## This PR

Tests only, no shipped file touched. `tests/cli/generators/linting.test.ts` gains a `describe` that runs the real `biome` binary against all three fixtures. §1 is a passing test. §2 and §3 are `it.fails` — the body states the behaviour we want and pins that it does not hold yet, so the suite reports it the moment it changes. Deleting the `.fails` is the whole test-side edit once this is decided.

Verified: `biome lint .` exit 0 · `tsc --noEmit --project src/cli/tsconfig.json` exit 0 · `vitest run` — 1018 passed, 2 expected fail, 18 skipped, 61 files.

## For your decision

`root: false` is right for a nested config and wrong for a root one, and we ship one file into both positions. Three ways out, none free:

1. **Transform on copy** — strip `root` in `copyPreset` for `biome`. Smallest surface, but the #428 drift model in `src/cli/utils/copied-assets.ts` hashes the *source* and compares it to the target, so a transformed copy reads as permanently `modified`. Doing it properly means both `copyPreset` and `classifyCopiedAssets` hashing one shared "materialise" step. Existing `copy biome` consumers would then correctly show `stale` and be repaired by `fix copied-assets` — arguably the right migration signal.

2. **Rename the shipped preset** so Biome never auto-discovers it (`tooling/biome/preset.json`), and drop `root` entirely. Kills the collision at the source and removes the footgun from the published file. Invisible to `extends` consumers (they go through the `./biome` exports entry) and to `copy biome` (target name unchanged). Breaks anyone who hardcoded the raw `node_modules/.../tooling/biome/biome.json` path, and touches the exports map, the `files` list, `PRESETS.biome.source`, and doctor's path references.

3. **Revert #469** and put the lint scripts back to an explicit path list. Restores the footgun-free preset at the cost of the whole-repo `biome check .` that #469 bought, and re-opens #467.

Separately, whichever is chosen, #3 above (nested `extends` pointer) stays broken unless the pointer can learn when it is nested — a `--nested` flag on `fix biome`, or doctor detecting a parent `biome.json`. I did not want to guess at that.

I'd take **2**: it deletes the problem rather than working around it, and `root` stops being something a consumer can inherit by accident.

### DIFF
--- a/tests/cli/generators/linting.test.ts
+++ b/tests/cli/generators/linting.test.ts
@@ -1,4 +1,5 @@
 import fs from 'fs-extra'
+import { spawnSync } from 'node:child_process'
 import { join } from 'node:path'
 import { describe, expect, it } from 'vitest'
 import type { ProjectConfig } from '../../../src/cli/commands/setup.js'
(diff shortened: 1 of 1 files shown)


rtorcato (commented):

Tests only. Fixtures live in tmp dirs (`useTmpDir`), `spawnSync` calls pass argv arrays (biome binary, git init) — no shell string-building, nothing written inside the repo. No injection surface.

On the finding itself: the only demonstrated effect of the silent fallback is `noExplicitAny` reverting ON and quote style reverting to defaults — i.e. Biome's built-in fallback is *stricter* here, not weaker, and no security-specific rule was shown to be lost. This is a config-integrity/DX regression (green build, unintended ruleset) rather than a demonstrated weakening of security posture — worth fixing per the author's plan, but not on security-severity grounds. The #428 hash-drift interaction for option 1 is already flagged by the author; nothing to add there.

### Before merging
Nothing.

rtorcato (commented):

Claims check out. The probe (`noExplicitAny`, off in preset/on in Biome defaults) correctly distinguishes "preset applied" from "preset silently evaporated" — a `noDebugger` probe (recommended in both) would have masked exactly the failure being tested for, as the author notes they discovered the hard way.

Both `it.fails` bodies assert the *desired* end state (`not.toContain('noExplicitAny')`, `not.toContain('nested root configuration')`), so removing `.fails` is the whole fix-side edit. `check()` never throws (spawnSync result is just stringified), so a failure can only come from the named string appearing/missing — no risk of a false pin from an unrelated crash.

All three fixtures spawn the real `biome` binary against real temp-dir project trees (git init'd, node_modules laid out to mirror `extends` resolution), built via `useTmpDir()`, nothing committed. `Refs #486` is correct — two of three sub-claims are open defects with a real trade-off, not resolved yet.

### Before merging

Nothing.

PR #492, opened 2026-08-21
fix(doctor): name both paths in the skill-fork diff hint

### LINKED ISSUE #484: doctor's skill-fork hint says to diff but never names the paths

Surfaced by `code-reviewer` on PR #483, non-blocking, so filed rather than left as a comment that dies on merge.

`doctor`'s fork hint in `src/base/checks.ts` says *"Diff it against the shipped copy"* but never names that path. The equivalent refusal in `src/base/fixers.ts` builds a runnable `diff "<realFile>" "<shippedFile>"`.

So someone acting on `doctor` output alone has nothing to diff against — and the fork case is exactly when they most want to see what diverged before deciding.

Fix: give the `doctor` hint the same resolved paths `fixers.ts` already computes, including `realFile` when the install is a symlink into a dotfiles checkout.

## Note on the label

`ai-suggested` is a trial of the scheme proposed in #478 — an agent-surfaced follow-up gets an issue rather than a paragraph in a PR comment. Per that proposal it is a **triage queue, never `ai-ready`**: promoting it is a human act.

rtorcato:

`doctor`'s fork hint said *"Diff it against the shipped copy"* and named neither file, so anyone acting on `doctor` output alone had nothing to diff.

- `claudeSkillStatus` now carries `realFile` and `shippedFile`.
- One `skillDiffCommand` helper builds the command for both `doctor` and `fix claude-skills`, so the two can never name different files (the #475 two-copies problem).
- `realFile` is the resolved target, so a stow symlink into a dotfiles checkout sends the reader to the bytes rather than the link.
- `--skills-dir` (#490) needs nothing extra: `realFile` derives from whatever directory `resolveSkillsDir` returns.

Tests assert the exact `diff "<installed>" "<shipped>"` string in the hint, plus the symlink case (target named, link not).

`biome lint`, `tsc --noEmit`, and `vitest run` (1022 passed) are clean.

Closes #484

### DIFF
--- a/src/base/checks.ts
+++ b/src/base/checks.ts
@@ -2,7 +2,11 @@ import path from 'node:path'
 import fs from 'fs-extra'
 import { BLOCK_START, hasStaleAgentBlock } from '../cli/generators/agent-rules.js'
 import { BADGE_START, hasPublicOnlyBadges } from '../cli/generators/badges.js'
-import { claudeSkillStatus, SHIPPED_SKILL } from '../cli/generators/claude-skills.js'
+import {
+	claudeSkillStatus,
+	SHIPPED_SKILL,
+	skillDiffCommand,
+} from '../cli/generators/claude-skills.js'
 import { DEPENDABOT_CONFIG_PATHS, dependabotIgnoreRules } from '../cli/generators/security.js'
 import { type DetectedLanguage, detectNestedLanguages } from '../cli/utils/detect-language.js'
 import type { CheckResult } from './types.js'
@@ -552,7 +556,9 @@ export async function checkClaudeSkills(skillsDir?: string): Promise<CheckResult
 	// A local fork is `ok` for the same reason a modified copied asset is (#448):
 	// it is somebody's deliberate work, so it is named once and never nagged as
 	// fixable — pointing at a `fix` that would refuse is worse than saying nothing.
-	if (status.contentState && status.contentState !== 'pristine') {
+	// `realFile` is set whenever `contentState` is — both mean something is
+	// installed. The extra test is TypeScript's, not a real condition.
+	if (status.realFile && status.contentState && status.contentState !== 'pristine') {
 		const why =
 			status.contentState === 'modified'
 				? `has local changes since ${status.installedVersion}`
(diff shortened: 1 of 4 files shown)


rtorcato (commented):

`skillDiffCommand` only builds a string for `hint`/`describeSkillFound` text — no call site in this diff feeds it to `exec`/`spawn`. The quoted-path paste risk (a path with `"` or `$(...)` breaking out when a human copies the line) is the same pattern `fixers.ts` already had before this PR; this just reuses it for the `doctor` hint, not a new exposure. The #483 fork guard (`status.contentState !== 'pristine'`) and `--force-skills` requirement are unchanged — the added `status.realFile` check is a redundant TypeScript narrowing, not a logic change.

### Before merging
Nothing.

rtorcato (commented):

Checked all four points: `skillDiffCommand` is the one builder both `checks.ts` and `fixers.ts` call (no second inline string), `realFile` resolves through `isSymlink`/`fs.realpath` so a stow link names the dotfiles target rather than the link, and both new tests assert the exact `diff "<path>" "<path>"` substring plus the symlink-target case — they'd fail if the hint reverted to the old unpathed wording. The `status.realFile && status.contentState` guard is correctly redundant (both are set together in `claudeSkillStatus`), matching its own comment.

One gap: `skillDiffCommand` interpolates both paths into double quotes with no escaping. A path containing `"` or `$(...)` would break or execute when the printed command is pasted. This isn't new — it matches the pre-existing string in `fixers.ts` — but centralizing it here would be the natural place to guard it (e.g. reject/escape embedded quotes) if these paths are ever attacker- or user-influenced.

### Before merging
Nothing blocking. Optional: consider escaping/rejecting `"`/`$` in `skillDiffCommand`'s inputs since it's now the single place both call sites build this string.

rtorcato:

**Merge it.** Both reviewers clean, all checks green.

The one non-blocking note — escaping `"`/`$` in `skillDiffCommand` — is filed as https://github.com/rtorcato/repo-tooling/issues/493 rather than left here to die on merge. Worth doing because this PR made it a single helper, so hardening it once now covers both `doctor` and `fix`.

PR #494, opened 2026-08-21
docs(site): refresh the homepage CLI transcripts against the real output

### LINKED ISSUE #488: Docs homepage still shows the pre-#470 setup wizard flow

`apps/docs/src/pages/index.tsx` hand-writes a transcript of `setup` output. #481 fixed the greeting and pinned it with a test, but the **flow** below it is still pre-#470:

```
? Project type:    📚 Library/Package
? TypeScript:      ✅ Yes
...
✅ Setup complete.
```

#470 made interactive `--preset` show a reviewable file list before writing, and added a `minimal` preset. Neither appears. The homepage demo is the first thing a prospective user sees, and it shows a wizard that no longer exists.

The structural problem is that the mock is a hand-written duplicate of CLI output — which is why it and the real greeting drifted together for months, with neither catching the other.

## Options

| Option | Notes |
|---|---|
| Refresh the transcript by hand | Cheapest. Drifts again at the next wizard change. |
| Generate it from a `--dry-run` snapshot | Faithful and self-maintaining; needs the wizard to have a non-interactive transcript mode it does not have today. |
| Drop the transcript for a shorter, flow-agnostic demo | Sidesteps drift entirely by not promising a specific flow. |

`ai-suggested` rather than `ai-ready`: refreshing it by hand is agent-doable, but choosing whether to keep hand-maintaining a duplicate is not.

rtorcato:

Closes #488

Took **option 1** (refresh by hand). Option 2 needs a non-interactive transcript mode the wizard does not have; option 3 throws away the first thing a prospective user sees. Option 1 is cheap and reversible.

## Which setup flow the demo shows

**`--preset library`, not the plain interactive wizard.**

The plain wizard is now **18 prompts** (`src/cli/commands/setup.ts:328-550`) — language, project name, type, TypeScript, tsconfig flavour, linter, oxlint, test framework, git hooks, commitlint, release tool, tree-shake, publint, badges, security, AI rules, bundler, bun. Compressing that to six lines is *exactly* what produced the stale mock: a shape that looks plausible and matches nothing.

The preset path is genuinely short, and it advertises the better story: one command, a **reviewable file list before a single file is written** (#470), and a checkbox to drop the opinionated extras. That is the thing worth putting on a homepage.

**Note for reviewers:** `--preset` sets `interactive = false` (`setup.ts:187`), so the `🛠️  Welcome to repo-tooling setup!` greeting does **not** appear on this path — its absence from the page is correct, not a regression of #481. The CLI greeting is still pinned by `tests/cli/commands/setup.test.ts:631`.

## `doctor` and `fix` had drifted too

Asked to check them — they were **not** fine. Neither matched the real output at all:

| | Homepage said | CLI actually prints |
|---|---|---|
| doctor header | `⚕️  Checking project health…` | `🩺 Diagnosing <dir> against @rtorcato/repo-tooling presets...` |
| doctor rows | `✔ TypeScript      base config extends …` | `✅ TypeScript — ok` + indented gray detail on the next line |
| doctor footer | `2 warnings, 1 missing → run \`fix\`` | `Summary: N ok, N drift, N missing, N not configured` + a `Next steps:` block |
| fix header | `🔧 Applying fixes…` | `🔧 N item(s) to address` |
| fix rows | `✔ vitest          reset coverage thresholds` | `Vitest (drift) → vitest` then `✅ wrote …` |
| fix footer | `3 fixes applied.` | `Summary: N applied, N skipped, N unsupported` |

Both are rewritten. Every check name, status word, detail string, fix target and verb is taken verbatim from `src/base/checks.ts`, `src/languages/js/checks.ts`, `src/cli/commands/fix-targets.ts` and `src/base/fixers.ts`. The three tabs now describe **one coherent repo**: doctor's four findings are the same four `fix --yes` walks, `Release token` correctly has no registered fixer, and `Claude skills` is correctly `explicitOnly` so `--yes` skips it.

Where a real run prints more than fits, the elision is marked (`… and 24 more`, `… 47 more checks`) rather than silently dropped — per the issue's "truncate visibly rather than invent a shorter flow".

## For your decision

The drift-reduction I did *not* build, in rough order of cost:

1. **What I did ship (free):** a comment above `EXAMPLES` naming the three source files each tab transcribes. Costs 4 lines; makes the next CLI-output change at least *findable* from the docs side, and vice versa.
2. **A cheap test (~20 lines, no new CLI code):** assert the transcripts contain strings the CLI genuinely exports — `PRESET_NAMES`, `REVIEWABLE_FEATURES[].label`, `FIX_TARGETS` values, the `STATUS_ICONS` labels. It would not catch layout drift, but it would have caught *every* drift actually present here, because all of it was in names and labels. This is the best value-per-line on the list.
3. **`setup --transcript` (option 2 proper):** a real non-interactive replay the docs build snapshots. Correct and self-maintaining, but it means teaching the wizard to render without a TTY — a much larger change than #488 justifies.

I'd take (2) as a follow-up if you want the guarantee; (1) alone leaves this drifting again at the next wizard change, which the issue already accepts.

## Verification

- `biome check apps/docs/src/pages/index.tsx` — clean
- `biome lint .` — 62 warnings, all pre-existing CSS `!important`/a11y warnings, none in the changed file
- `vitest run` — 61 files passed, 1020 tests passed, 2 expected-fail, 18 skipped

### DIFF
--- a/apps/docs/src/pages/index.tsx
+++ b/apps/docs/src/pages/index.tsx
@@ -183,9 +183,23 @@ type Example = {
 }
 
 // Hand-transcribed from the real CLI. When you change what these commands
-// print, update the matching entry here — the source of truth for each tab is
-// src/cli/commands/{setup,doctor,fix}.ts. Lines marked `…` are truncated on
+// print, update the matching entry here. Lines marked `…` are truncated on
 // purpose (a real run prints the full list); nothing else here is invented.
+//
+// Each command file renders the framing, but most of the per-item text does
+// not live there — to check a tab against the code, start at the command and
+// follow it out:
+//   setup  — src/cli/commands/setup.ts (file-count banner, feature
+//            checkboxes, phase lines); the file list itself is
+//            computeFileList() in src/cli/commands/setup-presets.ts.
+//   doctor — src/cli/commands/doctor.ts (heading, status lines, Summary,
+//            Next steps) also owns the Release token check; every other check
+//            name/detail/hint comes from src/base/checks.ts and
+//            src/languages/js/checks.ts.
+//   fix    — src/cli/commands/fix.ts (counts, wrote/skipped lines, Summary);
+//            the `check → target` arrows come from
+//            src/cli/commands/fix-targets.ts, and which checks have a fixer
+//            or are skipped unless named explicitly from src/base/fixers.ts.
 const EXAMPLES: Example[] = [
 	{
 		label: 'setup',
--- a/apps/docs/src/pages/index.tsx
+++ b/apps/docs/src/pages/index.tsx
@@ -182,22 +182,44 @@ type Example = {
 	code: string
 }
 
+// Hand-transcribed from the real CLI. When you change what these commands
+// print, update the matching entry here — the source of truth for each tab is
+// src/cli/commands/{setup,doctor,fix}.ts. Lines marked `…` are truncated on
+// purpose (a real run prints the full list); nothing else here is invented.
 const EXAMPLES: Example[] = [
 	{
 		label: 'setup',
-		file: 'npx @rtorcato/repo-tooling setup',
+		file: 'npx @rtorcato/repo-tooling setup --preset library',
 		language: 'bash',
-		code: `$ npx @rtorcato/repo-tooling setup
+		code: `$ npx @rtorcato/repo-tooling setup --preset library
 
-🛠️  Welcome to repo-tooling setup!
-? Project type:    📚 Library/Package
-? TypeScript:      ✅ Yes
-? Linter:          ⚡ Biome
-? Test framework:  ⚡ Vitest
-? Git hooks:       ✅ Husky + lint-staged
-? Releases:        🚀 semantic-release
+📄 This preset writes 32 files:
 
-✅ Setup complete.`,
+   package.json
+   .repo-tooling.json
+   .editorconfig
+   .nvmrc
+   .gitignore
+   knip.json
+   .vscode/extensions.json
+   tsconfig.json
+   … and 24 more
+
+? 🧹 Uncheck anything you do not want:
+❯◉ Git hooks (Husky + lint-staged)
+ ◉ Conventional commit linting (commitlint)
+ ◉ Automated releases (semantic-release)
+ ◉ Security automation (Dependabot + CodeQL)
+ ◯ README status badges
+ ◯ AI agent rules (AGENTS.md, CLAUDE.md, Cursor, Copilot)
+
+🛠️  Scaffolding library in ~/my-lib
+
+📝 Generating configuration files...
+
+📦 Installing dependencies...
+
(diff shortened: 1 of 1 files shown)


rtorcato (commented):

### Before merging
Nothing.

rtorcato (commented):

Transcripts are internally consistent (file counts, drift/missing/not-configured tallies in `doctor` line up with the 4 items `fix` walks, truncation is marked). The `--preset library` call and the greeting-absence reasoning both hold up.

One real inconsistency: the new maintainer-facing comment says "the source of truth for each tab is `src/cli/commands/{setup,doctor,fix}.ts`", but the PR's own before/after table says the doctor/fix strings actually came from `src/base/checks.ts`, `src/languages/js/checks.ts`, `src/cli/commands/fix-targets.ts`, and `src/base/fixers.ts` — not the command files. That's exactly the guardrail this PR is supposed to add against future drift, so pointing it at the wrong files defeats its own purpose.

### Before merging
Fix the comment above `EXAMPLES` to cite the actual source files (checks.ts / fixers.ts / fix-targets.ts) instead of `src/cli/commands/{setup,doctor,fix}.ts`.

rtorcato:

**Sent back.** `code-reviewer` graded this non-blocking; overriding to `ai-changes` because of what the finding *is*, not its size.

The PR adds a maintainer comment above `EXAMPLES` naming where each transcript's source of truth lives — that comment is the anti-drift guardrail this PR exists to add. It cites `src/cli/commands/{setup,doctor,fix}.ts`, but the PR's own before/after table shows the doctor and fix strings come from `checks.ts`, `js/checks.ts`, `fix-targets.ts` and `fixers.ts`.

So merging ships a pointer that sends the next person to files that do not contain the strings — and they will conclude the transcripts are fine. A wrong guardrail is worse than none, because it is trusted.

One line. Everything else in the review held up: transcripts internally consistent, tallies match, truncation marked, the `--preset` choice and the greeting-absence reasoning both sound.

rtorcato (commented):

Re-reviewed after 79ab3b9: it only expands a code comment (source-file pointers above `EXAMPLES`), no transcript string, import, or rendering path touched. Verdict unchanged.

### Before merging
Nothing.

rtorcato (commented):

Transcripts are unchanged since the last pass (same tallies, same items, same truncation marks) — no need to re-litigate those.

The corrected comment's attributions check out against the diff: doctor/fix per-item text now correctly points to `checks.ts`/`js/checks.ts` (doctor) and `fix-targets.ts`/`fixers.ts` (fix) instead of the command files, matching what the prior review's before/after comparison established. The command-file vs. helper-module split is applied consistently across all three tabs, and every visible line in the transcripts (banners, checkboxes, phase lines, per-check items, summaries) traces back to something named in the comment. No orphaned output.

### Before merging
Nothing.

PR #495, opened 2026-08-21
fix(biome): rename the shipped preset so Biome never auto-discovers it

### LINKED ISSUE #486: root: false on the shipped Biome preset is untested inside a nested monorepo

rtorcato:
🤖 *Automated — implementer via `ai-issue-loop`.*

Implements **option 2** from the owner's decision on #486: rename `tooling/biome/biome.json` → `tooling/biome/preset.json` and drop `root` entirely.

Closes #486

## Why the rename does the work

`biome.json` is a name Biome's config discovery matches, so the file the package ships *was* a config wherever it sat. That is the only reason `"root": false` was ever on it (#469) — to stop the package's own copy colliding with this repo's root config. Renaming removes the collision at the source, so `root` has nothing left to do and stops being something a consumer inherits by accident.

The target name is unchanged: `copy biome` still writes the consumer's file as `biome.json`. Only the source moved.

## What changed

| Site | Change |
|---|---|
| `tooling/biome/biome.json` | → `tooling/biome/preset.json`, `"root": false` removed |
| `package.json` `files` | `tooling/biome/preset.json` |
| `package.json` `exports` | `"./biome"` → `./tooling/biome/preset.json` (**key unchanged**) |
| `src/cli/utils/copy-preset.ts` | `PRESETS.biome.source`, plus a comment saying why the name is not `biome.json` |
| `apps/docs/biome.json` | relative `extends` retargeted |
| `scripts/copy-biome-config.sh`, `scripts/sync-biome-config.sh` | paths |
| `tooling/biome/README.md` | manual-copy path, and a note that the copy must still land as `biome.json` |
| `src/languages/js/checks.ts` | two stale path references **in comments only** (kept minimal — `ai-484` is in flight on this file) |

Whole-repo grep for `tooling/biome/biome.json` is clean apart from one deliberate historical mention in a test docstring.

## Proof, not assumption

- **`extends: ["@rtorcato/repo-tooling/biome"]` still works** — new test `applies through the extends pointer, resolved via the exports map` runs a real `biome check` against a fixture package whose `exports` map is the shipped one. Mutation-checked: pointing the fixture's `./biome` at the old `biome.json` name makes it fail on `expected … to contain 'Checked'`.

  Worth flagging: **this repo's own `biome check .` proves nothing here.** `node_modules/@rtorcato/repo-tooling` symlinks back to the working tree, so it resolves whatever is checked out, not what ships. I initially took a green `biome check .` as the proof and it was not one — hence the fixture test.
- **`copy biome` still writes `biome.json`** — the existing `copy biome writes biome.json to target dir` e2e test passes against a real `pnpm build` + real CLI.
- **This repo's `biome check .` still works** — 229 files, zero errors, no nested-root abort. (#469's goal, now achieved by the rename rather than by `root: false`.)

## The `it.fails` tests from #491

| Test | Now |
|---|---|
| `applies to a nested package, overriding the monorepo root config` | **Rewritten** — see below |
| `applies when copied in as the consumer's own root config` | `.fails` **removed**, real assertion, passes |
| `lets the scaffolded extends pointer sit inside a monorepo` | `.fails` **kept** — still fails, out of scope |

### The trade this makes — please read

One file cannot be both a root and a non-root, so dropping `root` did not erase a failure, it **moved** one. The third test in that block was passing on `main` and would now fail:

- **Fixed:** `copy biome` into a standalone repo used to be discarded in silence — consumer left on stock Biome, exit 0, no warning. It now applies.
- **Traded:** that same verbatim copy landed in a package *under someone else's monorepo root* is now a second root config, and Biome aborts loudly with `Found a nested root configuration`.

I rewrote that test to assert the new behaviour rather than delete it, and documented the trade in the block's docstring. It still asserts the parent config is genuinely in force, so the result is not vacuous.

I think the trade is worth taking, because it **converges** the residual: the verbatim copy and the scaffolded `extends` pointer now fail the same way for the same reason. One fix — teaching the scaffold when it is nested — closes both, instead of two divergent behaviours needing two fixes.

## The residual is unchanged

Confirming as asked: **this does not fix finding 2.** The scaffolded `extends` pointer still reads as a root config inside someone else's monorepo, because `root` does not inherit through `extends`. That still needs a `--nested` flag or doctor detecting a parent config. Left open; the `it.fails` test still pins it.

## Bonus: a test for the `files`/`exports` gap

The brief asked whether `tooling/tests/exports-resolution` covers this. **It does not** — it cross-checks `src/` subfolders against the `exports` map and never looks at `tooling/` or at `files`.

So this rename had three places that had to move together (`PRESETS.biome.source`, the `./biome` export, the `files` entry) with **nothing failing if one was missed**: `files` is globs, so a stale exports target still resolves in-repo and only 404s once published.

New `tests/cli/utils/copy-preset.test.ts` closes that with `npm pack --dry-run --json --ignore-scripts` — the authoritative answer to "what actually ships", with no reimplementation of npm's glob semantics. Asserts every `PRESETS[*].source` and every string-valued `exports` target both exists on disk and is in the pack manifest. ~3.5s.

## For your decision

**Does this warrant `feat!:`?** I have left the commit as `fix(biome):` rather than decide alone, since the squash subject cuts the release.

The argument for `feat!:`: anyone hardcoding `node_modules/@rtorcato/repo-tooling/tooling/biome/biome.json` breaks. That path appeared in this repo's own `tooling/biome/README.md` as "Option 2: Manual Copy", so it is a documented path, not merely a theoretical one.

The argument for `fix:`: the supported entry points — `extends` via `./biome`, and `copy biome` — are both unchanged, and the user-visible effect is that a silently-ignored config now applies.

Also worth your call, and the reason I did not just pick: the **nested-copy regression** above is arguably the breaking part, more than the path rename is.

## Verification

- `biome check .` — 229 files, 0 errors
- `vitest run` — 63 files, 1042 passed, 1 expected fail (the out-of-scope residual)

One caveat, reported rather than buried: I saw a single transient run with 6 failures and an anomalous skip count, immediately after `pnpm build` regenerated `dist/`. Four subsequent full runs were clean and I could not reproduce it; the detail was lost to output truncation. I checked the obvious suspect — the new test shelling out to `npm pack` — and ruled it out: there is no `prepack` script and `--ignore-scripts` is passed, so it cannot rebuild `dist` mid-suite. Flagging it in case CI shows the same thing.

### DIFF
--- a/apps/docs/biome.json
+++ b/apps/docs/biome.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://biomejs.dev/schemas/latest/schema.json",
   "root": false,
-  "extends": ["../../tooling/biome/biome.json"],
+  "extends": ["../../tooling/biome/preset.json"],
   "files": {
     "includes": ["!**/.docusaurus", "!**/build", "!**/node_modules", "!**/docs/changelog.md"]
   }
--- a/package.json
+++ b/package.json
@@ -90,7 +90,7 @@
 		"tooling/docusaurus/docs-helpers.mjs",
 		"tooling/docusaurus/theme-tokens.css",
 		"tooling/docusaurus/theme.css",
-		"tooling/biome/biome.json",
+		"tooling/biome/preset.json",
 		"tooling/bun/bunfig.toml",
 		"tooling/nx/nx.json",
 		"tooling/changesets/config.json",
@@ -201,7 +201,7 @@
 			"types": "./tooling/docusaurus/index.d.mts",
 			"import": "./tooling/docusaurus/index.mjs"
 		},
-		"./biome": "./tooling/biome/biome.json",
+		"./biome": "./tooling/biome/preset.json",
 		"./nx": "./tooling/nx/nx.json",
 		"./changesets": "./tooling/changesets/config.json",
 		"./release-please": "./tooling/release-please/release-please-config.json",
--- a/scripts/copy-biome-config.sh
+++ b/scripts/copy-biome-config.sh
@@ -13,7 +13,7 @@ if [ ! -d "node_modules/@rtorcato/repo-tooling" ]; then
 fi
 
 # Copy the base configuration
-cp node_modules/@rtorcato/repo-tooling/tooling/biome/biome.json ./biome.json
+cp node_modules/@rtorcato/repo-tooling/tooling/biome/preset.json ./biome.json
 
 echo "✅ Biome configuration copied successfully!"
 echo ""
--- a/scripts/sync-biome-config.sh
+++ b/scripts/sync-biome-config.sh
@@ -4,11 +4,11 @@
 echo "🔄 Syncing Biome configurations..."
 
 # Copy the base config to root (removing comments for compatibility)
-jq '.' tooling/biome/biome.json > biome.json
+jq '.' tooling/biome/preset.json > biome.json
 
 # Add project-specific overrides
 jq '.files.includes += ["!**/.pnpm-store", "!**/tooling/typescript/**/*.json"]' biome.json > biome.temp.json
 mv biome.temp.json biome.json
 
-echo "✅ Biome config synced from tooling/biome/biome.json"
+echo "✅ Biome config synced from tooling/biome/preset.json"
 echo "📝 Added project-specific excludes"
--- a/src/cli/utils/copy-preset.ts
+++ b/src/cli/utils/copy-preset.ts
@@ -37,7 +37,9 @@ export interface PresetDefinition {
 
 export const PRESETS: Record<PresetName, PresetDefinition> = {
 	biome: {
-		source: 'tooling/biome/biome.json',
+		// Not named `biome.json`: Biome auto-discovers any file by that name, so
+		// shipping one inside this package made it a second root config here (#486).
+		source: 'tooling/biome/preset.json',
 		target: 'biome.json',
 		desc: 'Biome formatter and linter configuration',
 	},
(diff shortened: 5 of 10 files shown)


rtorcato (commented):

No defects found.

### Before merging

Nothing.

rtorcato (commented):

Reviewed the trade characterization, the rewritten `it.fails` tests, the new `copy-preset.test.ts`, and the extends/exports fixture test — all check out. The rewritten nested-root test isn't vacuous: it separately proves the sibling package still resolves the parent config, so the abort assertion means something. The exports-map fixture test is a real `biome check` against the shipped shape, not the self-referential `node_modules` symlink.

On the open semver question: I'd lean `feat!:`. The nested-copy trade is a real behavior change for an existing supported path (`copy biome` into a package under someone else's monorepo root), not just a doc-only path rename — that's more than a `fix:` in spirit even though the two blessed entry points (`extends`, `copy biome` into a standalone repo) are unaffected or improved.

### Before merging

Nothing.

PR #497, opened 2026-08-21
fix(doctor): shell-quote both paths in the skill-fork diff hint

### LINKED ISSUE #493: skillDiffCommand builds an unescaped shell command from user-influenced paths

Surfaced by `code-reviewer` on PR #492, non-blocking.

`skillDiffCommand` builds a `diff "<installed>" "<shipped>"` string for a human to copy and paste. Nothing executes it — `security-expert` confirmed no call site feeds it to `exec`/`spawn` — but a path containing `"` or `$(...)` produces a line that does something other than a diff when pasted.

Those paths are user-influenced: `resolveSkillsDir` honours `--skills-dir` (#490), and `realFile` comes from resolving a symlink that can point anywhere.

## Why it is worth doing now rather than before

The risk is not new — `fixers.ts` built the same string the same way before #492. What changed is that #492 factored both call sites into **one** helper. So hardening it once now fixes `doctor` and `fix` together, where previously it would have meant two fixes that could drift apart.

## Scope

Escape the paths, or refuse to emit a copyable command when they contain shell metacharacters and print them on separate lines instead. The second is arguably better: a path with a quote in it is worth *seeing* rather than silently escaping.

Small. Worth a test that a metacharacter path does not yield a runnable-but-wrong command.

rtorcato:

`skillDiffCommand` interpolated both paths into a double-quoted `diff` line for a human to paste. Both are user-influenced — `--skills-dir` is an argument (#490), `realFile` is wherever a symlink resolves — so a path carrying `"`, `$(...)` or a backtick yielded a line that fails to parse or runs something other than a diff. Nothing in this package executes the string (confirmed: no `child_process`/`execa` on either path); the hazard is the human who pastes it.

**Chose escaping over refusing to emit a command.** POSIX single-quoting is the only form correct for arbitrary bytes, and it makes a strange path *more* visible rather than less — `diff '/tmp/$(echo x)/SKILL.md' …` shows the oddity and still runs. Refusing would need a metacharacter predicate (its own bug surface, and getting it wrong is a silent miss) plus a second output shape at both call sites, reintroducing exactly the drift #492 removed. One builder, one shape, still pasteable in the ordinary case.

Tests run the emitted command through a real `sh` with `diff` swapped for a printf of its argv, and assert the two paths come back byte-identical — so a path the shell splits, expands or executes fails. Verified the three hazardous cases (`"`, `$(...)`, backtick) do fail against the old double-quoted builder.

**Swept for the same shape elsewhere:** nothing else in `src/` builds a copy-paste shell command from a filesystem path — every other interpolated command uses a module constant or a hardcoded fixer target. One adjacent case worth its own issue, not fixed here: `src/cli/generators/skills-install.ts:33` interpolates `skills/<name>` directory basenames and the `repository` owner/repo into `npx skills add …` lines written into a consuming repo's README.

`biome lint` (62 warnings, byte-identical to `origin/main`, all pre-existing CSS), `tsc --noEmit`, and the full `vitest run` (1034 passed) are clean.

Closes #493

### DIFF
--- a/src/cli/generators/claude-skills.ts
+++ b/src/cli/generators/claude-skills.ts
@@ -203,14 +203,31 @@ export interface SkillInstallResult {
 }
 
 /**
+ * POSIX single-quote escaping — the only form correct for arbitrary bytes.
+ * Inside single quotes every character is literal, so `'` is the sole one
+ * needing care: end the quote, escape it, start a new one. Double quotes are
+ * not a substitute; `$(...)`, backticks and `\` all still expand inside them.
+ */
+function shellQuote(value: string): string {
+	return `'${value.replaceAll("'", "'\\''")}'`
+}
+
+/**
  * The command a human runs to see what their fork changed, before deciding
  * whether to take the shipped copy. One builder, because `doctor` and `fix`
  * both print it and two independently-built strings drift (#484). Always
  * `realFile`: through a stow symlink the nominal path is the link, and the
  * bytes worth reading are in the dotfiles checkout it points at.
+ *
+ * Both paths are user-influenced — `--skills-dir` is an argument (#490) and
+ * `realFile` is wherever a symlink happens to point — and this line exists to
+ * be pasted into a shell, so it is quoted rather than merely interpolated
+ * (#493). Quoting beats refusing to emit a command: single quotes make an odd
+ * path *more* visible, not less, and keep the hint usable in the odd case
+ * instead of only the ordinary one.
  */
 export function skillDiffCommand(paths: { realFile: string; shippedFile: string }): string {
-	return `diff "${paths.realFile}" "${paths.shippedFile}"`
+	return `diff ${shellQuote(paths.realFile)} ${shellQuote(paths.shippedFile)}`
 }
 
 /** Whether `file` is itself a symlink, as opposed to merely resolving through one. */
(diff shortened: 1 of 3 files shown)


rtorcato:

`shellQuote` is the standard POSIX idiom (`'` → `'\''`) and is correct for every byte class in single quotes: embedded `'`, `\`, `$(...)`, backticks, `;`, adjacent/trailing quotes all produce a well-formed literal. Verified by hand against the replacement, not just the test cases shown.

One untested byte: a literal newline in a path. It survives quoting correctly for `sh -c`/scripts, but pasted into an interactive terminal without bracketed-paste, the raw `\n` submits the line before the closing quote — the shell then waits on an unterminated string rather than running the wrong command. Not exploitable, but the human-paste threat model is where this bites, and it's the one byte class the new tests don't cover.

### Before merging
Nothing.

rtorcato:

### Before merging
File the follow-up issue for `skills-install.ts:33` (same unescaped-path-into-a-copyable-command shape) — the PR description names it but nothing tracks it once this merges.

rtorcato:
🤖 *Automated — `ai-issue-loop`.*

**Ready for you to merge.** Both reviewers passed, CI is green, `mergeStateStatus: CLEAN`.

**Not merged unattended** because this repo has no `release` environment with `required_reviewers` — merging `main` fires semantic-release straight to npm with no human in between. That gate, not the review verdict, is what holds it here.

`code-reviewer` asked for a follow-up issue on the same shape in `skills-install.ts:33`. Filed as #498 (`ai-suggested`) — nothing outstanding.

One correction to the PR description while you are here: it names `owner`/`repo` as user-influenced alongside the skill name. They are not — `parseRepository` matches both against `[\w.-]+`, so a metacharacter cannot reach that template. Only the `--skill <name>` directory basename is unconstrained. #498 records it accurately.

Squash subject is `fix(doctor): …` → patch release.

Issue #499, opened 2026-08-21
Pass 3 reads the review verdict from agent return text, so a lost return wedges the PR


rtorcato:
🤖 *Filed by Claude (AI agent) on rtorcato's behalf.*

Pass 3 reads each reviewer's verdict from the agent's **final text** — the `PASS` / `PASS-NOTES` / `CHANGES` word it returns to the orchestrator. That is conversation state, and the skill's own opening rule says there should not be any:

> **All state lives in GitHub labels.** A tick is a stateless, idempotent pass over that state, so a missed tick, a crash, or a restart costs nothing. Never keep pipeline state in the conversation.

## Observed

On #497 both reviewers posted good reviews and then went idle **without returning a verdict word** — twice each, no content either time. The reviews were on the PR; the verdict was gone.

It was recoverable only because the comment carries a `### Before merging` section, so the verdict could be read off the artifact instead. Had Pass 3 depended strictly on the return value, #497 would have sat on `ai-review` indefinitely with two complete reviews attached and nothing to act on — and no tick would ever resolve it, because a later tick re-spawns the reviewers rather than reading what the previous ones already wrote.

That failure is invisible in exactly the way the label design exists to prevent: the PR looks mid-review, which is indistinguishable from actually being mid-review.

## The fix, and the part of it that does not work yet

Read the verdict from the posted comment. But `### Before merging` alone cannot carry it:

| Section | Verdict |
|---|---|
| `Nothing.` | `PASS` — unambiguous |
| anything else | `PASS-NOTES` **or** `CHANGES` — indistinguishable |

A blocking defect and a read-this-first note both produce a non-empty section. So the comment has to say which, structurally.

**Use a hidden marker**, the trick Pass 1 already uses for its decision comment:

```
<!-- ai-issue-loop:verdict:CHANGES -->
🤖 *Automated review — `code-reviewer` via ai-issue-loop.*
```

Invisible in the rendered timeline, greppable via the API, and it makes each review comment fully self-describing. Then:

```bash
gh api "repos/$OWNER_REPO/issues/<N>/comments" \
  --jq '[.[] | select(.body | test("<!-- ai-issue-loop:verdict:"))] | .[-1].body'
```

## What this buys beyond not losing verdicts

- **Pass 3 becomes idempotent per reviewer, not per tick.** A reviewer that already posted is skipped; only a missing one is spawned. Today a re-run re-spawns both, which costs tokens against the monthly cap and can post duplicate reviews.
[...]

PR #500, opened 2026-08-21
fix(skills): guard the main checkout against going bare

### LINKED ISSUE #496: Main checkout intermittently goes bare (core.bare=true) during agent worktree runs

## What happened

After Pass 2 removed the last agent worktree, the **main checkout had `core.bare = true`**. Every ordinary git command then failed:

```console
$ git status
fatal: this operation must be run in a work tree

$ git worktree list
/Volumes/RtorcatoSSD/WORK/repo-tooling  (bare)
```

It reads like the repo has been destroyed. It has not.

## Recovery — one key

```bash
git config core.bare false
```

Everything was intact underneath: working tree files present, `.git` a normal directory, tree clean, no uncommitted work, `node_modules/.bin` shims all resolving. Only that one config key was wrong.

## Cause: undetermined, and the evidence is gone

Being explicit rather than guessing — the fix overwrote the state that would have identified it. What was checked *after* recovery:

- `git --version` → **2.55.0**
- no `extensions.worktreeConfig`
- no `.git/config.worktree`
- `.git/config` otherwise normal

The sequence that preceded it was three `git worktree remove --force` calls in a loop, each followed by `git branch -D`, ending with **zero** worktrees left. Whether removing the last one is load-bearing, or whether an earlier command in the session set it and nothing noticed until the worktree list was empty, is not established.

**Do not "fix" this by guessing at a cause.** The useful change is detection, below, which is worth doing whether or not the cause is ever pinned down.

## Proposed guard

Pass 2 already ends with a repair step for a related silent breakage — `node_modules/.bin` shims left pointing into a removed worktree. This belongs in the same place and is cheaper:

```bash
git -C "$ROOT" rev-parse --is-inside-work-tree >/dev/null 2>&1 || {
  echo "main checkout went bare after worktree removal — repairing"
  git -C "$ROOT" config core.bare false
}
```

Rationale for repairing rather than only reporting: the state is unambiguous — a main checkout with files on disk is not meant to be bare — and the repair is a single reversible config key with no data at risk.

If it recurs after the guard lands, the guard's own log line is the signal that finally pins the cause, since it records *when* it happened relative to the removals.

## Why it matters more than it looks

Nothing in the loop notices. No pass runs git in the main checkout after cleanup, so this surfaces later — in the human's next `git status` or `git pull` — with no visible connection to the loop that caused it. Same failure shape as the `node_modules` one already documented in Pass 2, which is the precedent for handling it there.

## Scope

`skills/ai-issue-loop/SKILL.md`, Pass 2, alongside the existing `node_modules` rebuild. Docs-only change to a shipped file.

rtorcato:

Adds a `core.bare` detect-and-repair guard to `skills/ai-issue-loop/SKILL.md`, in **Pass 0 as well as Pass 2**.

```bash
if [ "$(git -C "$ROOT" rev-parse --is-inside-work-tree 2>/dev/null)" != true ] && [ -d "$ROOT/.git" ]; then
```

**Fix round 1 — both reviewers were right, the first probe was dead code.** `rev-parse --is-inside-work-tree` exits `0` whether or not the checkout is bare; it only *prints* `false`. It now tests stdout, and requires `.git` to be a directory before repairing — a genuinely bare repo prints `false` too, and this file ships to npm into arbitrary users' `~/.claude/skills/`, so repairing one would be a real mis-repair.

**Verified in scratch repos on git 2.55.0**, all four states:

| state | probe |
|---|---|
| healthy checkout | skips |
| linked worktree (`.git` is a file) | skips |
| genuine `git init --bare` | skips |
| wrongly bare (`core.bare true` on a normal layout) | **fires**, and `config core.bare false` restores it |

Confirmed alongside it that `rev-parse --is-inside-work-tree` printed `false` at exit `0` while `git status` exited `128` — the exit code the original snippet relied on never existed.

**Also addressed:** the Pass 2 claim that it is "the last thing to touch the main checkout" was wrong — Pass 4 runs `git worktree add` there. Reworded to the reason that actually holds: Pass 2 does the most main-checkout work and runs immediately before Pass 4 branches new worktrees off it.

Other properties, unchanged from round 0:

- **Detection and repair only.** The trigger is unidentified and is *not* `git worktree remove` — plain `worktree add`/`remove` and a `git config` write from inside a worktree all fail to reproduce it. No cause claim in the file.
- The prose now records that it is **frequent, not a one-off**: three times in one tick, twice with implementers active.
- **The log line keeps its UTC timestamp**, the only instrumentation likely to pin the trigger.
- **The repair fails loudly.** A restrictive sandbox refuses the `.git/config` write (`could not lock config file .git/config: Operation not permitted`); the guard aborts rather than reporting healthy over a broken repo.

Docs-only change to a shipped copy asset. Scoped to Pass 0 and Pass 2 to stay clear of #499, which is editing Pass 3 of the same file.

Closes #496


rtorcato:

The probe never fires for the bug it targets: `git rev-parse --is-inside-work-tree` exits `0` even when `core.bare=true` is wrongly set on a non-bare layout — it only prints `false` to stdout, which `>/dev/null` discards. Verified locally (git 2.55): same misconfig, exit code `0`. The `||` repair branch is dead code.

A probe that does fail loudly here, e.g. `git status` / `rev-parse --show-toplevel` (both exit `128`, `fatal: this operation must be run in a work tree`), fails **identically** inside a genuine bare repo — also verified locally. Neither probe distinguishes "main checkout wrongly bare" from "this is legitimately bare," so any fix still risks `git config core.bare false` against a real bare clone for whoever installs this skill.

### Before merging
Swap the detection for something that (a) actually distinguishes the failure described in #496 from a healthy repo, and (b) checks for a `.git` subdirectory (non-bare layout) before concluding the bareness is a bug rather than the repo's real shape — otherwise this repairs a category it can't detect and mis-repairs one it can't distinguish.

rtorcato:

**The guard never fires.** `git rev-parse --is-inside-work-tree` prints `true`/`false` to stdout but always exits `0` — the `||` never triggers. Verified live: this checkout currently has `core.bare=true`, and `--is-inside-work-tree` exits `0` printing `false`, while `--show-toplevel` exits 128. Detection needs to test stdout, e.g. `[ "$(git -C "$ROOT" rev-parse --is-inside-work-tree 2>/dev/null)" = true ]`, or use `--show-toplevel` instead.

### Before merging
- Fix the detection command (above) — as written, this guard is dead code and will never repair anything.
- The Pass 2 claim "the last thing to touch the main checkout in a tick" isn't verifiable from this diff alone; confirm nothing later in the file (Pass 3/4) touches the main checkout before merging.

rtorcato:
🤖 *Automated — `ai-issue-loop` Pass 3.*

**Closing without merging.** The branch's second commit (`ee98321`) records deleting all 359 files in the repo. Rather than rewrite published history, #496 goes back in the queue and gets reimplemented from a clean worktree.

Nothing was lost: `main` never moved, and the deletions were only ever recorded on this branch.

## What happened, because it matters more than this PR

The commit is an artefact of **the very bug this PR was fixing**. During the fix round the main checkout was `core.bare = true`, and the worktree's index was wiped as a result — all 360 files went from tracked to untracked while sitting untouched on disk. The commit then faithfully recorded them as deleted.

Confirmed after the fact: every file was still present in the worktree, and a plain `git reset` to the previous commit produced a clean tree with a single modified file.

**So `core.bare` is not cosmetic.** #496 described it as a repo that *reads* as destroyed while being fine. That was too generous — it can silently corrupt an agent's commits. That raises the value of the guard considerably, and it is worth saying in the file.

## The review findings were correct and are already resolved

Both reviewers independently found that the original probe was dead code: `git rev-parse --is-inside-work-tree` exits `0` regardless and only *prints* the answer, so testing its exit status never fires. That defect came from the snippet in #496 itself, not from the implementer.

Verified on git 2.55.0 across all four states:

| repo state | `--is-inside-work-tree` | `git status` | `--git-dir` | `.git` subdir |
|---|---|---|---|---|
| healthy checkout | `true`, exit 0 | 0 | `.git` | yes |
| **wrongly bare** | `false`, **exit 0** | 128 | `.git` | **yes** |
| genuinely bare | `false`, exit 0 | 128 | `.` | **no** |

Two conclusions carry to the reimplementation:

1. **Test stdout, not the exit code.**
2. **`.git` must be a directory before repairing.** Every exit-code probe fails identically inside a genuinely bare repo, so nothing else separates the two. This file ships to users' `~/.claude/skills/`, where "repairing" someone's real bare clone would be the damage rather than the fix.

Also unresolved from `code-reviewer`: the Pass 2 claim that it is "the last thing to touch the main checkout in a tick" could not be verified from the diff, and should be substantiated or softened.

Refs #496

PR #502, opened 2026-08-21
fix(skills-install): quote the skill name in the README install commands

### LINKED ISSUE #498: npx skills add lines in README interpolate an unquoted skill directory name

Surfaced by the implementer of #493 while sweeping for the same shape elsewhere. Same class as #493, lower stakes — filed rather than folded into that PR, because it is a different file with a different threat model.

## What

`buildSkillsInstallBody` (`src/cli/generators/skills-install.ts:33`) writes copy-paste shell commands into a consuming repo's README:

```ts
.map((s) => `npx skills add https://github.com/${owner}/${repo} --skill ${s}`)
```

`s` is a directory basename from `findSkills`, which reads whatever is under `skills/`. Nothing constrains it. A directory named `x; echo pwned` yields a README line that runs two commands when pasted.

## Narrower than it first looks

The implementer's note named `owner`/`repo` as well. They are **not** a hazard: `parseRepository` (`src/cli/generators/badges.ts:36,40`) matches them against `[\w.-]+`, so a value that reaches this template cannot carry a metacharacter. Only `${s}` is unconstrained.

## Why it is weaker than #493

The directory name is created by whoever runs the tooling, on their own repo — self-inflicted, unlike #493's `realFile`, which comes from resolving a symlink that can point anywhere. There is no crossing of a trust boundary here.

What makes it worth fixing anyway is the *destination*: unlike #493's transient hint, this string is **committed to a README** and rendered on GitHub and npm. It outlives the machine that generated it and gets pasted by people who did not create the directory.

## Second, unrelated effect

The body is emitted inside a ```` ```bash ```` fence. A backtick in a directory name does not execute anything, but it can break the fence and mangle the rendered README. Quoting fixes both.

## Scope

Reuse the `shellQuote` helper #497 added to `src/cli/generators/claude-skills.ts` rather than writing a second one — two independently-built escapers drift, which is the same argument #492 made for one `skillDiffCommand`. That means exporting it, or moving it somewhere both can import.

Small. Worth a test that a skill directory named with a metacharacter does not yield a runnable-but-wrong command.

## Related

- #493 / #497 — the same shape, in the doctor hint
- #492 — one builder, because two drift

rtorcato:

`buildSkillsInstallBody` interpolated a `skills/<name>/` directory basename straight into `npx skills add … --skill ${s}`. Nothing constrains that basename — `findSkills` returns whatever is on disk — so a directory named `x; echo pwned` produced a README line that runs two commands when pasted. `owner`/`repo` are left alone: `parseRepository` already matches them against `[\w.-]+`.

The stakes are the destination, not the source. The directory is created by whoever runs the tooling, on their own repo, so no trust boundary is crossed — but the string is committed to a README and rendered on GitHub and npm, where it outlives the machine that generated it and gets pasted by people who did not create the directory.

## Where `shellQuote` lives

**Moved to `src/cli/utils/shell.ts`** rather than exported from `claude-skills.ts`. Both files are generators; importing one from the other would couple the README writer to the user-global `~/.claude` installer for a string utility that belongs to neither. `utils/` is where this package already keeps cross-generator helpers, and `claude-skills.ts` imports from it already (`utils/copy-preset.js`). One escaper, because two drift (#492).

## The fence, separately

Quoting alone does **not** fix the second effect the issue names. A backtick stays a literal backtick inside single quotes, and a basename may legally contain a newline *and* backticks — which puts arbitrary text at the start of a line inside the block and closes a 3-backtick fence early. Per CommonMark a fence is closed only by a run at least as long, so the opener is now `max(3, longest backtick run + 1)`. Ordinary repos are unaffected: the fence stays ```` ``` ````.

## Tests

Each emitted line is run through a real `sh` with `npx` swapped for a printf of its argv (#497's technique) and asserted to hand the name back as one byte-identical argument — `;`, spaces, `$(...)`, backticks, `"` and `'` — plus a fence case for a newline-and-backticks name.

**Verified they fail without the fix.** Reverting the interpolation and the fence to their previous form turned all 8 new/updated assertions red: `x; echo pwned` split into an extra argv entry, `$(echo pwned)` was substituted away, and the quote cases aborted `sh` with a parse error. Restored, all 1041 tests pass.

This repo's own generated README block is updated to match what the generator now emits, so the next `fix ai` is a no-op.

`biome check` (62 warnings, all pre-existing CSS, byte-identical to `origin/main`), `tsc --noEmit` and the full `vitest run` are clean.

Closes #498

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,8 +234,8 @@ MIT — see [LICENSE](LICENSE).
 Any agent that supports the [`skills`](https://www.npmjs.com/package/skills) CLI can install this repo's skills straight from GitHub — no clone, no package install:
 
 ```bash
-npx skills add https://github.com/rtorcato/repo-tooling --skill ai-issue-loop
-npx skills add https://github.com/rtorcato/repo-tooling --skill npm-publish
-npx skills add https://github.com/rtorcato/repo-tooling --skill repo-tooling
+npx skills add https://github.com/rtorcato/repo-tooling --skill 'ai-issue-loop'
+npx skills add https://github.com/rtorcato/repo-tooling --skill 'npm-publish'
+npx skills add https://github.com/rtorcato/repo-tooling --skill 'repo-tooling'
 ```
 <!-- js-tooling:skills:end -->
--- a/src/cli/generators/claude-skills.ts
+++ b/src/cli/generators/claude-skills.ts
@@ -10,6 +10,7 @@ import os from 'node:os'
 import path from 'node:path'
 import fs from 'fs-extra'
 import { getPackageRoot } from '../utils/copy-preset.js'
+import { shellQuote } from '../utils/shell.js'
 
 /** Skills this package owns the content of and keeps up to date. */
 export const SHIPPED_SKILL = 'ai-issue-loop'
@@ -200,16 +201,6 @@ export interface SkillInstallResult {
 	shippedFile: string
 	installedVersion: string | null
 	shippedVersion: string
-}
-
-/**
- * POSIX single-quote escaping — the only form correct for arbitrary bytes.
- * Inside single quotes every character is literal, so `'` is the sole one
- * needing care: end the quote, escape it, start a new one. Double quotes are
- * not a substitute; `$(...)`, backticks and `\` all still expand inside them.
- */
-function shellQuote(value: string): string {
-	return `'${value.replaceAll("'", "'\\''")}'`
 }
 
 /**
--- a/src/cli/generators/skills-install.ts
+++ b/src/cli/generators/skills-install.ts
@@ -4,6 +4,7 @@
 
 import path from 'node:path'
 import fs from 'fs-extra'
+import { shellQuote } from '../utils/shell.js'
 import { upsertBlock } from './agent-rules.js'
 import { parseRepository } from './badges.js'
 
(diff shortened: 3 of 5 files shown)


rtorcato:

fenceFor is correct: `Math.max(0, ...[])` guards the empty-match case, and since opener/closer both read `fenceFor(commands)` off the same value, they can't diverge. The extracted `shellQuote`'s `../utils/shell.js` import resolves in both callers.

The test's `commandLines()` isolates only the *first* physical line of a command — a skill name with a raw embedded newline (quoting doesn't escape newlines) would have its continuation silently excluded from `rest`, since it won't match the filtered prefix.

### Before merging
Optional: harden `commandLines` to consume through the next `npx skills add ` boundary (not one line), so a future newline-bearing test case can't quietly narrow its own assertion.

rtorcato:

`shellQuote` is the standard correct POSIX single-quote form for arbitrary bytes, and the tests prove it by round-tripping through a real `sh` rather than hand-tracing. `fenceFor` measures the longest backtick run in the *already-quoted* `commands` string and always adds one, so no run inside the fenced content can ever reach fence length — CommonMark's closing-fence rule can't fire early, including for names combining backticks with a newline.

### Before merging

Residual, not blocking: a skill name with an embedded newline (as in the new fence test) makes the rendered code block span multiple *visual* lines for what is really one shell argument. A human who copies only the first visible line — rather than the whole block — pastes a command with an unterminated `'`, leaving their shell in an open-quote continuation state that swallows whatever they type or paste next. Quoting and fence-widening both fix the raw text; neither fixes this copy-behavior gap. Worth a one-line README caveat or accepting as a known limitation, since it only bites when the directory name is itself adversarial (self-inflicted).

Separately, `shell.ts`'s generic name/location invites reuse in contexts where single-quoting isn't the right escape (Windows shells, double-quoted contexts) — a doc comment noting "POSIX sh only" would head that off.

rtorcato:

**Ready to merge.** Both reviewers passed, CI green.

Neither reviewer found anything that blocks. Their three non-blocking findings are now **issues, not homework** — they survive this merge instead of dying with the PR:

- **#503** — the test helper `commandLines()` isolates only a command's first physical line, so a future newline-bearing case could quietly narrow its own assertion; plus `shell.ts` wants a "POSIX `sh` only" doc comment before something reaches for it in a Windows or double-quoted context.
- **#504** — a newline in a skill name renders across several *visual* lines, so copying just the first one pastes an unterminated `'`. Correct as text, wrong as a copy target.

Both are labelled `ai-suggested` and are not queued.

`ai-notes` dropped: there is nothing left to read before merging.

**Not merged unattended** because this repo has no `release` environment with `required_reviewers` — merging `main` fires semantic-release straight to npm with no human in between. That gate, not the review verdict, is what hands this to you.

Squash subject is `fix(skills-install): …` → patch release.

Issue #508, opened 2026-08-21
Bare-checkout guard's exit 1 doesn't tell the agent to halt the tick


rtorcato:
🤖 *Filed by Claude (AI agent) on rtorcato's behalf.*

Found by `code-reviewer` on #505 — **but the review landed after the merge**, so this shipped rather than being fixed in the PR.

## What

Both copies of the bare-checkout guard end with:

```bash
git -C "$ROOT" config core.bare false || {
  echo "⚠ repair FAILED — main checkout still bare"; exit 1; }
```

`SKILL.md` is documentation an agent follows, not a script that runs. So `exit 1` terminates only the single Bash call the agent happens to make — **nothing in the file says what the agent does next**.

An agent that does not infer "halt the tick" carries on into Pass 1 and Pass 4 with `ROOT` still bare. That is precisely the state the guard exists to prevent, and Pass 4 is the dangerous one: it runs `git -C "$ROOT" worktree add` for every issue it picks up, branching new worktrees off a broken checkout.

Observed consequence of working in a bare-parent worktree, from the attempt this guard replaces: the index gets wiped while files sit untouched on disk, and the next commit records the whole repository as deleted. PR #500 died that way — `0 additions, 67703 deletions` across 359 files.

## Fix

Say in prose what the `exit 1` means: **the tick halts**. Do not run the remaining passes; report and stop. The snippet alone cannot carry that, because an agent reading a doc is not a shell honouring an exit code.

Worth stating once next to the Pass 0 copy and referring to it from Pass 2, rather than twice — the two snippets already duplicate, and prose duplicated in two places drifts.

## Also noted on #505, and already spent

The same review flagged the squash subject as `feat(skills)` where `fix(skills)` matched the substance — the change closes the corruption bug from #496/#500 and adds no loop capability. #500 had classified it as `fix`.

That is **not actionable now**: the commit is on `main`, so semantic-release will cut a minor. Recorded here only so the version bump has an explanation, and as a reminder that the squash subject is a release decision that a post-merge review cannot take back.

## Related

- #496 / #505 — the guard
- #507 — `GIT_DIR`/`GIT_WORK_TREE` overriding `-C` in the same snippets

rtorcato:
🤖 *Automated — `ai-issue-loop` Pass 4.*

[...]

Issue #506, opened 2026-08-21
Narrow the verdict-marker author gate to OWNER


rtorcato:
🤖 *Filed by Claude (AI agent) on rtorcato's behalf.*

Surfaced by `security-expert` reviewing #501, passed rather than blocked.

## What

#501 gates verdict-marker adoption on the review author's `author_association` being `OWNER`, `MEMBER` or `COLLABORATOR` — mirroring the set Pass 4 applies to issue authors. That closes the hole it was written for: `author_association` is computed server-side by GitHub from real repository permissions and cannot be set by the poster, so an outside contributor genuinely cannot forge any of the three.

**But the set is wider than this loop's trust model.** Every agent in the pipeline authenticates as the repo owner's own `gh` — one identity, always `OWNER`. Confirmed on this very PR: the agent-posted review carries `author_association: OWNER`. So `MEMBER` and `COLLABORATOR` admit posters that are never legitimate here, and admit them to the one signal that decides a verdict.

Combined with the read taking `last`, a collaborator or org member with write but not owner access could post a later review carrying `<!-- ai-issue-loop:verdict:sec:PASS -->` and have it override a genuine `CHANGES`.

## Why it was not blocking

Exploiting it requires existing write access to the repo — someone already trusted enough to push. And an issue PR still stops for a human at Pass 1 on any repo without a `release` environment carrying `required_reviewers`, so the forged verdict changes a label rather than landing code.

It bites on the `release`-gated repo, where a fully-passed PR may auto-merge. Today that is only `js-common`.

## The fix

Narrow to `author_association == "OWNER"`.

Worth noting this is a place where **copying Pass 4's pattern was slightly wrong**, and the divergence is worth a sentence in the file. Pass 4's wider set is right there because `ai-ready` is the hard gate and the association check is a backstop — a human applies the label, and on a public repo only collaborators can. Here there is no label; the marker is the only signal, so the association check *is* the gate and should be exactly as narrow as the trust model.

## Related

- #499 / #501 — the verdict marker this gates

rtorcato:
🤖 *Automated — `ai-issue-loop` Pass 4.*

**Already done — closing.** Folded into #510 and on `origin/main`; verified in `skills/ai-issue-loop/SKILL.md`. The PR said "folds in #506" rather than `Closes`, so it never auto-closed.

[...]

PR #510, opened 2026-08-21
fix(skills): reconcile the ai-issue-loop fork — restore the five running-copy blocks

### LINKED ISSUE #509: Reconcile the ai-issue-loop skill fork — union the shipped and running copies

## The fork

`skills/ai-issue-loop/SKILL.md` (shipped) and the running `~/.claude/skills/` copy (stow-symlinked into dotfiles, hand-edited) diverged — 840 diff lines. #480 shipped the *detection* (stale vs fork) but the content was never reconciled. Consequences this week: reviewer briefs written from the stale copy missed the `ai-suggested` rule, and five merged improvements (#479, #489, #501, #505) never reached the loop that actually runs.

## What lands here — the union

Five blocks existed **only** in the running copy and are grafted back:

1. Pass 1 decision-comment upsert (`<!-- ai-issue-loop:decision -->` marker, `// empty` guard)
2. Pass 1 `release`-environment auto-merge gate (probe, fail-closed, residual-risk note)
3. Pass 2 `node_modules` rebuild after worktree removal (`REMOVED`/`LIVE` gating, `⚠rebuild`)
4. Pass 4 dependency choice: symlink for app-only issues vs real install for workspace-package work (api-common #201 lesson)
5. Pass 4 decline-comment `## To lift this hold` shape

Also folded in, being one-line edits to the same file:

- #508 — prose stating a failed bare-checkout repair **halts the tick**, not just the command
- #506 — verdict-marker gate narrowed from `OWNER/MEMBER/COLLABORATOR` to the loop's own login (login, not association: an org-owned repo never yields `OWNER`)

After merge: `repo-tooling fix claude-skills` writes the shipped copy through the stow symlink, ending the hand-edited fork. The dotfiles copy becomes an installed artifact.

Closes with the PR.

rtorcato:

Closes #509. Also folds in #506 and #508 as one-line edits to the same file.

The union merge of `skills/ai-issue-loop/SKILL.md` and the hand-edited running copy — see #509 for the block-by-block list. One deliberate deviation from #506 as filed: the verdict gate is the loop's **login** (`gh api user`), not `author_association==OWNER`, because an org-owned repo never yields `OWNER` even for its admins — the association gate would silently never adopt a verdict there.

After merge, `repo-tooling fix claude-skills` overwrites the running copy through the stow symlink, ending the fork.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -9,7 +9,7 @@
     {
       "name": "repo-tooling",
       "source": "./",
-      "description": "Drive the @rtorcato/repo-tooling CLI (setup/doctor/fix) and follow the shared release rules. Three skills: repo-tooling (adopt/audit the presets), npm-publish (never hand-cut a release) and ai-issue-loop (the label-driven issue → PR pipeline)."
+      "description": "Drive the @rtorcato/repo-tooling CLI (setup/doctor/fix) and follow the shared release rules. Six skills: repo-tooling (adopt/audit the presets), npm-publish (never hand-cut a release), ai-issue-loop (the label-driven issue → PR pipeline), ai-workflow (burst the ai-ready queue in parallel worktrees), ai-issue (file agent-executable issues) and ai-loop-status (read-only pipeline status)."
     }
   ]
 }
--- a/README.md
+++ b/README.md
@@ -180,20 +180,22 @@ ln -sf ../../node_modules/@rtorcato/repo-tooling/tooling/claude/repo-tooling.md 
 ### Use with Claude Code (plugin)
 
 This repo is also a self-hosted Claude Code marketplace. Install the plugin to
-get three skills — `repo-tooling` (adopt/audit the presets via the CLI),
-`npm-publish` (never hand-cut a release) and `ai-issue-loop` (the label-driven
-issue → PR pipeline) — in any session:
+get six skills — `repo-tooling` (adopt/audit the presets via the CLI),
+`npm-publish` (never hand-cut a release), `ai-issue-loop` (the label-driven
+issue → PR pipeline), `ai-workflow` (burst the `ai-ready` queue in parallel
+worktrees), `ai-issue` (file agent-executable issues) and `ai-loop-status`
+(read-only pipeline status) — in any session:
 
 ```
 /plugin marketplace add rtorcato/repo-tooling
 /plugin install repo-tooling@repo-tooling
 ```
 
-`ai-issue-loop` also installs on its own, user-globally, so every repo on the
-machine shares one copy:
+The four `ai-*` skills also install on their own, user-globally, so every repo
+on the machine shares one copy:
 
 ```bash
-npx @rtorcato/repo-tooling fix claude-skills   # → ~/.claude/skills/ai-issue-loop/SKILL.md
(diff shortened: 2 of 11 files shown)


rtorcato (commented):

The union reads coherently and the Pass 1 gate matches the stated policy: two reviews first, `ai-notes` suppresses, required checks stay the real merge gate, Pass 0's disarm still applies. Four one-line inconsistencies in the grafted blocks — none worth a fix round, all worth fixing at merge.

### Before merging

- **`$REMOVED` is set on one path only.** The diff adds `REMOVED=1` to the merged-PR cleanup block, but the bullet below claims it is "set by every removal path above, merged-PR cleanup *and* stall reaping" — and calls the reaped worktree the one that "needs this most". If stall reaping issues its own `worktree remove` rather than reusing that block, the rebuild silently never runs on exactly the highest-risk path, which is the failure mode the whole section exists to end.
- **`holding` is not one of the loop's labels.** `"say which in a comment, then swap `holding` for `ai-ready`"` names a label absent from the state table, inside the one section whose entire job is stating the label move unambiguously. The surrounding decline text otherwise only drops `ai-ready` and explicitly rules out `ai-blocked`.
- **Two spellings of the same pnpm flag.** Pass 2's rebuild uses `--config.confirmModulesPurge=false`; the Pass 4 prose uses `--config.confirm-modules-purge=false`. The Pass 2 text says a human hitting this "needs this exact command", so pick one and use it in both places.
- **A sentence left stale by the new gate.** "An issue PR ends at *assigned to you* and waits there" now sits immediately under the newly inserted exception saying a fully-passed issue PR may auto-merge on a `release`-gated repo. One qualifier resolves it.

Nit, not blocking: the marker-upsert `--jq` uses a bare `.body | startswith(...)` while the Pass 3 read two sections later is careful to write `(.body // "")` for the same class of hazard.

rtorcato (commented):

Narrowing the verdict gate to `gh api user` is strictly safer than the association set — a stranger cannot forge that login, and a failed `gh api user` empties `$ME`, matching nothing and re-spawning rather than adopting. The auto-merge arm is collaborator-gated in practice: `ai-ok-*` labels are the trigger, and only collaborators can label on a public repo.

Follow-up: #511

### Before merging

The `release`-environment probe tests that the environment **exists** with `required_reviewers` — not that the publishing job actually declares `environment: release`. A repo carrying that environment while its release workflow never references it reads as gated, and the arm then auto-merges a PR that does publish to npm unattended. The probe is fail-closed against absence, not against misconfiguration. Confirm the wiring on each repo you enable this for (per CLAUDE.md, today only `js-common`), or have the arm check the workflow rather than the environment list.

rtorcato:
🤖 *Automated — orchestrator via ai-issue-loop.*

**Ready to merge** — both reviewers PASS-NOTES; their notes are addressed in e84fc91 rather than left as merge homework:

- `REMOVED=1` now set on every removal path (reap + orphan rows)
- pnpm flag spelling unified to `--config.confirmModulesPurge=false`
- "assigned to you" sentence qualified with "on an ungated repo"
- upsert jq made null-safe (`(.body // "")`)
- security note: the auto-merge arm now also greps the workflows for `environment: release` — environment-exists alone no longer reads as gated

One finding did not hold: `holding` **is** in the label table (line 85), so no change there.

Squash subject `fix(skills): …` → patch release.

PR #526, opened 2026-08-26
fix(skills): gate the ai-issue-loop decision-comment upsert on the loop's own login

### LINKED ISSUE #511: Gate the ai-issue-loop decision-comment upsert on the loop's own login
🤖 *Automated — `security-expert` via ai-issue-loop.*

Surfaced reviewing #510.

**What.** The Pass 1 decision-comment upsert in `skills/ai-issue-loop/SKILL.md` selects the comment to `PATCH` by marker prefix alone:

```
[.[] | select(.body | startswith("$MARKER"))] | .[0].id // empty
```

**Why it matters.** On a public repo anyone can comment. An outsider who posts a comment beginning with `<!-- ai-issue-loop:decision -->` before the loop does is picked by `.[0]` (oldest match) forever after, and the token has repo-write, so the `PATCH` succeeds — the loop's decision text lands inside a stranger-authored comment and the loop never posts one of its own. Same class of gap the Pass 3 verdict read closes with its author gate; the upsert just did not get it. Separately `.body` is unguarded, so a null body throws and aborts the filter, emptying `ID` and re-entering the duplicate-comment branch the section exists to prevent — Pass 3 already documents `(.body // "")` as load-bearing for exactly this.

**Fix sketch.** Reuse the Pass 3 identity: `ME=$(gh api user --jq .login)`, then `select(.user.login==$me and ((.body // "") | startswith($marker)))`, passing both via `--arg`.

rtorcato:
🤖 *Automated — implemented by an agent via ai-issue-loop.*

## What

Pass 1's decision-comment upsert in `skills/ai-issue-loop/SKILL.md` picked the comment to `PATCH` by marker prefix alone. On a public repo anyone can comment, so a stranger who posts a body starting with `<!-- ai-issue-loop:decision -->` before the loop does is selected by `.[0]` (the *oldest* match) forever after — and the token has repo-write, so the `PATCH` lands. Every decision the loop ever reaches would then be written inside a stranger-authored comment, and the loop would never post one of its own.

Fixed by reusing Pass 3's identity gate:

```bash
ME=$(gh api user --jq .login)
ID=$(gh api "repos/$OWNER_REPO/issues/<N>/comments" \
  | jq -r --arg me "$ME" --arg marker "$MARKER" \
      '[.[] | select(.user.login == $me and ((.body // "") | startswith($marker)))]
       | .[0].id // empty')
```

Both values now arrive as jq `--arg` **data** instead of being interpolated into the filter text, which also retires the nested `\"` escaping the old one-liner needed.

## Same flaw, one sibling

Root cause is "select comments by body prefix with no author gate", and the triage duplicate check (Pass 4) had it too — inverted: ungated, a stranger opening with the `🤖 *Automated — triage` header **suppresses** the decline comment, so an issue gets labelled with nothing on the timeline saying why. Gated on the same login. `.author.login` there, not `.user.login` — `gh issue view --json` is GraphQL and names the field differently from the REST payload the upsert reads.

## Notes

- `(.body // "")` was already present on the upsert; the prose now says why it is load-bearing, matching Pass 3.
- Docs-only (a skill markdown file), so `fix:` cuts a patch release with no code change — consistent with how the other `skills/` fixes in this repo are typed.
- Both jq filters were run against fixtures covering: stranger-first comment, null body, own comment present, own comment absent. Selection returns the loop's own comment, survives the null, and yields empty when the loop has not posted.

Closes #511

### DIFF
--- a/skills/ai-issue-loop/SKILL.md
+++ b/skills/ai-issue-loop/SKILL.md
@@ -409,8 +409,11 @@ it behind a hidden marker and upsert:
 
 ```bash
 MARKER='<!-- ai-issue-loop:decision -->'
+ME=$(gh api user --jq .login)   # the identity every loop agent posts as
 ID=$(gh api "repos/$OWNER_REPO/issues/<N>/comments" \
-      --jq "[.[] | select((.body // \"\") | startswith(\"$MARKER\"))] | .[0].id // empty")
+  | jq -r --arg me "$ME" --arg marker "$MARKER" \
+      '[.[] | select(.user.login == $me and ((.body // "") | startswith($marker)))]
+       | .[0].id // empty')
 if [ -n "$ID" ]; then
   gh api -X PATCH "repos/$OWNER_REPO/issues/comments/$ID" -f body="$MARKER
 $TEXT"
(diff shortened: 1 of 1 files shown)


rtorcato (commented):

Docs-only change to `skills/ai-issue-loop/SKILL.md`, closes #511. Fixes the exact gap described: comment selection for the decision-comment upsert (and the triage duplicate check) previously matched by marker-prefix alone, letting any stranger on a public repo pre-post a comment with that prefix and either hijack the `PATCH` target (repo-write token) or suppress the triage decline comment.

The fix gates both selections on `.user.login == $me` / `.author.login == $me` from `gh api user --jq .login`, and passes the login and marker through `jq --arg` rather than shell/string interpolation into the filter — correctly avoiding jq-filter injection, not just the identity gap. `(.body // "")` null-guard is preserved and documented for both filters. No secrets, no new dependencies, no shell command construction from untrusted input.

### Before merging
Nothing.

rtorcato (commented):

Fixes the exact bug in #511: the decision-comment upsert now gates the id
lookup on `.user.login == $me` (REST shape) via `jq --arg`, retiring the
old `startswith("$MARKER")` shell-interpolated filter that let a
stranger-authored comment win `.[0]` forever. The `// empty` / `(.body //
"")` guards are preserved.

The sibling fix to the triage duplicate check correctly uses
`.author.login` instead of `.user.login` — `gh issue view --json` is
GraphQL-backed (`IssueComment.author.login`), while the Pass 1 upsert
reads the REST comments endpoint (`user.login`); the PR body calls this
out explicitly and it checks out.

Both filters pass identity/marker as jq `--arg` data rather than shell
interpolation, closing the injection surface the old one-liner had.
Docs-only change (SKILL.md), scoped to the reported root cause plus one
same-class sibling in the same file — no separate issue needed for that.

### Before merging
Nothing.

PR #539, opened 2026-08-26
feat(schema): publish the .repo-tooling.json JSON Schema at its $schema URL

### LINKED ISSUE #529: Publish the .repo-tooling.json JSON Schema the lockfiles already point at

Every lockfile the CLI writes carries `"$schema": "https://rtorcato.github.io/repo-tooling/schemas/lockfile.json"` (`src/cli/utils/lockfile.ts:21`) — and that URL serves nothing. No file in the repo generates or publishes a schema, so editors validate nothing and the pointer is a dead link in every consuming repo.

For a tool that wants to be the first thing a dev runs on a new project, the schema is the highest-leverage DX gap: it's what turns `.repo-tooling.json` from *hope you typed it right* into autocomplete, hover docs, and inline validation — with zero documentation reading. It's also the prerequisite for growing the file into a central config surface (see the declarative-settings design issue): a bigger config with no schema is worse than a small one.

Done when:

- A JSON Schema is generated from the `Lockfile` interface (including `config`/`ProjectConfig`, `assets`, `aiLoop`) — generated, not hand-written, so it cannot drift from the type.
- It's published with the docs site at the exact URL the lockfiles already point at.
- CI fails if the committed/published schema is stale against the type.
- A `reference/repo-tooling-json.md` docs page documents the file — ideally rendered from the same schema (descriptions live once, in the TS doc comments).

The docs site deploys from `apps/docs` to `rtorcato.github.io/repo-tooling`, so the publish path already exists — the schema just needs to land in its static output.

## Dogfood

This repo is its own consumer: its `.repo-tooling.json` already carries the `$schema` pointer. Once the schema publishes, add a CI step (or a test) validating **our own lockfile** against the shipped schema — the repo that publishes the schema must never carry a lockfile that fails it.

rtorcato:

Closes #529

Every lockfile the CLI writes stamps `https://rtorcato.github.io/repo-tooling/schemas/lockfile.json` as its `$schema`, and that URL served nothing. This publishes it — plus the equally dead `project-config.json` URL that `CONFIG_SCHEMA.$id` has claimed all along.

## What changed

- **`lockfileSchema()`** (`src/cli/utils/lockfile.ts`) builds the JSON Schema next to the `Lockfile` interface. `satisfies Record<keyof Lockfile, object>` clauses bind the property key sets (top level and `aiLoop`) to the type, so adding or removing a field is a **compile error** until the schema names it. `config` embeds the existing `CONFIG_SCHEMA` (the same one `setup --config-schema` prints), with its nested `$schema`/`$id` dropped so the base URI isn't reset mid-document. It's a function rather than a const because lockfile.ts sits in an import cycle with setup-presets.ts (via the swift scaffolder) — a module-level read of `CONFIG_SCHEMA` hits its TDZ.
- **`pnpm schema:generate`** (`scripts/generate-lockfile-schema.mjs`) writes both schemas into `apps/docs/static/schemas/`, which the existing docs deploy serves verbatim at the exact URLs the `$id`s claim.
- **CI staleness gate**: `tests/cli/utils/lockfile-schema.test.ts` (runs in the existing `pnpm test --run` CI step) fails when a committed schema differs from the source, and pins both `$id`s to the deployed static paths.
- **`reference/repo-tooling-json` docs page** (MDX) renders its field tables from the published `lockfile.json` itself, so descriptions live in one place and the page cannot drift from the schema. Added to the sidebar and `llms.txt`.
- Generated `static/schemas/` excluded from Biome (formatter output is owned by the generator).

## Notes for review

- Verified locally: `pnpm run verify` (biome + typecheck + 1084 tests + build) and `pnpm docs:build` both pass; the schemas land in `apps/docs/build/schemas/`.
- Deliberate ceiling (marked with a `ponytail:` comment): the `satisfies` binding catches added/removed fields, not a changed primitive type — that still needs both lines edited by hand. A full AST-driven generator would need a new dependency and the classic TS compiler API, which the repo's TypeScript 7 no longer ships.
- The issue's "descriptions live once, in the TS doc comments" ideal is approximated: descriptions live once *in the schema* (which the docs page renders); the interface's doc comments remain for code readers.

### DIFF
--- a/apps/docs/biome.json
+++ b/apps/docs/biome.json
@@ -3,6 +3,12 @@
   "root": false,
   "extends": ["../../tooling/biome/preset.json"],
   "files": {
-    "includes": ["!**/.docusaurus", "!**/build", "!**/node_modules", "!**/docs/changelog.md"]
+    "includes": [
+      "!**/.docusaurus",
+      "!**/build",
+      "!**/node_modules",
+      "!**/docs/changelog.md",
+      "!**/static/schemas"
+    ]
   }
 }
--- a/apps/docs/docs/reference/repo-tooling-json.mdx
+++ b/apps/docs/docs/reference/repo-tooling-json.mdx
@@ -1,0 +1,91 @@
+---
+title: .repo-tooling.json
+description: The lockfile the CLI writes — every field, rendered from the published JSON Schema.
+---
+
+import lockfileSchema from '@site/static/schemas/lockfile.json'
+
+export const typeOf = (prop) =>
+	prop.enum ? prop.enum.map(String).join(' | ') : prop.format ? `${prop.type} (${prop.format})` : prop.type
+
+export const Fields = ({ schema }) => (
+	<table>
+		<thead>
+			<tr>
+				<th>Field</th>
+				<th>Type</th>
+				<th>Required</th>
+				<th>Description</th>
+			</tr>
+		</thead>
+		<tbody>
+			{Object.entries(schema.properties).map(([name, prop]) => (
+				<tr key={name}>
+					<td>
+						<code>{name}</code>
+					</td>
+					<td>
+						<code>{typeOf(prop)}</code>
+					</td>
+					<td>{(schema.required ?? []).includes(name) ? 'yes' : 'no'}</td>
+					<td>{prop.description ?? ''}</td>
+				</tr>
+			))}
+		</tbody>
+	</table>
+)
+
+<p>{lockfileSchema.description}</p>
+
+`setup` creates it, `fix` keeps it current, and `doctor` reads it to know what the repo opted into.
+Commit it — it is the repo's memory of its own tooling choices.
+
+## Editor validation
+
+Every written file carries a `$schema` pointer, so editors that understand JSON Schema
+(VS Code out of the box) validate, autocomplete, and show these descriptions inline:
+
+<pre>
+<code>{JSON.stringify({ $schema: lockfileSchema.$id, version: 3 }, null, 2).replace('\n}', ',\n  ...\n}')}</code>
+</pre>
+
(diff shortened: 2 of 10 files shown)


rtorcato (commented):

No injection, secrets, or shell/SQL concerns. `scripts/generate-lockfile-schema.mjs` only imports built CLI modules and writes static JSON via `fs.writeFileSync` — no user input, no `exec`/`eval`. `lockfileSchema()` in `src/cli/utils/lockfile.ts` builds a static object from typed constants. The new MDX docs page renders the schema through JSX (`{}` interpolation, no `dangerouslySetInnerHTML`), so React escaping applies — no XSS vector even though schema content ultimately derives from source-controlled data. No new npm dependencies were added; only a new `schema:generate` script entry. Published URLs (`$id`) match the existing docs deploy path, no new attack surface introduced.

### Before merging
Nothing.

rtorcato (commented):

Verified against #529's requirements: schema is generated (not hand-written) via `lockfileSchema()` with `satisfies` bindings that make an added/removed field a compile error, published at the exact `$id` URLs both lockfiles already stamp, and gated by `lockfile-schema.test.ts` against staleness. `generate-lockfile-schema.mjs` uses `path.join`/`import.meta.url` correctly. The docs MDX renders text via JSX interpolation (no `dangerouslySetInnerHTML`), so no XSS concern. The known ceiling (type changes not caught by `satisfies`) is honestly disclosed with a `ponytail:` comment.

One acceptance-criterion gap: issue #529's "Dogfood" section asked for a CI step/test validating this repo's own `.repo-tooling.json` against the shipped schema — not present in this diff. Doesn't block the core deliverable, filed as follow-up.

Follow-up: #540

### Before merging
Nothing.

PR #544, opened 2026-08-26
feat(doctor): flag an older lockfile and migrate it in place with fix lockfile

### LINKED ISSUE #531: doctor never flags an older lockfile version; repo-tooling's own lockfile is stuck at v2

`doctor` flags a lockfile that is *newer* than the CLI (`doctor.ts:182`) but only reports, never flags, one that is *older* — `migrate()` upgrades in memory on read and the file is rewritten only on the next config write, which for a stable repo may never come.

Concrete evidence: repo-tooling's own `.repo-tooling.json` is `version: 2` with no `assets` while `LOCKFILE_VERSION` is 3 — so asset-drift detection (#428), the newest doctor capability, is inert on the repo that ships it. Any consuming repo set up before v3 is in the same state, silently.

Done when:

- `doctor` reports an older-than-current lockfile as `optional-missing` (not `drift` — nothing is wrong, a capability is just dormant), with the hint to run the existing `fix lockfile` target.
- `fix lockfile` migrates the file on disk to the current version, populating `assets` hashes for presets it can identify.
- This repo's own lockfile is migrated to v3 as the dogfood proof.

## Dogfood wrinkle — the self-repo guard

The `preAction` guard (`src/cli/index.ts:387`) blocks every mutating `fix` inside this repo; only `doctor` has the `REPO_TOOLING_ALLOW_SELF=1` escape. So "migrate this repo's own lockfile" cannot run the very fixer it ships. Pick one in the PR:

- extend the escape to `fix lockfile` specifically (it writes only `.repo-tooling.json`, no scaffolding), or
- migrate our file by hand once and let the new doctor check prove it stays current.

The first is preferred — a dogfood path that requires hand-editing is not dogfooding.

rtorcato:

Closes #531

## What

- `doctor` now reports an older-than-current `.repo-tooling.json` as `optional-missing` (not `drift` — nothing is wrong, a capability is just dormant), with the hint to run `fix lockfile`. Root cause: `migrate()` bumped `version` in memory on read, so doctor could never see the on-disk value; it now preserves it, and the write path stamps `LOCKFILE_VERSION` as before.
- `fix lockfile` migrates an existing file on disk instead of re-inferring: it keeps the recorded config and populates `assets` hashes for presets it can identify — target file byte-matches the shipped asset, so it is provably an unmodified copy. Files that differ stay untracked (indistinguishable fork vs stale copy), and recorded hashes always win over identification.
- Self-repo guard: `REPO_TOOLING_ALLOW_SELF=1 fix lockfile` is the one mutating escape (it writes only `.repo-tooling.json`, no scaffolding), per the issue's preferred option.
- Dogfood proof: this repo's own lockfile is migrated v2 → v3 by the new fixer in this PR, picking up two identifiable docusaurus asset hashes; `doctor` on self now reports the lockfile `ok` at v3.

## Tests

- doctor: older lockfile → `optional-missing` with `fix lockfile` hint
- lockfile utils: read preserves on-disk version for v1/v2 files
- fix: migrating an older lockfile keeps its recorded config and records an identifiable preset hash

1091 tests pass; `biome check` and `tsc` clean.

### DIFF
--- a/.repo-tooling.json
+++ b/.repo-tooling.json
@@ -1,7 +1,8 @@
 {
   "$schema": "https://rtorcato.github.io/repo-tooling/schemas/lockfile.json",
-  "version": 2,
+  "version": 3,
   "config": {
+    "language": "js",
     "projectName": "@rtorcato/repo-tooling",
     "projectType": "library",
     "typescript": {
@@ -23,12 +24,15 @@
     "commitLint": true,
     "semanticRelease": true,
     "securityAutomation": true,
-    "bundler": "none",
-    "language": "js"
+    "bundler": "none"
+  },
+  "assets": {
+    "docusaurus-sync-changelog": "de05aecb200657c2023dc3540dbdd2f9f2f17f1027e67a06266119d12cc05fea",
+    "docusaurus-theme": "3ead9ab407c2676bc404f7df552977a0fb0412b0f0d0bcde223c271a51860005"
   },
   "aiLoop": {
     "agentUser": "rtorcato-bot"
   },
   "writtenBy": "@rtorcato/repo-tooling@3.11.0",
-  "writtenAt": "2026-08-20T17:32:17.649Z"
+  "writtenAt": "2026-08-26T14:58:22.032Z"
 }
--- a/src/cli/commands/doctor.ts
+++ b/src/cli/commands/doctor.ts
@@ -184,6 +184,16 @@ function checkLockfile(lock: Lockfile | null): CheckResult {
 			hint: 'Upgrade @rtorcato/repo-tooling to a release that supports this lockfile version',
 		}
 	}
+	// Not drift: nothing is wrong, a newer capability (e.g. v3 asset-drift
+	// tracking) is just dormant until the file is rewritten (#531).
+	if (lock.version < LOCKFILE_VERSION) {
+		return {
+			check: 'lockfile',
+			status: 'optional-missing',
+			detail: `.repo-tooling.json is v${lock.version}; this CLI writes v${LOCKFILE_VERSION} — newer doctor capabilities stay dormant until it's migrated`,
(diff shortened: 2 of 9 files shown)


rtorcato (commented):

No injection, secrets, or shell/SQL construction in this diff — it's file I/O and JSON only (lockfile read/write, sha256 preset-hash comparison via existing `hashFile`).

The one trust-boundary-adjacent change is the self-repo escape extension in `src/cli/index.ts`: `fix lockfile` now also bypasses the mutating-command guard when `REPO_TOOLING_ALLOW_SELF=1`. It's narrowly scoped (only fires for `name === 'fix' && args[0] === 'lockfile'`, mirrors the existing `doctor` escape, and per the PR description writes only `.repo-tooling.json`, no scaffolding) — consistent with the issue's explicit request. No new dependencies, no path traversal (preset source/target paths are internal constants, not user input).

### Before merging
Nothing.

rtorcato (commented):

Verified the core semantic change: `migrate()` no longer bumps `version` in memory (only fills defaults), so `readLockfile` returns the on-disk version and `checkLockfile` can correctly report `optional-missing` for older files while `writeLockfile` still stamps `LOCKFILE_VERSION` on write. The `fix lockfile` fixer now branches on `lock` presence, preserves the recorded config instead of re-inferring, and merges `identifiablePresetHashes` with recorded hashes winning (matches the PR's stated intent). The self-repo escape is scoped narrowly (`fix` + `lockfile` target only), consistent with the issue's preferred option. Test updates track the semantic shift correctly — `doctor.test.ts`'s `writeLock` helper was bumped to `LOCKFILE_VERSION` (needed since migrate no longer does it for "ok" baselines) while `fix.test.ts`'s `writeLock` intentionally stays at v1 as migration input. Dogfood proof (`.repo-tooling.json` → v3 with two identified asset hashes) matches the issue's closing criteria.

### Before merging
Nothing.

PR #550, opened 2026-08-26
docs(skills): harden ai-issue-loop Pass 0 — worktree bail, agent PR adoption, git env

### LINKED ISSUE #507: ai-issue-loop: core.bare guard doesn't account for GIT_DIR/GIT_WORK_TREE overriding -C
🤖 *Automated — `security-expert` via ai-issue-loop.*

Surfaced reviewing #505. The new `core.bare` guard in `skills/ai-issue-loop/SKILL.md` (Pass 0 and Pass 2) uses `git -C "$ROOT" rev-parse --is-inside-work-tree` to detect the bug and `git -C "$ROOT" config core.bare false` to repair it. Both rely on `-C "$ROOT"` to scope the operation to the main checkout.

If `GIT_DIR` and/or `GIT_WORK_TREE` are exported in the shell that runs the loop, they take precedence over `-C`, so both the probe and the write would silently operate on whatever repo those env vars point to instead of `$ROOT`. Since this skill ships to npm and installs into arbitrary users' `~/.claude/skills/`, a consumer with either variable set in their environment (e.g. some tooling wrapper) could have the guard misdiagnose a different repo, or the repair write to the wrong `.git/config`.

Fix: unset both before invoking git, e.g. `env -u GIT_DIR -u GIT_WORK_TREE git -C "$ROOT" ...`, or use `git --git-dir="$ROOT/.git" --work-tree="$ROOT"` explicitly instead of `-C` so the intended repo always wins regardless of the caller's environment.

### LINKED ISSUE #547: ai-issue-loop: Pass 0's idle bail-out skips Pass 2, so merged worktrees and their ai-wip labels never get cleaned

`skills/ai-issue-loop/SKILL.md:351-353` ends Pass 0 with a bail-out: no open PRs
carrying any `ai-*` label **and** no eligible `ai-ready` issues means the tick
skips straight to Pass 5 with `SUMMARY=idle`.

Pass 2's cleanup is keyed off worktrees **on disk** — `find "$WT_ROOT" … -name
'ai-*'` at `SKILL.md:560` — not off open PRs. So the two are not equivalent, and
there is a state where the bail fires while real cleanup is outstanding: the
moment the last open PR is merged by hand, both bail conditions become true while
its worktree is still present and its issue still carries `ai-wip`. That label is
only ever cleared at `SKILL.md:608`, inside Pass 2. Every subsequent tick meets
the same two conditions, so the worktree and the label survive indefinitely while
the loop reports `idle`.

The two leaks also protect each other. The orphan-worktree rule at
`SKILL.md:644` matches a worktree *"whose issue is not `ai-wip`"* — so the stale
label is exactly what stops the one rule that would otherwise reap the worktree on
a later tick.

## Observed

2026-08-26 on this repo. PRs #543 and #546 were merged by hand. The next tick
found no open PRs and an empty `ai-ready` queue, so it met the bail condition —
while `../repo-tooling-worktrees/ai-541-merge-ready-label` and
`../repo-tooling-worktrees/ai-542-burst-completion-handoff` were both still on
disk, and closed issues #541 and #542 were both still labelled `ai-wip`. They were
only cleaned because that tick deviated from the bail rule deliberately.

Scope of the damage: the 6-in-flight cap counts *open* issues labelled `ai-wip`,
so no concurrency slot leaks here. What does leak is disk, an issue list that
reads as though agents are still working, and Pass 2's tail — the main checkout's
`node_modules` rebuild, which is gated on `REMOVED=1` and therefore never runs.

## Done when

Pass 0's bail condition additionally requires that no `ai-*` worktree exists:

```bash
find "$WT_ROOT" "$ROOT/.claude/worktrees" -maxdepth 1 -name 'ai-*' -type d 2>/dev/null
```

Non-empty ⇒ the tick runs Pass 2 rather than reporting `idle`, even with no open
PRs and an empty queue. A reviewer can check this by reading the amended Pass 0
text against the scenario above: a leftover `ai-*` worktree with no open PR must
route to Pass 2.

## Files

- `skills/ai-issue-loop/SKILL.md` — Pass 0, the bail-out paragraph.
- `apps/docs/docs/guides/ai-issue-loop.md:205` summarises the same rule as
  *"Bail to Pass 5 with `idle` if there is nothing at all"*; check whether that
  line stays accurate once the condition changes.

Doc-only — the skill is prose the orchestrator follows, and no test asserts this
paragraph. `find` rather than a glob is deliberate and already explained at
`SKILL.md:562`.

### LINKED ISSUE #549: ai-issue-loop: an agent-opened PR outside the pipeline is assigned to nobody, because assignment is keyed on review verdicts

Assignment to the human happens in exactly one place — `skills/ai-issue-loop/SKILL.md`
Pass 1, *"Hand a ready PR over properly"* (~line 456) — and it selects on **review
verdicts**: every non-Dependabot PR carrying both `ai-ok-code` and `ai-ok-sec` and
not `ai-changes`. `skills/ai-workflow/SKILL.md` step 4 delegates to that same rule.

So the selector assumes a PR arrived through Pass 4 → implementer → `ai-review`.
A PR an agent opens by any other route carries no `ai-*` label at all, matches no
pass, and is therefore assigned by nothing — it never reaches *Assigned to you*,
which is the view where merges actually happen.

Pass 0 already solves the identical problem for the other author that skips Pass 4:
*"Adopt unlabelled Dependabot PRs"* (~line 334) pulls any `dependabot[bot]` PR with
no `ai-*` label into the pipeline by labelling it `ai-review`. There is no
equivalent for a PR the loop's own identity opened.

## Observed

2026-08-26. PR #548 was opened with a plain `gh pr create` while working on a task
that had no `ai-ready` issue behind it. All five required checks passed and GitHub
reported *Able to merge*, yet Assignees read *"No one—assign yourself"* and Labels
read *"None yet"*. It was only assigned after the owner noticed it by hand.

This is the same class as #542 (*"#537 and #539 were merged by hand before any tick
ran, never appearing in Assigned to you"*). #542 closed the window for PRs the burst
opens; this is the window for PRs opened outside either driver.

## Done when

Pass 0's adoption step also adopts agent-opened PRs, gated on author the same way
the rest of the pipeline gates on identity — `gh api user --jq .login`, as Pass 1's
decision upsert and Pass 3's verdict read already do:

- An open PR whose author is the loop's own login, carrying **no** `ai-*` label,
  **and** whose body begins with the `🤖 *Automated …*` / `🤖 *Opened by …*` header,
  is labelled `ai-review` and joins the pipeline.
- The header is the discriminator that matters: every agent authenticates as the
  owner, so author login alone cannot distinguish a PR an agent opened from one the
  owner opened by hand. **A PR the owner wrote themselves must not be swept in** —
  that would put agent reviewers on work they never asked to have reviewed.
- Pass 1 then assigns it on the existing path once both arms pass, so no second
  assignment rule is needed.

A reviewer can check this by reading the amended Pass 0 against two cases: an
agent-opened unlabelled PR must end up `ai-review`; an owner-written unlabelled PR
must be left alone.

## Files

- `skills/ai-issue-loop/SKILL.md` — Pass 0, beside the Dependabot adoption block.
- `apps/docs/docs/guides/ai-issue-loop.md:205` summarises Pass 0 as *"Adopt
  unlabelled Dependabot PRs"*; check whether that line needs the second case.

Doc-only. Note that `<!-- ai-issue-loop:* -->` markers and the `🤖` header text are
wire format read back off GitHub — match them, do not redefine them.

rtorcato:

Three defects in Pass 0 of `skills/ai-issue-loop/SKILL.md`, batched deliberately: they edit the same region and cannot be reverted independently. Doc-only.

**#547 — the idle bail-out skipped Pass 2's cleanup.** The bail was keyed on open PRs and the `ai-ready` queue; Pass 2's cleanup is keyed on worktrees *on disk*. Merging the last open PR by hand made both bail conditions true while its worktree was still present and its issue still carried `ai-wip` — the label only Pass 2 clears — so every later tick met the same conditions and reported `idle` forever. The two leaks protected each other: the orphan-worktree rule matches only a worktree *whose issue is not `ai-wip`*. The bail now also requires that `find` returns no `ai-*` worktree under either root.

**#549 — an agent-opened PR outside the pipeline was assigned to nobody.** Assignment is keyed on review verdicts, so a PR carrying no `ai-*` label matches no pass and never reaches *Assigned to you*. Pass 0 now adopts these the same way it already adopts unlabelled Dependabot PRs, gated on the loop's own login **and** the `🤖` header — author login alone cannot distinguish an agent-opened PR from one the owner wrote by hand, so the header is what keeps the owner's own PRs out. Pass 1 then hands it over on the existing path.

**#507 — the `core.bare` guard lost to `GIT_DIR`/`GIT_WORK_TREE`.** Both override `-C "$ROOT"`, so with either exported the probe diagnosed a different repo and the repair wrote that repo's `.git/config` — silently, in a skill that installs into arbitrary users' `~/.claude/skills/`. Both invocations now run under `env -u GIT_DIR -u GIT_WORK_TREE`, in **both** Pass 0 and Pass 2.

`apps/docs/docs/guides/ai-issue-loop.md`'s pass table is updated to match: Pass 0 now adopts both kinds of unlabelled PR and bails only when there is no leftover worktree either.

Wire format untouched — `<!-- ai-issue-loop:verdict:* -->`, `<!-- ai-issue-loop:decision -->`, and the `🤖` header are matched, not redefined.

Closes #547
Closes #549
Closes #507

### DIFF
--- a/apps/docs/docs/guides/ai-issue-loop.md
+++ b/apps/docs/docs/guides/ai-issue-loop.md
@@ -202,7 +202,7 @@ Passes run cheapest first, so a quiet repo exits fast.
 
 | Pass | Does |
 |---|---|
-| **0 — orient** | Resolve the main checkout, fetch, list open PRs and `ai-wip` issues. Adopt unlabelled Dependabot PRs. Bail to Pass 5 with `idle` if there is nothing at all. |
+| **0 — orient** | Resolve the main checkout, fetch, list open PRs and `ai-wip` issues. Adopt unlabelled PRs — Dependabot's, and any the loop's own identity opened with the `🤖` header. Bail to Pass 5 with `idle` only if there is nothing at all: no labelled PR, no eligible issue, and no leftover worktree. |
 | **1 — merge** | Auto-merge only *Dependabot* PRs that passed both reviews. Assign every other ready PR to you and drop `ai-review`. Send back anything GitHub reports as not `CLEAN`. |
 | **2 — clean up** | Remove worktrees whose PR merged (confirming the squash is on `main` first), then reap stalls. |
 | **3 — review** | Spawn the missing reviewers for `ai-review` PRs; dispatch a fix round for `ai-changes`. |
--- a/skills/ai-issue-loop/SKILL.md
+++ b/skills/ai-issue-loop/SKILL.md
@@ -266,9 +266,9 @@ immediately after a `worktree remove` and some with nothing removed at all. The 
 is unidentified, so this is detection and repair only:
 
 ```bash
-if [ "$(git -C "$ROOT" rev-parse --is-inside-work-tree 2>/dev/null)" != true ] && [ -d "$ROOT/.git" ]; then
+if [ "$(env -u GIT_DIR -u GIT_WORK_TREE git -C "$ROOT" rev-parse --is-inside-work-tree 2>/dev/null)" != true ] && [ -d "$ROOT/.git" ]; then
   echo "⚠ main checkout bare at $(date -u +%FT%TZ) — repairing"
-  git -C "$ROOT" config core.bare false || {
+  env -u GIT_DIR -u GIT_WORK_TREE git -C "$ROOT" config core.bare false || {
     echo "⚠ repair FAILED — main checkout still bare"; exit 1; }
 fi
 ```
@@ -293,6 +293,14 @@ way and only *prints* the answer, so an exit-code probe is dead code. Verified o
 too, and nothing else separates the two — this skill ships to users' `~/.claude/skills/`,
 where "repairing" someone's real bare clone is the damage rather than the fix. The same
 check skips a linked worktree, whose `.git` is a file.
+
+**`GIT_DIR` and `GIT_WORK_TREE` beat `-C`, so unset them.** When either is exported —
+some tooling wrappers do — git ignores `-C "$ROOT"` and operates on whatever they
+point at, so the probe would diagnose a *different* repo and the repair would write
+that repo's `.git/config`. Both failures are silent, and both are worse than the bug
+being guarded against. This skill installs into arbitrary users' `~/.claude/skills/`,
+so the caller's environment is not ours to assume; `env -u` scopes the unset to the
+one command rather than to the tick.
 
 **Fail loudly.** The repair writes `$ROOT/.git/config`, which a restrictive sandbox
 refuses with `error: could not lock config file .git/config: Operation not permitted` —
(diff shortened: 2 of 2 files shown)


rtorcato (commented):

**#507** — both `git -C "$ROOT"` calls (probe and repair) are now prefixed with `env -u GIT_DIR -u GIT_WORK_TREE`, in Pass 0 and Pass 2. The `-d "$ROOT/.git"` guard that exempts a genuinely bare repo (no `.git` dir) and a linked worktree (`.git` is a file) is untouched, so the fix closes the env-var override without weakening that exemption.

**#549** — adoption gates on `gh api user --jq .login` matched against the PR's `author.login` (a GitHub-attested field an outsider can't set) **and** requires the body start with the `🤖 ` header. Neither signal alone would pass; together they keep a stranger's PR, and an owner-authored PR without the header, out of the pipeline. Matches the existing identity convention used elsewhere in the loop.

**Shell**: `$ME` is passed via `jq --arg`, never interpolated into the filter string; `.body // ""` guards a null body. No command injection, no secrets, no dependency changes — doc-only diff.

### Before merging
Nothing.

rtorcato (commented):

Checked all three "Done when" sections against the diff.

- #547: bail condition now also requires the exact `find "$WT_ROOT" "$ROOT/.claude/worktrees" -maxdepth 1 -name 'ai-*' -type d` check from the issue; non-empty output means the tick continues past Pass 0 into Pass 1/2 instead of bailing. Matches.
- #549: adoption is gated on **both** `.author.login == $me` **and** `(.body // "") | startswith("🤖 ")` — the login-alone trap is avoided, since an owner-written PR without the `🤖` header is excluded.
- #507: the `env -u GIT_DIR -u GIT_WORK_TREE` prefix is applied to both the probe and the repair in Pass 0, and separately in Pass 2's re-check. Both occurrences fixed.
- Wire format (`<!-- ai-issue-loop:verdict:* -->`, `<!-- ai-issue-loop:decision -->`, `🤖` header) is matched, not redefined, anywhere in the diff.
- Docs table (`apps/docs/docs/guides/ai-issue-loop.md`) updated consistently with the SKILL.md prose.

### Before merging

Nothing.

PR #551, opened 2026-08-26
test(skills-install): harden commandLines against a command spanning lines

### LINKED ISSUE #503: Harden the skills-install test helper and mark shellQuote as POSIX-only

Two non-blocking findings from the reviews on #502, filed so they outlive the merge rather than sitting in a `### Before merging` section nobody reads twice.

## 1. The test helper can quietly narrow its own assertion

`code-reviewer`: `commandLines()` in `tests/cli/generators/skills-install.test.ts` filters lines starting with `npx skills add `, so it isolates only the **first physical line** of a command. Quoting does not escape a newline, so a skill name carrying one produces a command spanning several lines — and the continuation is silently excluded from the `rest` the test asserts is empty.

No current test case trips this. The risk is a future newline-bearing case that looks like it asserts more than it does. Harden the helper to consume through the next `npx skills add ` boundary rather than to end-of-line.

## 2. `shell.ts` is named for wider use than it is correct for

`security-expert`: `src/cli/utils/shell.ts` exporting a bare `shellQuote` invites reuse in contexts where POSIX single-quoting is the wrong escape — a Windows shell, or a double-quoted context. The implementation is right for what it does; the name and location oversell it.

A doc comment saying **POSIX `sh` only** heads that off. Cheap, and the file is new, so now is when it costs nothing.

## Related

- #498 / #502 — the work these came out of

rtorcato:

Two small hardening items from the #502 reviews. No behaviour change — a test helper and a doc comment.

## 1. `commandLines()` could quietly narrow its own assertion

It filtered lines starting with `npx skills add `, so it isolated only the **first physical line** of a command. Quoting does not escape a newline, so a skill name carrying one produces a multi-line command whose continuation was silently excluded from the `rest` the test asserts is empty — the assertion looked wider than it was.

It now takes the fenced block and splits on the next `npx skills add ` boundary. `shellArgv` also separates argv with NUL rather than a newline, so a newline-bearing name round trips instead of coming back as two arguments.

A new `a newline` case in the `it.each` table proves the fix rather than asserting it. Against the old helper it fails on the truncated, unbalanced-quote command:

```
Error: Command failed: sh -c printf '%s\0' skills add https://github.com/rtorcato/x --skill 'one
sh: -c: line 0: unexpected EOF while looking for matching `''
```

## 2. `shellQuote` is now documented as POSIX `sh` only

The implementation was already right; the bare name invited reuse where single-quoting is the wrong escape — `cmd.exe`/PowerShell, or a double-quoted context. The doc comment says so.

## Verification

`vitest run tests/cli/generators/skills-install.test.ts` — 13 passed. `biome check` and `tsc --noEmit --project src/cli/tsconfig.json` clean (pre-commit ran both).

`test:` type — semantic-release cuts no release.

Closes #503

### DIFF
--- a/src/cli/utils/shell.ts
+++ b/src/cli/utils/shell.ts
@@ -1,8 +1,11 @@
 /**
  * Quote a value as one literal argument **for a POSIX `sh` command line, and
- * nothing else**. The result is wrong in `cmd.exe` and PowerShell, which do not
- * treat `'` as a quote, and wrong anywhere the value lands inside double quotes
- * or an existing quoted string — it is a whole argument, not a fragment.
+ * nothing else**. The result is wrong in `cmd.exe`, which does not treat `'` as
+ * a quote at all, and wrong in PowerShell, which does — but escapes an embedded
+ * `'` by doubling it rather than POSIX's `'\''`, so this output still mangles
+ * any value containing one. It is also wrong anywhere the value lands inside
+ * double quotes or an existing quoted string — it is a whole argument, not a
+ * fragment.
  *
  * POSIX single-quote escaping — the only form correct for arbitrary bytes.
  * Inside single quotes every character is literal, so `'` is the sole one
--- a/src/cli/utils/shell.ts
+++ b/src/cli/utils/shell.ts
@@ -1,4 +1,9 @@
 /**
+ * Quote a value as one literal argument **for a POSIX `sh` command line, and
+ * nothing else**. The result is wrong in `cmd.exe` and PowerShell, which do not
+ * treat `'` as a quote, and wrong anywhere the value lands inside double quotes
+ * or an existing quoted string — it is a whole argument, not a fragment.
+ *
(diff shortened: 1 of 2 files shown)


rtorcato (commented):

The new `a newline` case genuinely proves the fix rather than just asserting it: on the old line-filtering `commandLines()`, the extracted `command` is truncated with an unbalanced quote, so `shellArgv` throws (`unexpected EOF while looking for matching '`) instead of silently passing. `shellArgv`'s switch to NUL-delimited argv correctly round-trips a newline inside an argument. Diff scope matches #503 exactly.

Nit: the `shellQuote` comment says cmd.exe *and* PowerShell "do not treat `'` as a quote." True for cmd.exe; PowerShell does quote with `'`, it just escapes an embedded `'` by doubling it rather than POSIX's `'\''` — output is still wrong there, just for a different reason than stated.

### Before merging
Nothing.

PR #553, opened 2026-08-26
feat(cli): add `loop guard` for bare-checkout repair and node_modules rebuild gating

### LINKED ISSUE #519: Drain ai-issue-loop mechanics from SKILL.md prose into tested CLI code

## What

`skills/ai-issue-loop/SKILL.md` is a ~1,200-line prose program interpreted by an LLM. Prose drifts and nothing fails when it does — the #480 fork proved it, and #500 (repo-deleting commit) happened in the gap. The countermeasure already working in this repo: mechanics live in tested CLI code, the skill pins to it (`labels.ts` owning the label table, asserted against the doc by a test).

Continue that migration: **prose for judgment, code for mechanics.**

## First slice: `repo-tooling loop guard`

A subcommand doing deterministically what the two most dangerous prose-with-shell blocks describe:

- Pass 0/2 bare-checkout detection and repair (`core.bare` flip, stdout-not-exit-code probe, `.git`-is-a-directory discriminator, halt-the-tick exit code)
- Pass 2 `node_modules` rebuild gating (`REMOVED`/live-worktree check, `--frozen-lockfile`)

The skill's snippets become one line each: `npx @rtorcato/repo-tooling loop guard --root "$ROOT"` with documented exit codes. Tests hold the invariants; the skill shrinks.

## Later slices (each its own issue when ripe)

- Pass 1 decision-comment upsert (marker find + PATCH, `// empty` guard) → `loop comment`
- Pass 3 verdict read-back (`--slurp`, login gate) → `loop verdict <pr>`
- Pass 5 status file + digest → `loop status`

## Success measure

SKILL.md line count goes down every month. If it grows two months running, this issue is being lost.

rtorcato:

Closes #519

First slice of #519: the two most dangerous prose-with-shell blocks in `skills/ai-issue-loop/SKILL.md` become one tested command. Prose drifts and nothing fails when it does — PR #500's `0 additions, 67703 deletions` commit happened in exactly that gap.

## `repo-tooling loop guard --root <path>`

**Bare-checkout detection and repair.** Every detail the prose established is preserved and now held by a test:

- Probes **stdout, not the exit code** — `git rev-parse --is-inside-work-tree` exits `0` either way and only *prints* the answer, so an exit-code probe is dead code.
- Repairs only when `.git` is a **directory**. A genuinely bare repo prints `false` too, and this ships to consumers' machines, where flipping a real bare clone is the damage rather than the fix. The same discriminator skips a linked worktree, whose `.git` is a file.
- A failed repair exits non-zero so the tick halts, instead of reporting a healthy repo while it stays broken.

**`node_modules` rebuild gating.** The three load-bearing conditions (`--removed`, a `pnpm-lock.yaml`, no live `ai-*` worktree in either location) and `--frozen-lockfile --config.confirmModulesPurge=false`. A deferral or a failed rebuild is reported, never fatal — it cannot corrupt a commit the way a bare checkout does.

**Documented exit codes**, so each skill snippet can become one line:

| Code | Meaning | Caller |
|---|---|---|
| `0` | Root is a usable work tree — healthy, or repaired in place | Continue the tick |
| `1` | Repair attempted and failed; still bare | **Halt the tick** |
| `2` | Not a repairable main checkout (bare clone, linked worktree, not a repo) | **Halt the tick**; needs a human |

Also on `--help`, in `README.md` and `AGENTS.md`.

## Tests

The SKILL.md invariant table is the spec, asserted twice: against the classifier, and against repositories git actually created — healthy checkout, wrongly bare, genuinely bare, linked worktree. The table is a claim about git's behaviour, and a pure test can only re-state it. The genuinely-bare case additionally asserts `core.bare` is still `true` afterwards: not damaging a real bare clone is the invariant, not a side effect.

## Second commit — a real bug found while testing

`git` exports `GIT_DIR` / `GIT_INDEX_FILE` to everything a hook runs, and they outrank both `cwd` and `-C`. Under `pre-push`, `realGitExec` answered about *this* repo rather than the directory it was handed — for a guard whose entire job is deciding whether one specific checkout has gone bare, that is the mass-deletion failure mode itself. It also already broke `tests/cli/generators/security.test.ts`, which does `git init` in a temp dir and was silently operating on the real index; CI has no hook environment, so it only ever failed on a developer's machine. `realGitExec` now strips the ambient repo vars, `vitest.config.mjs` drops them for the whole run, and a test covers it (verified failing without the fix).

## Not in this PR

**`skills/ai-issue-loop/SKILL.md` is untouched** — wiring the skill to this command is a deliberate follow-up, kept separate to avoid conflicting with concurrent edits to that file. The later slices in #519 (`loop comment`, `loop verdict`, `loop status`) each get their own issue when ripe.

---

`pnpm verify` (biome + tsc + 1122 tests + build) passes; the full suite also passes under a simulated hook environment.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -24,6 +24,7 @@ Every command supports `--json` and a non-interactive mode. Combine with `--yes`
 | `fix --dry-run` | ✅ | ✅ | Print what each fixer would write without writing. Combine with `--json`. |
 | `list --json` | ✅ | ✅ | Enumerate the library's surface area. Each entry has `{ name, description, exports, fixTarget }`. |
 | `copy <name>` | ✅ | text only | Copy a single preset (`biome`, `tsconfig`) into the current directory. |
+| `loop guard --root <path>` | ✅ | ✅ | Guard an `ai-issue-loop` tick: repair a wrongly-bare main checkout, gate the `node_modules` rebuild (`--removed`). Exit `0` continue, `1` repair failed, `2` root is not a repairable checkout — both non-zero halt the tick. |
 
 ## Recommended workflows
 
@@ -104,6 +105,7 @@ A fixer may also **refuse** — the target file holds something the generator ca
 - `src/cli/commands/doctor.ts` — all checks and the public `runDoctor(dir)` / `evaluateNodeVersion(version)` / `nextStepSuggestions(results)`
 - `src/cli/commands/fix.ts` — `Fixer` interface, fixer registry, `fixCommand`
 - `src/cli/commands/fix-targets.ts` — shared check → fix target map (used by both doctor's footer and fix's lookup)
+- `src/cli/commands/loop-guard.ts` — `loop guard`: the `--is-inside-work-tree` / `.git` invariant table and the `node_modules` rebuild gate, drained out of the ai-issue-loop skill's prose (#519)
 - `src/cli/generators/` — one file per concern (linting, testing, build, git, github-actions, security, misc)
 - `tooling/` — every shipped preset, mirrored 1:1 with `package.json` `exports`
 
--- a/README.md
+++ b/README.md
@@ -91,6 +91,7 @@ See the [Getting Started guide](https://rtorcato.github.io/repo-tooling/guides/g
 | `copy <config>` | Copy a single config file into the current project. | `npx @rtorcato/repo-tooling copy biome` |
 | `doctor` | Diagnose an existing project for missing or drifted tooling. | `npx @rtorcato/repo-tooling doctor` |
 | `fix [target]` | Apply scaffolders for what `doctor` flagged (`--yes`, `--dry-run`, `--diff`). | `npx @rtorcato/repo-tooling fix` |
+| `loop guard` | Repair a main checkout that has gone `core.bare = true`, and gate the `node_modules` rebuild after a worktree removal. Exits `1` if the repair failed and `2` if the root is not a repairable checkout — see `--help`. | `npx @rtorcato/repo-tooling loop guard --root .` |
 
 Prefer to run the audit in CI? `doctor` also ships as a GitHub Action:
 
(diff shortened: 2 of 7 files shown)


rtorcato (commented):

Verified the bare-checkout repair only fires on `wrongly-bare` (`.git` dir + `false` probe); genuine bare clones and linked worktrees route to exit 2 without touching config, and ambient `GIT_DIR`/`GIT_INDEX_FILE`/etc. are stripped before every git spawn so a hook environment can't redirect the write — all covered by tests. All `spawn()` calls (git, pnpm) use array args, no shell, no interpolation. Rebuild gating requires no live `ai-*` worktree, a lockfile, and `exitCode === 0`, using `--frozen-lockfile`. No secrets logged, no new dependencies.

### Before merging
Nothing.

rtorcato (commented):

Verified the safety-critical invariants directly, not just by reading:

- `classifyRoot` matches the SKILL.md table exactly: stdout (not exit code) drives the verdict, `.git`-is-a-directory gates repair, a genuinely-bare repo and a linked worktree (`.git` is a file) are both left untouched. Ran the 25-test file — all pass, including the genuinely-bare case asserting `core.bare` stays `true`.
- `realGitExec` has exactly one other caller (`checkGitIdentity`), and it already passes `cwd`. No caller relied on ambient git env.
- Reproduced the bug: exported `GIT_DIR`/`GIT_INDEX_FILE` pointing at this worktree's own `.git` and re-ran `security.test.ts` + `loop-guard.test.ts` — 47/47 pass, confirming the `vitest.config.mjs` strip works under a simulated hook environment.
- `loop guard --root /tmp --json` exits `2` with the documented shape, matching AGENTS.md/README.md.
- `skills/ai-issue-loop/SKILL.md` is untouched, as #519 required.
- `loop` is outside the `isSelfRepo` preAction hook, consistent with the stated intent that this guard must run against this repo too.

### Before merging
Nothing.

PR #543, opened 2026-08-26
feat(labels): give the human's pickup state a merge-ready label instead of an absence

rtorcato:
🤖 *Automated — implementer via ai-issue-loop.*

"Waiting on the human to merge" was the only state in the label machine expressed as an absence (`ai-ok-code + ai-ok-sec` with no `ai-review`). This gives it a label.

- `merge-ready` added to `LOOP_LABELS` (`src/base/labels.ts`) — `#8250df`, GitHub's merged-purple. `fix labels` and the doctor drift check pick it up from the table automatically; the bootstrap-block sync test covers the skill copy.
- Pass 1 applies it in the handoff `gh pr edit`, gated on the `mergeStateStatus: CLEAN` probe — so it asserts reviews passed **and** GitHub will accept the merge.
- Pass 1 strips it on the non-CLEAN send-back and reconciles it idempotently every tick: carrying it while no longer CLEAN or missing a pass label ⇒ stripped. Derived state; the conjunction stays the source.
- Pass 3's fix-round relabel strips it (diff changed ⇒ readiness stale).
- Dependabot arm: a bot PR held for the human (`ai-notes`) gets it too; auto-merge-armed ones don't.
- State diagram, label table, and the two "reads one of two ways" rows updated; `ai-loop-status` and the docs guide follow suit.

Naming: `merge-ready`, unprefixed — it names the PR's state for the human consuming it, like `holding`, not an agent's claim.

Closes #541

PR #554, opened 2026-08-26
fix(skills): send CI-red PRs back to a fix round and collapse the ready-PR labels

### LINKED ISSUE #527: Reconcile ai-* skill worktree deps with the worktree.symlinkDirectories check after #406

## What

`repo-tooling` ships two subsystems that give opposite advice about agent-worktree dependencies, and one of them is now stale.

- **The doctor check** (`checkClaudeWorktreeSettings`, `src/languages/js/checks.ts:1109`, wired at `doctor.ts:429`) tells every consumer to symlink `node_modules` into agent worktrees, and `fix ai` writes `worktree.symlinkDirectories` for the root plus each workspace package (#396, #406).
- **`skills/ai-issue-loop/SKILL.md`** (Pass 4) says to symlink *only* for app-confined issues and to run a real `pnpm install` for anything touching a workspace package. `SKILL.md:1224` says outright: *"Replaces worktree.symlinkDirectories."*

## Why it matters

`worktree.symlinkDirectories` is a **Claude Code** setting, honoured by `EnterWorktree`. The ai-* skills forbid `EnterWorktree` and create worktrees with raw `git worktree add`, so **the setting is inert for exactly the worktrees the loop creates**. `doctor` can report `Claude worktree settings: ok` while every agent worktree gets dependencies by some other path — as happened on #511/PR #526, where the orchestrator hand-installed.

The skill's stated reason for avoiding symlinks looks stale: it cites per-package `node_modules` not being symlinked (`api-common`, 37/37 packages in main vs 1 in the worktree; #201 handed back `ai-blocked` over it). **#406 closed that gap** — `workspaceSymlinkDirs` globs the repo's own workspace globs and symlinks each package's `node_modules`.

What still holds is a separate hazard, independent of coverage: `pnpm install` against a symlinked tree purges the **main checkout's** `node_modules` (`ERR_PNPM_ABORTED_REMOVE_MODULES_DIR_NO_TTY`), and `worktree remove --force` destroys its `.bin` shims — `browser-common` #145 → PR #147, where the implementer was explicitly told not to install and did anyway. That hazard is why the loop carries its Pass 2 rebuild step.

## The decision

Given #406, should the ai-* skills drop their bespoke symlink/install logic and just honour `worktree.symlinkDirectories`, keeping only the never-install rule (and Pass 2's rebuild as the backstop)?

| Option | Consequence |
|---|---|
| **A** — skills honour the setting | One source of truth; `doctor` becomes meaningful for loop worktrees. Re-exposes the purge/`.bin` hazard, so the never-install rule becomes load-bearing rather than advisory. |
| **B** — keep the split, document it | No behaviour change. The check keeps reporting `ok` on repos where it governs nothing, which is the misleading part. |
| **C** — real install always for skill worktrees | Safest (measured 11.5s here, since pnpm hardlinks from the store). Costs a duplicate `node_modules` per worktree and makes the check irrelevant to the loop. |

Not labelling `ai-ready` — deciding this *is* the work, and it spans a security-adjacent blast radius (a mutation that reaches the main checkout).

## Evidence

Observed 2026-08-26 running `/ai-workflow` on #511. Option C was used ad hoc for PR #526's worktree; the main checkout's `.bin` came through intact (32 shims), which is the outcome the Pass 2 rebuild exists to restore after a symlinked worktree is removed.

**Docs are in scope:** `apps/docs/docs/guides/ai-setup.md` §"Worktrees: symlink `node_modules` instead of reinstalling it" documents the symlink approach as the recommendation — whatever option this issue lands on, that section must say the same thing.

### LINKED ISSUE #545: ai-issue-loop: send CI-red PRs back to the agent instead of waiting for a human to notice

A loop PR whose CI goes red currently waits for a human to notice. Pass 1 only reacts to non-`CLEAN` at *handoff time* (both reviews passed), and the `ci-red` count only covers auto-merge-armed Dependabot PRs. Observed on #543 (2026-08-26): `build` failed within minutes of the PR opening, both reviewers passed the diff (they never see CI), and nothing in the pipeline was ever going to dispatch a fix — the human found it via the red ✗ on the PR page.

## What

Add a **CI-red send-back rule** to `skills/ai-issue-loop/SKILL.md` (Pass 1, beside the existing non-`CLEAN` check):

Any open non-Dependabot PR carrying `ai-*` labels where `statusCheckRollup` has a **completed FAILURE on a required check**, and no `ai-reviewing-*` claim is active (let in-flight reviewers finish — the fix-round relabel re-runs them anyway):

1. `gh pr edit <N> --add-label ai-changes --remove-label ai-ok-code --remove-label ai-ok-sec --remove-label ai-notes --remove-label merge-ready` — `ai-changes` **is** the send-back label; Pass 3 dispatches the fix-round implementer off it, same 2-round budget.
2. Marker-upsert comment (Pass 1's existing mechanism) naming the failing check and pasting the relevant `--log-failed` excerpt — the fix-round prompt reads comments as instructions, so this is what the agent acts on.
3. Count it as `ci-red` in Pass 5's summary with the `⚠`.

The comment must also say: **the fix may not be code.** #543's failure was the dogfood check finding a bootstrap gap (new label in the table, not yet on the repo) — the correct fix was `gh label create`/`fix labels` or an `ACCEPTED` entry in `scripts/dogfood.mjs`, not a branch edit. The implementer has repo-write; the comment should leave that path open rather than steering it into editing code that isn't wrong.

No new label: `ai-changes` + the comment already express "sent back, and why". If list-view distinction between review-rejected and CI-rejected ever matters, add `ci-failing` as a *rider* on `ai-changes` then — not speculatively now.

⚠ Same file as PR #543 — label `ai-ready` only after #543 merges, or the worktree conflicts.

rtorcato:

Three changes to `skills/ai-issue-loop/SKILL.md`, batched because they edit the same file and cannot be reverted independently.

## 1. CI-red send-back — Closes #545

Pass 1 only reacted to a non-`CLEAN` `mergeStateStatus` at handoff time, so a PR whose CI went red waited for a human to spot the ✗ (observed on #543). Reviewers are diff-scoped and never see CI, so nothing in the pipeline was going to dispatch a fix.

Pass 1 now sends back any open **non-Dependabot** PR carrying an `ai-*` label with a completed `FAILURE` on a **required** check: `ai-changes`, the superseded labels stripped, and a marker-upsert comment naming the check and pasting the `--log-failed` excerpt — including the issue's point that **the fix may not be code** (#543's red check was a bootstrap gap fixable with `gh label create`, not a branch edit). Counted `ci-red` for Pass 5, which already carries the `⚠`.

Two carve-outs: a live `ai-reviewing-*` claim (per the issue), and a PR **already** carrying `ai-changes` — re-applying is not free, since Pass 3 counts label applications off the timeline and stops at three, so a 15-minute loop would exhaust the fix-round budget within the hour and mark the issue `ai-blocked` with no agent having done anything.

**One deviation from the issue's spec:** step 1 also drops `ai-review`. Leaving it on has Pass 3 spawn reviewers *and* a fix round against one PR, reviewing a diff being rewritten underneath them. It mirrors what a `CHANGES` verdict already does, and the implementer re-adds it on push.

Dependabot PRs keep the old behaviour — flagged `ci-red`, never sent back, since Pass 3 treats `ai-changes` on a bot PR as terminal.

`--required` rather than the whole rollup: `statusCheckRollup` also carries optional and third-party contexts, and an advisory check going red should not spend a fix round.

## 2. Worktree dependency advice reconciled — Closes #527

**The issue's claim about #406 verified against the code and holds.** `workspaceSymlinkDirs` (`src/cli/generators/agent-rules.ts:100`) globs the repo's own `pnpm-workspace.yaml` / `package.json` `workspaces` and emits `<workspace>/node_modules` for each one that exists, so `worktree.symlinkDirectories` is complete. The skill's Pass 4 hand-rolled loop was the incomplete one — it linked the root and `apps/*` only, reproducing the exact `api-common` 37-of-37-vs-1 gap on any repo that nests packages elsewhere.

The issue's other point is the one that shapes the fix: `worktree.symlinkDirectories` is a **Claude Code** setting honoured by `EnterWorktree`, which the ai-* skills forbid — so it is inert for exactly the worktrees the loop creates. So neither Option A (honour the setting) nor C (always install) as written. Pass 4 now **reads the list as data** and does the `ln -s` itself: one source of truth for *what* to link, two mechanisms for *how*, and `doctor`'s *Claude worktree settings* check becomes meaningful for loop worktrees. No list, or no `.claude/settings.json` → a real `pnpm install` in the (unsymlinked, therefore safe) worktree.

That drops the app-confined carve-out — the branch is now keyed on data rather than on the implementer's judgement about which issue is "confined to an app", which is what got it wrong. The never-install-against-a-symlink rule becomes load-bearing rather than advisory, as the issue predicted, and says so.

`apps/docs/docs/guides/ai-setup.md` §Worktrees, `apps/docs/docs/guides/ai-issue-loop.md`, `skills/ai-workflow/SKILL.md` (which restated the old rule verbatim), and the `checkClaudeWorktreeSettings` doc comment all now say the same thing.

## 3. Collapse the ready-PR label set — no issue

Pass 1 added `merge-ready` without removing the two labels it supersedes, so a handed-off PR wore `ai-ok-code`, `ai-ok-sec` **and** `merge-ready` — one fact stated three times, with the reader having to know which is strongest. The handoff now strips both pass labels, leaving exactly:

| Labels | Means |
|---|---|
| `merge-ready` | Merge freely. |
| `merge-ready`, `ai-notes` | Passed, but open the comments first. |

`ai-notes` is never stripped; that rule stands untouched.

Checked every later reader of the pair, and adjusted three:

- **The reconcile rule** stripped `merge-ready` from a PR "missing either pass label" — which after this change would undo the previous tick on every handed-off PR, leaving it with no labels at all, matching no selector in any pass. It now keys on `CLEAN` (and `ai-changes`) only, and the file says why.
- **Pass 1's own handoff selector** and **the Dependabot auto-merge selector** now accept `merge-ready` in place of the pair, as does the gated-repo auto-merge arm. Stated once as a rule: `merge-ready` satisfies every later test for the pair.
- **Pass 3** is unaffected — it acts on `ai-review`, which the handoff drops — and the fix-round relabel already removed all four labels.

`src/base/labels.ts` and the SKILL.md bootstrap block are untouched, so `tests/base/labels.test.ts` stays green. `skills/ai-loop-status/SKILL.md` already read `merge-ready` first with the pair as a fallback.

## Verification

`pnpm run check` (0), `pnpm run typecheck` (0), `pnpm test --run` — 1073 passed, and `pnpm dogfood` clean (61 checks) after a build.

⚠ During this run the **main checkout flipped to `core.bare = true`** again (the #500 failure mode Pass 0 documents), staging 740 phantom deletions in the worktree index. Caught before any second commit; repaired with `git config core.bare false` + `git reset`, and this branch is one clean 5-file commit. Noting it as another datapoint — it happened right after `pnpm build-cli`, with no `worktree remove` involved.

### DIFF
--- a/apps/docs/docs/guides/ai-issue-loop.md
+++ b/apps/docs/docs/guides/ai-issue-loop.md
@@ -67,7 +67,7 @@ below.
 Two consequences, and both are load-bearing:
 
 1. **Approval is a label**, not a review. `ai-ok-code` / `ai-ok-sec` record that
-   an agent passed the diff.
+   an agent passed the diff, until the handoff replaces them with `merge-ready`.
 2. **Required status checks stay the real merge gate.** Never set
    `required_pull_request_reviews` on the protected branch — required review
    deadlocks every PR the loop opens.
@@ -129,11 +129,23 @@ that state, so a missed tick, a crash, or a restart costs nothing.
 | `ai-review` | PR | Awaiting agent review. |
 | `ai-reviewing-code` | PR | `code-reviewer` claimed and running. Cleared with its verdict. |
 | `ai-reviewing-sec` | PR | `security-expert` claimed and running. Cleared with its verdict. |
-| `ai-ok-code` | PR | `code-reviewer` passed. |
-| `ai-ok-sec` | PR | `security-expert` passed. |
-| `ai-changes` | PR | A reviewer requested changes. |
+| `ai-ok-code` | PR | `code-reviewer` passed. In-flight only — the handoff strips it. |
+| `ai-ok-sec` | PR | `security-expert` passed. In-flight only — the handoff strips it. |
+| `ai-changes` | PR | A reviewer requested changes, or Pass 1 sent the PR back over red CI. |
 | `ai-notes` | PR | Passed, but a reviewer left something to read before merging. |
-| `merge-ready` | PR | Both agent reviews passed and the PR is mergeable — waiting on a human. |
+| `merge-ready` | PR | Both agent reviews passed and the PR is mergeable — waiting on a human. Supersedes the `ai-ok-*` pair rather than joining it. |
+
+A PR handed over for you to merge therefore carries exactly one of two label
+sets, and the difference is legible without opening anything:
+
+| Labels | Means |
+|---|---|
+| `merge-ready` | Merge freely. |
+| `merge-ready`, `ai-notes` | Passed, but open the comments first. |
+
+`merge-ready` asserts strictly more than `ai-ok-code` + `ai-ok-sec` — both
+reviews passed *and* GitHub reports the PR mergeable — so the handoff drops the
+pair rather than stacking three labels that all say "passed".
 
 Colours carry meaning here — `ai-ready` is green and `ai-blocked` red precisely
 so the two states a maintainer must tell apart are legible at a glance. `doctor`
@@ -151,7 +163,8 @@ PR: ai-review ─> ai-reviewing-* ─┬─> ai-ok-code + ai-ok-sec ─┬─ is
                                  │        (± ai-notes)       │              ─> YOU merge
                                  │                           └─ dependabot ─> auto-merge
                                  └─> ai-changes ─> fix round (max 2) ─> ai-review
-                                                             └─ round 3 ─> ai-blocked
+                                     ▲                       └─ round 3 ─> ai-blocked
+                                     └─ Pass 1 sends back: not CLEAN, or a required check FAILED
 ```
 
 `ai-reviewing-code` / `ai-reviewing-sec` are the claim step. Pass 3 applies one
@@ -193,8 +206,11 @@ gh api repos/$OWNER_REPO/branches/main/protection \
 ```
 
 Worktrees also want `node_modules` symlinked in, so an agent can typecheck
-without a full install per issue. `fix ai` writes that into
-`.claude/settings.json`.
+without a full install per issue. `fix ai` writes that list — the root plus every
+workspace package — into `.claude/settings.json` as
+`worktree.symlinkDirectories`, and Pass 4 reads it and creates the links itself.
+Without it the loop still works: each worktree gets a real `pnpm install`
+instead, which costs a duplicate `node_modules` per issue.
 
 ## The tick
 
@@ -203,7 +219,7 @@ Passes run cheapest first, so a quiet repo exits fast.
 | Pass | Does |
 |---|---|
 | **0 — orient** | Resolve the main checkout, fetch, list open PRs and `ai-wip` issues. Adopt unlabelled PRs — Dependabot's, and any the loop's own identity opened with the `🤖` header. Bail to Pass 5 with `idle` only if there is nothing at all: no labelled PR, no eligible issue, and no leftover worktree. |
-| **1 — merge** | Auto-merge only *Dependabot* PRs that passed both reviews. Assign every other ready PR to you and drop `ai-review`. Send back anything GitHub reports as not `CLEAN`. |
+| **1 — merge** | Auto-merge only *Dependabot* PRs that passed both reviews. Hand every other ready PR to you as `merge-ready`, dropping `ai-review` and both `ai-ok-*`. Send back anything GitHub reports as not `CLEAN`, or with a required check red. |
 | **2 — clean up** | Remove worktrees whose PR merged (confirming the squash is on `main` first), then reap stalls. |
 | **3 — review** | Spawn the missing reviewers for `ai-review` PRs; dispatch a fix round for `ai-changes`. |
 | **4 — pick up** | Claim eligible `ai-ready` issues, create the worktree, spawn an implementer. |
--- a/apps/docs/docs/guides/ai-setup.md
+++ b/apps/docs/docs/guides/ai-setup.md
@@ -132,6 +132,18 @@ Only the `worktree.symlinkDirectories` key is touched — your `hooks`,
 yourself are kept. A file that doesn't parse is left alone rather than
 clobbered; `doctor` reports it as drift for you to repair by hand.
 
+### Two things honour that list
+
+`worktree.symlinkDirectories` is a **Claude Code** setting, and Claude Code
+honours it when *it* creates the worktree — i.e. via `EnterWorktree`. The `ai-*`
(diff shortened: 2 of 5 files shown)


rtorcato (commented):

Reviewed the new shell in Pass 1's CI-red send-back and Pass 4's symlink loop. The `gh pr checks | jq` pipeline uses a static filter over `gh`'s own JSON (no PR-title/branch-name text spliced into the filter), and the worktree symlink loop reads `worktree.symlinkDirectories` from the trusted main-checkout `.claude/settings.json`, not from the untrusted PR diff — consistent with the loop's existing trust model. No marker format (`ai-issue-loop:verdict:*` / `:decision`) is touched, and the decision-comment author gate is untouched. `checks.ts` only gained a doc comment — no behavior change, no write-outside-repo or auto-fixer risk.

One gap: the CI-red rule's step 2 says to paste a `gh run view <run-id> --log-failed` excerpt into the marker-upsert comment, but gives no concrete command showing *how* — CI log content is attacker-influenceable (a PR author's own branch) and pasting it into a `--body "$(...)"` string via command substitution is the classic shell-injection shape. The prose defers to "Pass 1's existing mechanism," which this diff doesn't show, so I can't confirm it writes the excerpt to a file (`--body-file`) rather than interpolating it inline.

### Before merging
Confirm the marker-upsert comment mechanism writes the `--log-failed` excerpt via `--body-file` (or otherwise avoids inline command substitution) rather than splicing raw CI log text into a `--body "$(...)"` shell string.

rtorcato (commented):

#545: send-back correctly skips Dependabot PRs and honours both carve-outs (`ai-reviewing-*` claim, existing `ai-changes`). Step 1 also drops `ai-review` — a disclosed deviation, consistent with the existing `CHANGES`-verdict precedent.

Label collapse: checked every later reader (Pass 1 selector, both auto-merge arms, reconcile rule) — all now accept `merge-ready` in place of the pair; nothing still assumes `ai-ok-*` survives handoff. `ai-notes` only stripped on send-back, never on clean handoff. `labels.ts` untouched, matches the table.

#527: correctly reconciled — the loop reads `worktree.symlinkDirectories` as data and links it itself, since `EnterWorktree` (the setting's only consumer) is forbidden here.

### Before merging
Nothing.

PR #555, opened 2026-08-26
fix(scripts): strip git's ambient repo env vars in the integration scripts

rtorcato:

**#553 closed two of the three paths, not all three — and the main checkout flipped to `core.bare = true` again after it merged.** That is what surfaced this.

Git exports `GIT_DIR` / `GIT_INDEX_FILE` and friends to every child of a hook, and they outrank `cwd` and `-C`. #553 guarded `realGitExec` and vitest. The integration scripts are the third path, and they are the ones that actually run `git init`:

- `scripts/integration/no-dotfiles.mjs:113` — `run('git', ['init', '-q'], projectDir)`
- `scripts/integration/preset-lifecycle.mjs:206` — same

Neither goes through `realGitExec`, and neither is loaded by vitest, so neither guard applied. Under a hook, those `git init` calls write to the real `.git`.

## Why a shared helper

The var list now lives once in `scripts/lib/git-env.mjs` and is used by both scripts *and* `vitest.config.mjs`, which previously carried its own copy. A third copy is exactly the drift shape this repo already tracks elsewhere.

`src/base/git-identity.ts` keeps a separate copy by necessity: it is TypeScript compiled into `dist/` for consumers, while this one is loaded raw by `.mjs` scripts that run before any build. Both carry a comment pointing at the other.

## Deliberately not stripped

`GIT_AUTHOR_*` / `GIT_COMMITTER_*`. They do not redirect *which repository* git operates on, so removing them would change commit authorship for no benefit. The test asserts they survive.

## Verification

`tests/scripts/git-env.test.ts` — 3 tests, green. Asserts every var is removed, that unrelated and identity vars are untouched, and that `GIT_DIR`/`GIT_WORK_TREE` specifically stay in the list, so a future trim cannot silently reopen #500.

### DIFF
--- a/scripts/integration/no-dotfiles.mjs
+++ b/scripts/integration/no-dotfiles.mjs
@@ -17,6 +17,11 @@ import { execFileSync, spawnSync } from 'node:child_process'
 import fs from 'node:fs'
 import os from 'node:os'
 import path from 'node:path'
+import { stripAmbientGitEnv } from '../lib/git-env.mjs'
+
+// Before anything spawns git: this script runs `git init` in a temp project, and
+// under a hook GIT_DIR would redirect that into the real repo. See the helper.
+stripAmbientGitEnv()
 
 const REPO = process.cwd()
 const REAL_HOME = os.homedir()
--- a/scripts/integration/preset-lifecycle.mjs
+++ b/scripts/integration/preset-lifecycle.mjs
@@ -19,6 +19,11 @@ import { execFileSync } from 'node:child_process'
 import fs from 'node:fs'
 import os from 'node:os'
 import path from 'node:path'
+import { stripAmbientGitEnv } from '../lib/git-env.mjs'
+
+// Before anything spawns git: this script runs `git init` in a temp project, and
+// under a hook GIT_DIR would redirect that into the real repo. See the helper.
+stripAmbientGitEnv()
 
 const REPO = process.cwd()
 const CLI = path.join(REPO, 'dist', 'cli', 'index.js')
--- a/scripts/lib/git-env.mjs
+++ b/scripts/lib/git-env.mjs
@@ -18,14 +18,43 @@
  * TypeScript compiled into `dist/` for consumers, while this one is loaded
  * directly by `.mjs` scripts that run before any build.
  */
+/**
+ * This is `git rev-parse --local-env-vars` verbatim — git's own answer to
+ * "which variables bind a process to one specific repository", and what
+ * githooks(1) tells you to clear before touching a different one. Do not curate
+ * it by hand: an earlier version of this list was assembled from the vars that
+ * looked repository-ish and missed the `GIT_CONFIG*` family, two of which are
+ * exported by the maintainer's own shell.
+ *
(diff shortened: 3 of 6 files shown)


rtorcato (commented):

Strip placement (both `.mjs` scripts and `vitest.config.mjs`) is correct — before any git spawn, and `git-env.mjs` itself spawns nothing at import time. The two `AMBIENT_GIT_REPO_VARS`/`AMBIENT_REPO_VARS` copies are byte-identical (verified), so the TS/`.mjs` duplication is justified, not drifted. Excluding `GIT_AUTHOR_*`/`GIT_COMMITTER_*` is correct — they affect authorship, not repo targeting. Test isn't vacuous.

The list is missing vars. Ran `git rev-parse --local-env-vars` (git's own canonical "clear before touching a foreign repo" set, per `githooks(1)`): it includes `GIT_CONFIG`, `GIT_CONFIG_COUNT`, and `GIT_CONFIG_PARAMETERS`, none of which are in either copy. `GIT_CONFIG` redirects where the `config` subcommand reads/writes regardless of `-C`/cwd — exactly the `core.bare` write in `loop-guard.ts:197` and the `git -C cwd config` write in `git-hooks.ts:23`, which `preset-lifecycle.mjs`'s `setup` invocation exercises. `GIT_CONFIG_PARAMETERS` is how git itself propagates `-c` down to every child process, so it's genuinely present in real hook chains, not theoretical.

### Before merging
Add `GIT_CONFIG`, `GIT_CONFIG_COUNT`, `GIT_CONFIG_PARAMETERS` to `AMBIENT_GIT_REPO_VARS` (git-env.mjs) and `AMBIENT_REPO_VARS` (git-identity.ts). Confirmed via `git rev-parse --local-env-vars`. `GIT_CEILING_DIRECTORIES`/`GIT_DISCOVERY_ACROSS_FILESYSTEM`/`GIT_ATTR_NOSYSTEM`/`GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM` are absent from that same canonical list — correctly left out.

rtorcato (commented):

Verified all four points. (1) `expect(AMBIENT_GIT_REPO_VARS).toEqual(expect.arrayContaining(gitLocalEnvVars()))` is oriented correctly — the curated list is the received value, so it fails if any var `git rev-parse --local-env-vars` reports is missing from it; not vacuous. (2) The sorted-array equality test catches divergence regardless of order, and the two copies are genuinely identical (verified by hand: same 16 entries, same order, matching `git rev-parse --local-env-vars` plus `GIT_NAMESPACE`). (3) `GIT_NAMESPACE`'s comment is technically accurate — it scopes ref visibility within a repo rather than repo selection, so git's own list correctly excludes it, and keeping it here is a reasoned defense-in-depth call, not a misclassification. (4) No caller in this repo relies on inherited `-c` overrides via `GIT_CONFIG_PARAMETERS`/`GIT_CONFIG_COUNT` — every real git invocation here targets an explicit `cwd`/`-C`, consistent with the module's stated intent, so the wider strip doesn't remove anything relied upon.

### Before merging
Nothing.

PR #556, opened 2026-08-26
feat(lockfile): add requiredSkills and mcp.recommended with advisory doctor checks

### LINKED ISSUE #533: requiredSkills in .repo-tooling.json + doctor staleness check for installed ai skills

A repo whose workflows depend on the ai-* skills has no way to say so, and nothing detects a stale installed copy. Observed 2026-08-26: the running `~/.claude/skills/ai-issue-loop` was three merged PRs behind `main` — missing the decision-comment security gate (#526) and the decay rule (#528) — and ran without complaint. The #480 skill fork was a month of the same failure mode. Absence fails loudly; staleness fails silently, and staleness is the one that hurts.

## What

Add `requiredSkills` to `.repo-tooling.json` (e.g. `["ai-issue-loop", "ai-workflow"]`, validated against `SHIPPED_SKILLS`), plus a doctor check:

- For each listed skill, find `~/.claude/skills/<name>/SKILL.md` (honouring `--skills-dir`) and compare its stamped `repo-tooling-hash` against the shipped asset's hash — the stamps `fix claude-skills` already writes (#523). Missing **or** hash-mismatched ⇒ report, with the hint to run `fix claude-skills`.
- Severity is `optional-missing`, never `drift` — the check probes the machine, not the repo, and a contributor without Claude must not fail doctor. Skip the check entirely when the lockfile has no `aiLoop` key: that key is already the "this repo uses the pipeline" signal.
- A locally modified skill (hash differs from *every* shipped version) reports as `modified`, mirroring the assets model (#428) — `fix claude-skills` already refuses those without `--force-skills`.

**Check and hint only — the fixer is never run automatically from repo config.** A committed file directing writes into `~/` is the shape of a supply-chain attack even when the content is benign; doctor says stale, the human runs `fix claude-skills`.

## Required MCP: declined — superseded by recommended MCP (#534)

Claude Code natively supports a committed repo-scoped `.mcp.json` with its own first-use consent prompt — that *is* the required-MCP mechanism, and re-declaring it in the lockfile would be a second copy of one fact. Auto-installing MCP servers from repo config is a supply-chain vector (they execute code). None of the ai-* skills need MCP today. Decided here so it isn't re-litigated; revisit only if a shipped skill grows an MCP dependency. **Update:** reshaped as *recommended* (advisory names + importance + why, no executable config) in #534 — the requirement version stays declined.

## Sequencing

- Depends on #529 — the field should land in a published schema, not ahead of one.
- Feeds the #532 design: `requiredSkills` is a "fact about this repo" per that issue's fleet-policy/repo-fact split.
- Docs: document the field in the AI-setup guide and the lockfile reference page (#529's deliverable).

## Dogfood

Set the field in **this repo's own** `.repo-tooling.json` in the same PR: `"requiredSkills": ["ai-issue-loop", "ai-workflow", "ai-issue", "ai-loop-status"]` — the four in `SHIPPED_SKILLS`, which this repo's own pipeline runs on. The new doctor check then guards the exact staleness that bit on 2026-08-26 (installed skills three PRs behind). Note the self-repo guard: `doctor` needs `REPO_TOOLING_ALLOW_SELF=1` here, which CI already uses (#273).

### LINKED ISSUE #534: Add a "mcp": {"recommended": []} field to .repo-tooling.json — advisory list with importance, never required

Follow-on to #533, which declined *required* MCP. Reshaped on review: **recommended**, never required — advisory metadata, not an install directive.

## What

Add `mcp.recommended` to `.repo-tooling.json`:

```json
"mcp": {
  "recommended": [
    { "name": "pencil", "importance": "important", "why": "edits the .pen design files under design/" }
  ]
}
```

- `name` — the MCP server name as it would appear in `.mcp.json`.
- `importance` — enum, three levels: `nice-to-have` | `important` | `critical`. Signals how much of the repo's workflow assumes the server, nothing more.
- `why` — one line, free text. The thing `.mcp.json` structurally cannot say.

Doctor check, **informational only**: for each recommended entry, report whether `.mcp.json` at the repo root declares a server of that name. Missing ⇒ info-level line quoting the importance and why — never `drift`, never a CI failure, and no probing of user-scoped MCP config (machine-private).

## The line that must hold

**Names, importance, and reasons only — no executable config.** No `command`, `args`, or `env` in the lockfile, and no fixer that installs or enables an MCP server from it. Actual server config lives in the native `.mcp.json`, which carries Claude Code's own first-use consent prompt. The lockfile says *what and why*; `.mcp.json` says *how*; the human says *whether*.

## Sequencing

- Depends on #529 — the field lands in the published schema (the `importance` enum is exactly what schema validation is for).
- Sibling of #533 (`requiredSkills`) and a "fact about this repo" per #532's split.
- Docs: lockfile reference page + AI-setup guide.

rtorcato:

Two new `.repo-tooling.json` fields, batched because they add to the same schema block and #534 is an explicit follow-on to #533. Both are facts the repo states about itself, never instructions anything acts on.

## `requiredSkills` (#533)

Lists the shipped agent skills a repo's workflows depend on, validated against `SHIPPED_SKILLS` by the published schema and again at runtime.

It exists for **staleness**, not absence. A missing skill fails loudly the moment something reaches for it; a copy several releases behind runs to completion without complaint. The new `Required skills` check compares each listed skill's installed `SKILL.md` — its stamped `repo-tooling-hash`, honouring `--skills-dir` — against the shipped asset's, and reports missing, stale, or (mirroring the assets model) matching no shipped version at all.

Gated on `aiLoop`: that key is already the "this repo uses the pipeline" signal, so a repo without it gets no line rather than an empty one.

## `mcp.recommended` (#534)

`{ name, importance, why }` per entry, `importance` an enum of `nice-to-have | important | critical`. The `Recommended MCP` check reports which of those names the repo-scoped `.mcp.json` does not declare, and stops there. User-scoped MCP config is machine-private and is not probed.

## The line that holds for both

Neither check ever returns `drift` or `missing` — they probe the *machine*, not the repo, so a contributor without Claude installed does not fail `doctor`, and neither can fail CI.

Neither has a fixer, and neither is in `FIX_TARGETS` (there's a comment there saying why). Committed repo config that directs writes into `~/`, or that installs a code-executing MCP server, is the shape of a supply-chain attack even when the content is benign. So the MCP entry shape carries no `command`/`args`/`env`, `doctor` only ever *names* `fix claude-skills`, and the human runs it. `.mcp.json` keeps Claude Code's own first-use consent prompt.

## Dogfood

This repo now declares all four shipped skills. Running `doctor` against it immediately caught the real thing:

```
Required skills | optional-missing | ai-issue-loop is stale — installed 3.22.1, this package ships 3.25.0; ...
```

No `mcp` block is set here — the issue's `pencil` entry is an example, not a default, and this is a public library.

## Notes for review

- `lockfile.ts` now imports `SHIPPED_SKILLS`, which closes a second import cycle (via `copy-preset.ts`). Safe for the same reason the existing `CONFIG_SCHEMA` one is: it's read inside `lockfileSchema()` at call time, not at module eval. The comment there says so.
- The hand-rolled validator in `lockfile-schema.test.ts` gained `type: 'array'` + `items` — the `ponytail:` note said to reach for ajv the day arrays appeared; six lines is still cheaper than the dependency.
- `apps/docs/static/schemas/lockfile.json` is regenerated with `pnpm schema:generate`, not hand-edited.
- Verified: `check`, `typecheck`, and the full suite (67 files, 1130 passed) all green.

Closes #533
Closes #534

### DIFF
--- a/.repo-tooling.json
+++ b/.repo-tooling.json
@@ -33,6 +33,7 @@
   "aiLoop": {
     "agentUser": "rtorcato-bot"
   },
+  "requiredSkills": ["ai-issue-loop", "ai-workflow", "ai-issue", "ai-loop-status"],
   "writtenBy": "@rtorcato/repo-tooling@3.11.0",
   "writtenAt": "2026-08-26T14:58:22.032Z"
 }
--- a/apps/docs/docs/guides/ai-setup.md
+++ b/apps/docs/docs/guides/ai-setup.md
@@ -98,6 +98,58 @@ against your repo's own remote (`gh api repos/{owner}/{repo}/assignees/<user>`)
 and reports drift when it isn't assignable. Absent field ⇒ `ok`, not
 applicable.
 
+## `requiredSkills`: catching a *stale* skill, not just a missing one
+
+A repo whose workflows depend on the shipped skills can say so:
+
+```json
+{
+  "requiredSkills": ["ai-issue-loop", "ai-workflow", "ai-issue", "ai-loop-status"]
+}
+```
+
+Absence isn't the interesting failure — a skill that isn't installed fails loudly
+the moment something reaches for it. Staleness is. An installed copy several
+releases behind runs happily to completion while missing whatever the newer
+releases added, and nothing complains. So `doctor` compares each listed skill's
+installed `SKILL.md` — its stamped `repo-tooling-hash`, under `~/.claude/skills`
+or wherever `--skills-dir` points — against the copy this package ships, and
+reports anything missing, stale, or matching no shipped version at all.
+
+Two rules keep the field safe to commit:
+
+- **It never fails your build.** The check probes your machine, not the repo, so
+  it reports "not configured" and never `drift` or `missing` — a contributor
+  without Claude installed doesn't fail this repo's `doctor`. It's also skipped
+  entirely unless the file has an `aiLoop` key.
+- **It only ever hints.** `doctor` names `fix claude-skills`; running it is
+  yours. Committed repo config that directs writes into your home directory is
+  the shape of a supply-chain attack even when the content is benign.
+
+## `mcp.recommended`: names and reasons, never an install directive
+
+The same file can name the MCP servers a repo's workflow assumes — and only
+name them:
+
+```json
+{
+  "mcp": {
+    "recommended": [
+      { "name": "some-server", "importance": "important", "why": "edits the design files under design/" }
+    ]
+  }
+}
+```
+
+`importance` is `nice-to-have`, `important` or `critical`; `why` is the one line
+`.mcp.json` structurally cannot carry. `doctor` reports which of these names
+`.mcp.json` doesn't declare, informationally, and stops there.
+
+There's deliberately no `command`, `args` or `env`, and no fixer. MCP servers
+execute code, so real config belongs in `.mcp.json` — the file below, which
+carries Claude Code's own first-use consent prompt. The lockfile says *what and
+why*, `.mcp.json` says *how*, and you say *whether*.
+
 ## Worktrees: symlink `node_modules` instead of reinstalling it
 
 A Claude Code worktree starts empty, so an agent working one pays a full
--- a/apps/docs/docs/reference/repo-tooling-json.mdx
+++ b/apps/docs/docs/reference/repo-tooling-json.mdx
@@ -79,6 +79,57 @@ enum-valued records — see the schema for their exact shapes.
 
 See the [AI issue loop guide](../guides/ai-issue-loop.md) for how this is used.
 
+## `requiredSkills` — agent skills this repo depends on
+
+A list of the skills this package ships (`ai-issue-loop`, `ai-workflow`, `ai-issue`,
+`ai-loop-status`) that the repo's workflows assume. It exists for **staleness**, not absence:
+a skill that isn't installed fails loudly the moment something reaches for it, while a copy
+three releases behind runs to completion without complaint.
+
+```json
+"requiredSkills": ["ai-issue-loop", "ai-workflow"]
+```
+
+`doctor` compares each listed skill's installed `SKILL.md` — its stamped `repo-tooling-hash`,
+under `~/.claude/skills` or wherever `--skills-dir` points — against the hash of the copy this
+package ships, and reports anything missing, stale, or matching no shipped version at all.
+
+Two rules make it safe to commit:
+
+- **Never `drift`, never `missing`.** The check probes your machine, not the repo, so a
+  contributor with no Claude installed must not fail its `doctor`. It only ever reports
+  "not configured", which leaves the exit code alone. It is also skipped entirely unless the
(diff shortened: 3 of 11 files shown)


rtorcato (commented):

Reviewed the trust-boundary question directly: this diff makes a committed, in-repo file declare facts about the user's machine (`requiredSkills`, `mcp.recommended`).

- **No auto-actuation.** `fix-targets.ts` deliberately omits both new checks from `FIX_TARGETS`, with a comment explaining why — confirmed no fixer is wired to either. Both checks return `optional-missing` at worst, never `drift`/`missing`, so nothing can fail CI or a build.
- **Path handling is sound.** `checkRequiredSkills` filters `names` against the hardcoded `SHIPPED_SKILLS` allowlist (`src/cli/generators/claude-skills.ts:23`) and returns early on any unknown name — *before* `claudeSkillStatus` ever builds a path or calls `readFile`. Since only whitelisted literal skill names ever reach `path.join(dir, name, 'SKILL.md')`, arbitrary/traversal names from the repo config cannot influence path construction, independent of the JSON-schema enum.
- **No leakage.** Reported detail strings are limited to skill names (from the fixed allowlist) and version-stamp strings; `checkRecommendedMcp` only reads the repo-local `.mcp.json` (never user-scope config) and echoes back `name`/`importance`/`why` values that originated in the committed lockfile itself — nothing from `~` is read or echoed.
- **Schema (`mcp.recommended`)** uses `additionalProperties: false` and a fixed `{name, importance, why}` shape, so `command`/`args`/`env` are rejected by the schema — verified by the new schema test asserting an entry with `command` fails validation. No new dependencies introduced.

### Before merging
Nothing.

rtorcato (commented):

Both checks always return `optional-missing` on their non-ok paths — never `drift`/`missing` — matching #533/#534's stated line. `requiredSkills` is gated on `lock?.aiLoop && lock.requiredSkills?.length` (doctor.ts:408), so a repo without `aiLoop` gets no line. `--skills-dir` is threaded through to `claudeSkillStatus`. The missing/stale/modified partitions are disjoint (`needsInstall` implies pristine, so a hash-mismatched copy only ever lands in `modified`), and all four outcomes (missing, stale, modified, ok) plus the unknown-skill-name path are exercised in `tests/base/required-skills.test.ts`. `requiredSkills` is validated against `SHIPPED_SKILLS` both in the schema enum and at runtime in `checkRequiredSkills`. `mcp.recommended` only ever names/reports; the schema's `additionalProperties: false` on the entry blocks `command`/`args`/`env`, and that's asserted directly in `lockfile-schema.test.ts`. No `@rtorcato`- or `pencil`-specific value leaked into shipped code, defaults, or docs — the example was genericized to `some-server`, and this repo's own `.repo-tooling.json` sets no `mcp` block. `apps/docs/static/schemas/lockfile.json` matches the `lockfileSchema()` output, and the self-validation test still passes since the dogfooded `requiredSkills` values are all in `SHIPPED_SKILLS`.

### Before merging
Nothing.

PR #552, opened 2026-08-26
test(lockfile): validate this repo's own .repo-tooling.json against lockfileSchema()

rtorcato:
🤖 *Opened by Claude (AI agent) on the owner's behalf.*

#539 published the lockfile JSON Schema and gates the committed copy under `apps/docs/static/schemas/` against staleness — but nothing asserted that *this repo's own* `.repo-tooling.json` conforms to it. The repo that publishes the schema must never carry a lockfile that fails it.

## What this adds

Two tests in the existing `tests/cli/utils/lockfile-schema.test.ts`:

1. **`validates against lockfileSchema()`** — reads the repo's real `.repo-tooling.json` and walks it against the schema object returned by `lockfileSchema()`.
2. **`rejects a drifted lockfile`** — mutates an in-memory copy five ways and asserts each produces at least one error: missing required key, stray top-level key, wrong scalar type, unknown `aiLoop` key, out-of-enum `config.bundler`. This exists so the assertion above can never quietly go vacuous.

The walker covers the keyword subset the schema actually uses — `type` (object/string/integer/boolean), `enum`, `minLength`, `required`, `properties`, `additionalProperties` (`false` or a sub-schema) — and recurses, so the embedded `ProjectConfig` and `aiLoop` subtrees are validated too, not just the top level.

## Why hand-rolled instead of ajv

The repo has no JSON Schema validator today, and the schema leans on six keywords. That is not enough schema to justify a new devDependency and the lockfile churn. A `ponytail:` comment names the ceiling — `format: date-time`, arrays and composition keywords are unchecked because none appear in the schema — and says to reach for ajv the day one does.

Only refactor to existing code: `schemasDir` became a `repoRoot` constant, since the new tests need the repo root too.

## Verification

- `vitest run` — 1075 passed, 1 expected fail (pre-existing), 18 skipped.
- `biome check` on the touched file — clean.
- `tsc --noEmit` — clean.
- Confirmed the new check is non-vacuous via the drift test, which fails the walker on every mutation.

Commit type is `test:` deliberately — it cuts no release.

Closes #540

PR #561, opened 2026-08-26
feat(doctor): record deliberate deviations as declared exceptions in .repo-tooling.json

### LINKED ISSUE #558: doctor has no way to record a deliberate deviation, so real drift hides among permanent false findings

Split out of #532, which conflates two separable things. This is the small, GitHub-agnostic half that pays for itself now; #532 keeps the large "declarative settings with per-forge appliers" refactor, whose justification was GitLab parity (#517, closed).

## The gap

`doctor` reports every check against a fleet standard, and a repo has no way to say *"we deviate here, deliberately, for this reason"*. #431 said the same in one line — *"doctor cannot record opt-outs"* — and was closed.

Measured on this repo, 2026-08-26, one run. Three findings that are **correct for a consumer and wrong here**, because this repo *is* the package:

- `package.json — drift: @rtorcato/repo-tooling not in dependencies or devDependencies`
- `TypeScript — missing: no tsconfig.json found` (it lives at `src/cli/tsconfig.json`)
- `Vitest — drift: vitest.config.mjs found but does not import "@rtorcato/repo-tooling/vitest/config"` (it imports `./tooling/vitest/vitest.config.mjs`, the local source of that very preset)

They recur on every run forever. Meanwhile a **real** finding sits in the same report and is easy to miss among 16 `➖` lines:

```
⚠️  Release gate — drift
    ci.yml `release` publishes with no `environment:` and the repo has no
    `release` environment — anything that lands on the default branch publishes
```

That is the exact gate `ai-issue-loop` Pass 1 probes before permitting an unattended merge.

The cost is not the noise itself, it is that **a permanently-noisy report trains the reader to skim** — the same failure mode the loop's own comment budget and `ai-notes` rules exist to prevent.

## Done when

A repo can declare an exception in `.repo-tooling.json`, keyed by `CheckResult.check` — a stable string already used as an identifier by `FIX_TARGETS` (`src/cli/commands/fix-targets.ts:53`) — with a **mandatory non-empty reason**, schema-enforced.

Behaviour, each part load-bearing:

- **Declared exceptions are shown, never hidden.** The check reports as declared, printing its reason, and the summary carries a `declared: N` count. Suppressing a finding into silence would make this a mute button, which is the failure mode to avoid; the point is to move a deviation from *invisible* to *argued in a PR*.
- **Exit code**: a declared exception does not fail the run. That is the practical purpose.
- **A stale exception is itself a finding.** If an exception names a check that does not exist — typo, or a check renamed or removed — report it. Otherwise a rename silently un-suppresses a finding, or a typo silently does nothing, and both are invisible.
- **Exceptions are per-repo by construction** — an exception for `TypeScript` here is legitimate and would hide a real problem in a consumer, which is fine, because it lives in that repo's own file.
- The published lockfile schema must gain the field so it autocompletes and validates from day one, and the self-validation test added in #552 must stay green.

A reviewer can check this against four cases: a check with no exception behaves exactly as today; a check with one reports `declared` with its reason and does not fail the run; an exception naming an unknown check is reported; and the schema rejects an exception with a missing or empty reason.

## Deliberately out of scope for v1

**Expiry.** An exceptions surface can rot into a permanent mute list, and the strong guard is a deadline — the same shape as the loop's 30-day `ai-suggested` decay. It is not in v1 because a mandatory reason plus visible `declared: N` is the cheap 80%, and expiry needs a decision about what expiring *does* (warn? fail? auto-remove?) that should not be guessed at. If the count starts growing, that is the signal to add it.

## Files

- `src/cli/utils/lockfile.ts` — the type and schema
- `src/base/types.ts` — `CheckResult`, if the status set needs a member
- wherever `doctor` renders results and computes its exit code
- `apps/docs/docs/reference/repo-tooling-json.mdx`

rtorcato:

Closes #558

## What

A repo can now declare a deliberate deviation in `.repo-tooling.json`:

```json
"exceptions": {
  "TypeScript": "this repo is the package; the tsconfig lives at src/cli/tsconfig.json"
}
```

Keyed by the doctor check name (`CheckResult.check`), with a mandatory non-empty reason enforced by the published JSON Schema (`minLength: 1`).

## Behaviour

- **Shown, never hidden**: the check reports as a new `declared` status with its reason appended to the detail, and the summary line carries a `declared: N` count.
- **Exit code**: `declared` feeds neither `drift` nor `missing`, so a declared exception no longer fails the run.
- **A stale exception is itself a finding**: an entry naming a check the run doesn't know (typo, renamed, removed) is reported as `drift` under a `Declared exceptions` check, so it can't silently do nothing or silently un-suppress.
- **Bulk `fix` skips declared checks**; a targeted `fix <target>` still applies (naming the fixer is the same explicit override the declined-in-lock path gets).
- Published lockfile schema regenerated (`pnpm schema:generate`); the #552 self-validation test stays green.

## Reviewer's four cases (all tested)

1. No exception → behaves exactly as today (existing suite, unchanged).
2. Exception on a failing check → `declared` with reason, not counted toward the failing exit code.
3. Exception naming an unknown check → reported as drift.
4. Schema rejects a missing/empty/non-string reason.

Deliberately out of scope per the issue: expiry, and dogfooding exceptions into this repo's own `.repo-tooling.json` (a one-line follow-up once this lands).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -18,7 +18,7 @@ Every command supports `--json` and a non-interactive mode. Combine with `--yes`
 | `setup --config <path>` | ✅ | `--dry-run` only | Scaffold with a full `ProjectConfig` JSON file. See `setup --config-schema`. |
 | `setup --config-schema` | ✅ | ✅ (JSON Schema) | Print the JSON Schema for `ProjectConfig`. Use to validate configs before scaffolding. |
 | `setup --dry-run` | ✅ | ✅ | Print resolved config + file list without writing. Pair with `--preset` or `--config`. |
-| `doctor --json` | ✅ | ✅ | Audit a project. Returns `{ directory, results: [{ check, status, detail, hint? }] }`. Status: `ok` / `drift` / `missing` / `optional-missing`. |
+| `doctor --json` | ✅ | ✅ | Audit a project. Returns `{ directory, results: [{ check, status, detail, hint? }] }`. Status: `ok` / `drift` / `missing` / `optional-missing` / `declared`. |
 | `fix --json --yes` | ✅ | ✅ | Walk every doctor finding, apply fixers. Returns `FixActionRecord[]` with `status: applied | dry-run | skipped | already-ok | unsupported`. |
 | `fix <target> --json --yes` | ✅ | ✅ | Apply one fixer. Targets from `list --json`. |
 | `fix --dry-run` | ✅ | ✅ | Print what each fixer would write without writing. Combine with `--json`. |
--- a/apps/docs/docs/guides/for-ai-agents.md
+++ b/apps/docs/docs/guides/for-ai-agents.md
@@ -82,7 +82,7 @@ npx @rtorcato/repo-tooling doctor --json -d ./existing-repo
 }
 ```
 
-Status values: `ok`, `drift` (file exists but doesn't extend the preset), `missing` (required file absent), `optional-missing` (optional convention absent).
+Status values: `ok`, `drift` (file exists but doesn't extend the preset), `missing` (required file absent), `optional-missing` (optional convention absent), `declared` (a real deviation `.repo-tooling.json` `exceptions` records on purpose, with its reason — don't fix it).
 
 ### `fix` — apply fixers for items doctor flagged
 
@@ -174,7 +174,7 @@ pnpm install
 DOCTOR=$(npx @rtorcato/repo-tooling doctor --json -d .)
 
 # 2. Parse, decide whether to fix everything or specific items
-#    (For agents: filter results where status !== "ok" && status !== "optional-missing")
+#    (For agents: filter results where status !== "ok" && status !== "optional-missing" && status !== "declared")
 
 # 3. Apply all fixable findings
 npx @rtorcato/repo-tooling fix --yes --json -d .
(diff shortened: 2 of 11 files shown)


rtorcato (commented):

Traced the exception path end to end: `exceptions[r.check]` (doctor.ts) only flips `status` and string-concats `reason` into `detail` for display; `Object.keys(exceptions)` unknown-check loop only builds a static drift message. Neither touches `fix.ts`, which filters purely on `status !== 'declared'` — no fixer lookup, no path construction, no write ever reads the reason or check-name strings. Reason is rendered via `console.log`/chalk as plain text, never `eval`'d, exec'd, or used in a template that executes. Schema enforces `minLength: 1`, string-only. No new dependencies, no secrets, no shell/SQL construction anywhere in the diff. `.repo-tooling.json` is repo-owned config, not attacker-supplied issue input, so the trust boundary holds.

### Before merging

Nothing.

rtorcato (commented):

Traced `applyExceptions` in `src/cli/commands/doctor.ts` against all four reviewer cases from #558: no exception leaves results untouched; a declared exception on a failing check becomes `declared` with the reason appended and drops out of the `drift`/`missing` exit-code count; an exception naming an unknown check appends a `Declared exceptions` drift finding; the schema's `minLength: 1` rejects empty/non-string reasons (confirmed in `lockfile-schema.test.ts`).

Bulk `fix` correctly excludes `declared` (`fix.ts`), while the targeted `fix <target>` path runs before that filter, so an explicit override still applies as documented.

Ran the suite in the existing `ai-558-declared-exceptions` worktree: all 261 tests in `lockfile-schema.test.ts`, `doctor.test.ts`, and `fix.test.ts` pass, including the #552 self-validation test against the actual `.repo-tooling.json` (unaffected since `exceptions` isn't dogfooded into it yet, as the PR states). Re-ran `pnpm run schema:generate` — zero diff, so the published schema is in sync.

Docs (`AGENTS.md`, `for-ai-agents.md`, `repo-tooling-json.mdx`, `SKILL.md`) all updated consistently with the new `declared` status.

### Before merging

Nothing.

PR #562, opened 2026-08-26
feat(doctor): scan recent commit history for placeholder or generated author emails

### LINKED ISSUE #557: doctor checks the git identity that is set, never the commits already made with a bad one

`checkGitIdentity` (`src/base/git-identity.ts:174`) reads exactly one thing — `git config --get user.email` — and classifies it. That catches a bad identity *going forward*. Nothing in `src/` ever looks at the author of a commit already made: `grep -rn "GIT_AUTHOR\|%ae" src/` returns nothing.

So the damage #327 describes — 703 mis-attributed commits across 23 repos, this repo worst at 79 of 397 — is invisible to `doctor`. It reports the identity you will use next, never the one you already used.

## Observed, 2026-08-26

The `GIT_DIR` leak fixed in #555 wrote `user.email test@example.com` / `user.name Test` into this repo's **live** `.git/config`. The source was `tests/cli/commands/loop-guard.test.ts:35` setting a throwaway identity in a temp project, redirected into the real repo by the ambient env vars.

`doctor` did flag it, correctly:

```
➖ Git identity — not configured
   git user.email is a placeholder that can never receive mail: test@example.com
```

But only because someone happened to run `doctor` while the config was still poisoned. Any commit authored during that window would carry `test@example.com` permanently, and once the config was cleaned **nothing would ever report those commits again**. This repo escaped with zero — the squashes were authored server-side by GitHub — but that was luck, not detection.

The prevention landed in #555. The detection did not, and the two are independent: a repo can be poisoned by any mechanism (a stale `--local` override, a copied `.git/config`, a hook, a different tool) long after this particular leak is closed.

## Done when

`doctor` reports commits in recent history whose author email classifies as `placeholder` or `generated`, reusing the existing pure `classifyGitEmail` — do not write a second classifier, the two would drift.

Constraints, each load-bearing:

- **`optional-missing` severity, never `drift`/`missing`.** Same reasoning the existing check already documents: history is not something the repo's author can fix by editing the repo, and failing CI over it would be permanent noise. It must not change the exit code.
- **Skipped on CI**, like `checkGitIdentity` already is.
- **Bounded scan.** `doctor` runs often; do not walk an unbounded history on every invocation. Cap it (a commit count or a date window) and **say in the output what was scanned**, so "0 found" cannot be read as "history is clean" when only part was examined.
- **Report the count and a sample**, not every offending hash — the hint is what matters, and it should be honest that these are not retroactively fixable without a history rewrite (the existing `HINT` already says why: the address can never be verified as a secondary email).
- A shallow clone has no history to scan. Detect that and say so rather than reporting a false all-clear.

A reviewer can check this by reading the new check against three cases: a repo whose recent commits are all fine, one with a placeholder-authored commit in range, and a shallow clone.

## Files

- `src/base/git-identity.ts` — `classifyGitEmail` is already exported and pure, so the classification is reusable as-is.
- Wherever `checkGitIdentity` is registered for `doctor`, alongside it.

Note this is detection only — **no fixer.** There is nothing to fix: rewriting history is not something a tool should offer to do unprompted, and the existing hint already explains why the commits cannot be re-linked.

rtorcato:

Closes #557

`checkGitIdentity` reads only the identity the *next* commit will use; commits already made with a bad one were invisible to doctor. This adds the retrospective half as a new base check, **Git author history**:

- Scans the last 200 commits' author emails through the **same** exported `classifyGitEmail` — no second classifier to drift.
- A finding is `optional-missing`, never `drift`/`missing` — exit code untouched, and skipped on CI, matching the existing check's reasoning.
- The output always names how many commits were scanned, so "0 found" cannot be read as "all history clean".
- Reports the count plus a 3-commit sample; the hint is honest that these are unfixable without a history rewrite.
- A shallow clone is detected via `git rev-parse --is-shallow-repository` and reported as unscannable rather than a false all-clear.
- Detection only: no fixer, deliberately absent from `FIX_TARGETS`. Composes with #561's declared exceptions automatically via its check name.

Tests cover the issue's three reviewer-checkable cases (clean history, placeholder in range, shallow clone) plus empty-history and CI-skip.

### DIFF
--- a/src/base/git-identity.ts
+++ b/src/base/git-identity.ts
@@ -183,3 +183,79 @@ export async function checkGitIdentity(dir: string, exec?: GitExec): Promise<Che
 		hint: HINT,
 	}
 }
+
+const HISTORY_CHECK = 'Git author history'
+
+/**
+ * Bounded so doctor stays cheap on every invocation; the output always names
+ * how many commits were actually examined so "0 found" is never read as "all
+ * history is clean".
+ */
+export const HISTORY_SCAN_LIMIT = 200
+
+/** How many offending commits to name; the count carries the rest. */
+const HISTORY_SAMPLE = 3
+
+const HISTORY_HINT =
+	'These commits are already made: the address can never be verified as a secondary email, so they cannot be re-linked to a forge account without a history rewrite (#327). Fix the identity going forward (see the Git identity check) and treat the history as recorded loss.'
+
+/**
+ * The retrospective half of `checkGitIdentity` (#557): that check reads the
+ * identity the *next* commit will use, this one scans recent history for
+ * commits already made with a bad one. Same classifier — `classifyGitEmail` —
+ * so the two can never drift. Same STATUS reasoning too: history is not
+ * something the repo's author can fix by editing the repo, so a finding is
+ * `optional-missing`, never `drift`/`missing`, and the exit code is untouched.
+ * Detection only — there is deliberately no fixer, because the only "fix" is a
+ * history rewrite no tool should offer unprompted.
+ */
+export async function checkGitIdentityHistory(dir: string, exec?: GitExec): Promise<CheckResult> {
+	if (!(await fs.pathExists(path.join(dir, '.git')))) {
+		return { check: HISTORY_CHECK, status: 'ok', detail: 'not a git repository' }
+	}
+	if (process.env.CI) {
+		return { check: HISTORY_CHECK, status: 'ok', detail: 'skipped on CI' }
+	}
+
+	const git: GitExec = exec ?? ((args) => realGitExec(args, dir))
+	// A shallow clone has almost no history; scanning its stub and reporting
+	// "0 found" would be a false all-clear.
+	if ((await git(['rev-parse', '--is-shallow-repository'])) === 'true') {
+		return {
+			check: HISTORY_CHECK,
+			status: 'ok',
+			detail: 'shallow clone — commit history is not available to scan',
+		}
+	}
+	// %h %ae: abbreviated hash + author email, one commit per line.
+	const log = await git(['log', `-${HISTORY_SCAN_LIMIT}`, '--format=%h %ae'])
+	if (log === null || log === '') {
+		return { check: HISTORY_CHECK, status: 'ok', detail: 'no commits to scan' }
+	}
+	const commits = log.split('\n').map((line) => {
+		const sp = line.indexOf(' ')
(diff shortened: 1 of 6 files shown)


rtorcato (commented):

Reviewed `checkGitIdentityHistory` against the untrusted-repo threat model.

- Spawn: both new calls (`rev-parse --is-shallow-repository`, `log -${HISTORY_SCAN_LIMIT} --format=%h %ae`) go through the existing `realGitExec`/`repoScopedEnv` path — array args, `shell:false`, ambient `GIT_*` stripped, 5s timeout. `HISTORY_SCAN_LIMIT` (200) and the format string are hardcoded constants, never built from repo content.
- Output: only `%h %ae` (hash + email) is read from the untrusted repo; no commit message or other free text is echoed. Sample is bounded to `HISTORY_SAMPLE = 3`, and every branch states how many commits were scanned.
- Shallow clones are detected and reported rather than yielding a false all-clear.
- No new dependencies; env handling is inherited unchanged from `checkGitIdentity`.

Info only, not blocking: `commits.map` splits each log line on the first space (`indexOf(' ')`) to separate hash from email. A hand-crafted commit object with a space in the author email (legal RFC 5322 quoted-local-part) would misparse into a wrong hash/email pair — cosmetic, no injection or code-exec path, since the result only ever reaches display strings.

### Before merging
Nothing.

rtorcato (commented):

Matches #557: reuses `classifyGitEmail` (no second classifier), `optional-missing`-only and CI-skipped, bounded scan with the scanned count named in every branch, shallow clone reported as unscannable rather than a false all-clear, no fixer/no `FIX_TARGETS` entry, and registers in `doctor.ts` as a plain check.

The placeholder test builds a bad-email commit via a fake `GitExec` and asserts on the exact count/hint text, so it would catch a broken implementation, not just pass vacuously.

### Before merging

Nothing.

rtorcato (commented):

Re-reviewed after the fix round. Delta is exactly `3e16a08`: adds `'Git author history'` to the expected-checks fixtures in `tests/languages/{python,swift,perl}/fixers.test.ts`. No production code changed — prior PASS on the git-spawn construction, output bounding, and env handling in `src/base/git-identity.ts` still holds.

### Before merging
Nothing.

rtorcato (commented):

Fix round only adds `'Git author history'` to the three language fixer-coverage lists (`tests/languages/{python,swift,perl}/fixers.test.ts`), immediately after `'Git identity'` — correct grouping, no logic touched. Prior PASS stands.

### Before merging

Nothing.

Issue #559, opened 2026-08-26
Design: .repo-tooling.json is two documents sharing one file — a tool-written record and a human-written ruleset


rtorcato:
🤖 *Filed by Claude (AI agent) on the owner's behalf.*

**Design issue — the decision is the work. Not for the AI queue.** Sits underneath both #532 and #558; deciding it first makes both cleaner.

## The tension

The file is named a lockfile (`$schema: …/lockfile.json`) and half of it behaves like one — a **record**, written by the tool, describing what happened:

- `config` — what `setup` chose
- `assets` — content hashes of copied files
- `writtenBy` / `writtenAt` — provenance stamps, **required** by the schema (`src/cli/utils/lockfile.ts:121`), written from `packageJson.version` (`:271`), and reported by `doctor` as authority (`doctor.ts:202`: *"written by …"*)

The other half is a **ruleset** — intent, written by humans, reviewed in PRs, the owner's stated direction that this file be "the rules of the repo":

- `aiLoop.agentUser` (#525)
- `requiredSkills` (#556)
- `mcp.recommended` (#556)
- declared exceptions, incoming (#558)

## Evidence the tension is already producing lies

This repo's own file, right now: `"writtenBy": "@rtorcato/repo-tooling@3.11.0"` — while containing `requiredSkills`, a field that exists only in the 3.26.0 schema. #556 (`acd35da`) edited the file by hand in a PR and correctly did *not* touch the stamp, because the tool didn't write that change. The stamp is now false the moment anyone edits the rules half, and every rules-field added makes it falser. `doctor` still prints it as provenance.

This is not a bug to patch (auto-bumping the stamp on hand edits would make it lie in the other direction). It is the symptom of two documents with different lifecycles — *record: tool-written, stamp meaningful* vs *rules: human-written, stamp meaningless* — sharing one schema, with nobody having decided that.

## Options

| Option | Shape | Cost |
|---|---|---|
| **A. The file is the rules** | Rename the mental model (schema title, docs); `writtenBy`/`writtenAt` demoted to optional or scoped to the `config`/`assets` subtree they actually describe | Schema change; doctor's "written by" line reworded |
| **B. Split the file** | `.repo-tooling.json` stays the tool-written lockfile; rules move to a second committed file | Two files to teach; migration for `aiLoop`/`requiredSkills` |
[...]

PR #565, opened 2026-08-27
feat(rules): compare repo rules against a reference repo, and seed setup from one

### LINKED ISSUE #563: Rules are per-repo: doctor compares them against a reference repo, and setup can start from one

Successor to #560, implementing the decision recorded there: repo rules are unique to each repo and live in its `.repo-tooling.json`; there is never a guideline package. Sharing a guideline means **pointing at a reference repo**. Two capabilities follow, sharing one fetcher:

## 1. Compare rules against a reference repo

`doctor` (a flag such as `--rules-from owner/repo`, or a committed field — implementer's judgement, but document the choice in the PR) fetches that repo's `.repo-tooling.json` via `gh api repos/{owner/repo}/contents/.repo-tooling.json` and reports where this repo's `config`/rules **differ** from the reference's.

- **Informational only** — differences are reported, never `drift`, and never affect the exit code. Two repos legitimately differ; the point is seeing where.
- Compare the rules-side fields (`config`, `requiredSkills`, `mcp`, `exceptions`), never the record-side (`assets` hashes, `writtenBy`/`writtenAt`) — hashes and stamps differing between repos is meaningless.
- The reference repo is **untrusted input**: its file may be missing (report plainly), malformed (report, do not crash), or huge (validate against the published schema before diffing; reject what fails).
- Never *apply* anything from the reference — this is a read-and-report feature. No fixer.

## 2. Start a new repo from a reference repo's rules

`setup` accepts a reference (e.g. `--from owner/repo`) and seeds its answers from that repo's `.repo-tooling.json` instead of the interactive defaults. The wizard still runs — seeded, not skipped — and record-side fields are never copied (a fresh repo gets a fresh `writtenBy`, empty `assets`).

## Constraints

- One shared fetch/validate helper for both paths; schema-validate before use.
- Public library: no `@rtorcato` defaults anywhere; the reference is always explicit user input.
- GitHub via `gh` first; structure the fetcher so another forge is a second function, not a rewrite.
- Tests: reference file missing, malformed, valid-with-differences, valid-identical; and seeding maps config→wizard defaults correctly.

## Files

- `src/cli/utils/lockfile.ts` (read side), a new fetch helper, `src/cli/commands/doctor.ts`, `src/cli/commands/setup.ts`, docs in `apps/docs/docs/reference/repo-tooling-json.mdx`.

**Note for pickup:** overlaps `lockfile.ts`/schema with #559's v4 restructure — do not run both concurrently; #559 lands first and this builds on the record/rules split it introduces (compare the `rules` subtree, seed from the `rules` subtree).

rtorcato:

Rules are per-repo and **there is no guideline package** — so sharing a guideline means pointing at a *reference repo*. This adds both directions of that, over one shared fetcher, on top of the `record`/`rules` split from #559.

Closes #563

## The two capabilities

**`doctor --rules-from owner/repo`** — reports where this repo's rules differ from that repo's.

```
📐 Rules vs rtorcato/repo-tooling (informational — differences are not drift and do not affect the exit code)

  config.bundler
     here: "esbuild"
     rtorcato/repo-tooling: "none"
  requiredSkills
     here: ["ai-issue"]
     rtorcato/repo-tooling: ["ai-issue-loop","ai-workflow","ai-issue","ai-loop-status"]

  2 difference(s).
```

**`setup --from owner/repo`** — seeds the wizard's defaults from that repo's recorded config. Seeded, **not skipped**: every question is still asked, so nothing is written the user did not see.

## Design decisions worth reviewing

- **A flag, not a committed field.** A committed `rulesFrom` would make one repo's rules a standing input to another's audit, which is the guideline-package shape wearing a different hat. A flag keeps the comparison something a human asks for, on the run they ask for it.
- **Not a `CheckResult`.** The comparison is printed in its own section and, under `--json`, as a sibling `rulesReference` key. Keeping it out of `results` is what makes "never drift, never affects the exit code" *structural* rather than a promise the next check has to remember — it cannot reach `summarize()` at all.
- **What is compared:** `record.config` plus everything under `rules`. `assets` hashes and the `writtenBy`/`writtenAt` stamps are excluded — they differ between any two repos by construction, so including them would bury every difference that means something. There is a test that asserts exactly this (identical rules, deliberately different record halves → zero differences).
- **`setup --from` copies `record.config` only.** Not `projectName` (a new repo is not the reference repo), not the record-side stamps, and **not `rules`** — an `exceptions` entry excuses a deviation in the repo that argued for it, so copying one would mute a check somewhere it is still a real finding. If you want `aiLoop`/`requiredSkills`/`mcp` to travel too, say so and I will follow up; I left it out deliberately rather than by omission.

## Untrusted input, end to end

The reference repo is never trusted:

- `owner/repo` is shape-checked against a regex **before** it reaches the gh API path (it is interpolated there). `evil/../../repos` is rejected without spawning `gh`.
- The response is size-capped at 128 KB before it is parsed.
- The JSON is migrated (so an older flat reference is judged on its rules, not its layout) and then **validated against the published schema** before a single field is read.
- Missing, malformed, oversized, schema-invalid, or gh-unavailable are each reported plainly and the comparison is skipped — never as a finding about *your* repo, never a crash.
- Nothing from the reference is ever written. No fixer.

## Shape of the code

- `src/cli/utils/reference-rules.ts` — the one fetcher both paths share. `fetchFromGitHub` is the only forge-specific function; the size cap, parse, schema validation and diff downstream of it are forge-agnostic, so another forge is a second function dispatched from `fetchReferenceLockfile`, not a rewrite.
- `src/cli/utils/json-schema.ts` — the hand-rolled schema validator **moved out of `tests/cli/utils/lockfile-schema.test.ts`**, unchanged. It had to live in `src` for the untrusted-input check; the test now imports it, so the self-validation suite and the reference check are provably the same validator. Still no ajv.
- `parseLockfile()` split out of `readLockfile()` so a lockfile fetched over `gh` goes through the exact same v1–v4 migration as one read off disk.

## Tests

`tests/cli/utils/reference-rules.test.ts` (13 cases) covers the four the issue names — missing, malformed, valid-with-differences, valid-identical — plus: reference-string rejection before any API call, non-lockfile JSON, schema-invalid, oversized, gh failure distinguished from 404, an older flat reference migrated before validation, and a repo with no lockfile of its own. `tests/cli/commands/setup.test.ts` adds the config→wizard-defaults mapping (including the two derived answers, `releaseTool` and `orchestrator`) and asserts the record side does not cross over.

- `pnpm run verify`: **1142 passed, 1 expected fail** (pre-existing), lint/typecheck/build clean.
- Smoke-tested the built CLI against a scratch repo with a real `gh`: identical, with-differences, 404 reference, and malformed reference string all render correctly, and the exit code stays driven by the checks alone.
- No new doctor check, so the pinned check lists in `tests/languages/{python,swift,perl}/fixers.test.ts` are untouched.
- **Not verified:** the interactive `setup --from` terminal session. The defaults are asserted by inspecting the question list handed to inquirer (the #463 pattern), and `--from`'s CLI wiring was confirmed against the built binary, but I did not drive a live TTY wizard.

Per the repo's install guidance I did not run `pnpm install`; no dependencies changed.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -19,6 +19,8 @@ Every command supports `--json` and a non-interactive mode. Combine with `--yes`
 | `setup --config-schema` | ✅ | ✅ (JSON Schema) | Print the JSON Schema for `ProjectConfig`. Use to validate configs before scaffolding. |
 | `setup --dry-run` | ✅ | ✅ | Print resolved config + file list without writing. Pair with `--preset` or `--config`. |
 | `doctor --json` | ✅ | ✅ | Audit a project. Returns `{ directory, results: [{ check, status, detail, hint? }] }`. Status: `ok` / `drift` / `missing` / `optional-missing` / `declared`. |
+| `doctor --rules-from <owner/repo>` | ✅ | ✅ | Report how this repo's `config`/`rules` differ from a reference repo's `.repo-tooling.json` (read over `gh`). Informational only — adds a `rulesReference` key to the JSON, never a check result, never affects the exit code. |
+| `setup --from <owner/repo>` | ❌ (seeds the wizard) | `--dry-run` only | Seed the wizard's defaults from a reference repo's recorded config. Seeded, not skipped — every question is still asked. |
 | `fix --json --yes` | ✅ | ✅ | Walk every doctor finding, apply fixers. Returns `FixActionRecord[]` with `status: applied | dry-run | skipped | already-ok | unsupported`. |
 | `fix <target> --json --yes` | ✅ | ✅ | Apply one fixer. Targets from `list --json`. |
 | `fix --dry-run` | ✅ | ✅ | Print what each fixer would write without writing. Combine with `--json`. |
--- a/apps/docs/docs/reference/repo-tooling-json.mdx
+++ b/apps/docs/docs/reference/repo-tooling-json.mdx
@@ -181,6 +181,60 @@ Three rules keep it from becoming a mute button:
 A bulk `fix` / `fix --yes` skips declared checks; a targeted `fix <target>` still applies,
 since naming the fixer is an explicit override.
 
+## Comparing against another repo
+
+There is **no guideline package**, and there never will be: the rules are unique to each repo,
+and a rule that travels as a dependency stops being the repo's own. Sharing a guideline means
+pointing at a **reference repo** — one whose `.repo-tooling.json` you consider exemplary.
+
+```bash
+# Report how this repo's config and rules differ from another repo's
+npx @rtorcato/repo-tooling doctor --rules-from owner/repo
+```
+
+The reference is read over `gh` (`repos/owner/repo/contents/.repo-tooling.json`), so it works
+for any repo your `gh` login can see, public or private.
+
+Three properties make this safe to point at a repo you do not control:
+
+- **Informational, always.** Differences are printed in their own section, never as check
+  results — they are not `drift`, they do not appear in the summary counts, and they cannot
+  change the exit code. Two repos legitimately differ; the point is seeing where.
+- **Read and report, never apply.** There is no fixer, and nothing from the reference is ever
+  written into your repo.
+- **The reference is untrusted input.** The `owner/repo` string is shape-checked before it
+  reaches an API path, the response is size-capped, and the JSON is validated against the
+  published schema before a single field is read. A missing, malformed, oversized or
+  schema-invalid reference is reported plainly and the comparison is skipped — never as a
+  finding about your repo.
+
+Only the comparable half is diffed: `record.config` and everything under `rules`. The `assets`
+hashes and the `writtenBy`/`writtenAt` stamps are excluded — they differ between any two repos
+by construction, so including them would bury every difference that means something.
+
+With `--json`, the comparison rides alongside the results under `rulesReference`, and only when
+the flag is given.
+
+## Starting a new repo from another one
+
+The same reference, the other direction:
+
+```bash
+npx @rtorcato/repo-tooling setup --from owner/repo -d ./my-new-lib
+```
+
+The wizard is **seeded, not skipped**: every question is still asked, with that repo's answers
+as the defaults instead of the built-in ones. Nothing is written that you did not see.
+
+Only `record.config` crosses over. `projectName` is never seeded (a new repo is not the
+reference repo), the record-side stamps are not copied (the new repo writes its own `writtenBy`
+and starts with no `assets`), and `rules` stays behind — an `exceptions` entry excuses a
+deviation in the repo that argued for it, so copying one would mute a check somewhere it is
+still a real finding.
+
+`--from` seeds the wizard, so it is ignored (with a warning) alongside `--preset` or `--config`,
+which skip the wizard entirely.
+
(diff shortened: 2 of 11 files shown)


rtorcato (commented):

Reviewed the reference-repo fetch path (`reference-rules.ts`) for injection, traversal, and untrusted-input handling: the `owner/repo` string is regex-gated (single slash, alnum-anchored segments, no shell metacharacters) before it's used as a single array element passed to `gh` — not shell-interpolated, so no command/flag injection. Response is size-capped before parse, migrated then schema-validated before any field is read, and nothing from the reference is ever written (no fixer). Design matches the issue's untrusted-input requirements, and the new hand-rolled schema validator avoids adding a dependency. No `@rtorcato`-specific values baked into emitted config or checks — seeding only copies `record.config`, gated behind schema validation of the source.

### Before merging
Nothing.

rtorcato (commented):

Architecture matches #563: no guideline package, one shared fetcher, reference read via `gh`, never written back. Input validation (owner/repo regex checked before the gh call, 128KB cap, schema validation before use) is sound.

Gap: the PR body claims "missing, malformed, oversized, schema-invalid, or gh-unavailable are each reported plainly ... never a crash" end to end, but that only holds for `doctor --rules-from`. `setup --from`'s `seedFromReference` throws a raw `Error` on any fetch failure instead of reporting and declining the way `resolveConfig`'s other failure paths do, and there's no test for it — only the success case and the `--preset`/`--config` ignore case are covered.

### Before merging
- `setup --from`'s failure path (missing/malformed/unreachable reference) throws uncaught and is untested — should report plainly and decline like `doctor --rules-from` does, per the PR's own "never a crash" claim.

rtorcato (commented):

Re-reviewed after `6ec0480`. The prior finding is resolved: `seedFromReference` now reports plainly (`console.error` + decline) instead of throwing, matching the `doctor --rules-from` "not compared" pattern, and `tests/cli/commands/setup.test.ts`'s "declines plainly when the reference cannot be read" test covers it end to end — asserts no `Setup failed` dump, no `process.exit`, wizard never invoked, and no file written.

Checked the rest of the diff for what the fix itself might have introduced: none. `resolveConfig`'s `--from`/`--config`/`--preset` interaction, the `REFERENCE` regex rejecting multi-segment paths, and the `json-schema.ts` extraction are all unchanged from the prior pass and correct.

### Before merging
Nothing.

rtorcato (commented):

Re-reviewed from scratch after `6ec0480`. Confirmed: `gh` is invoked via `spawn(args[])` with no shell, so the regex-validated `owner/repo` can't reach a shell even before the regex; the regex itself rejects anything not starting alnum (blocks `../`, extra slashes). Size cap (128 KB) and schema validation both run before any field of the remote JSON is read. No new dependencies.

The commit under review swaps a raw `throw new Error(...)` for a caught decline (`console.error` + return `null`). This is a net risk reduction, not a new leak: the old path would have hit a generic `Setup failed:` handler and likely dumped `error.stack` (local absolute paths); the new path prints only the pre-validated `reference` string and a `reason` built from constants / a truncated `gh` stderr line. Test added for the decline path is adequate.

### Before merging
Nothing.

Issue #567, opened 2026-08-27
ai-issue-loop: verdict-marker read is not scoped to the head commit, so a fix round adopts stale verdicts


rtorcato:
🤖 *Filed by an agent via ai-issue-loop.*

Pass 3 reads a reviewer's verdict back off its posted review, taking `last`:

```
capture("<!-- ai-issue-loop:verdict:<ARM>:(?<v>[A-Z-]+) -->").v] | last // empty
```

Nothing scopes that to a commit. **After a fix round the newest marker is still the pre-fix one**, so the tick adopts a verdict about a diff that no longer exists.

Followed literally on #565 today it would have re-applied `ai-changes` for the finding the fix round had just resolved:

| | |
|---|---|
| reviews submitted | `17:34:19Z`, `17:34:56Z` |
| fix pushed (`6ec0480`) | `17:38:54Z` |
| markers still readable | `code=CHANGES`, `sec=PASS` |

That is round 2 of 2 spent on nothing, leaving #563 one `ai-changes` from `ai-blocked` — handed to a human half-finished, with a PR whose objection was already fixed. The `PASS` half is the mirror failure and the more dangerous one: a stale pass adopted onto a rewritten diff marks it reviewed when nothing read it.

## What to change

Scope the read to the current head. `pulls/<N>/reviews` entries carry `commit_id`, so filter on it:

```
select(.commit_id == $head)
```

with `$head` from `gh pr view <N> --json headRefOid`. A marker from an earlier commit then reads as absent, which re-spawns that arm — the correct outcome, and already the behaviour for an arm that never posted.

## Why the recovery case still works

The read exists to recover a reviewer that died between posting and labelling (#497). That reviewer posted against the *current* head, so a head-scoped filter still finds it. Only genuinely stale markers stop matching, which is the whole point.

Related: the fix-round spawn has no claim label, filed separately.

PR #568, opened 2026-08-27
feat(ai-issue-loop): claim the fix round with an ai-fixing label

### LINKED ISSUE #566: ai-issue-loop: fix-round implementer spawns with no claim label, so a mid-run tick duplicates it

Pass 3 spawns a fix-round implementer off `ai-changes` **without applying a claim label first**, so a stateless tick landing mid-run spawns a second implementer into the same worktree.

Reviewers already have this guard — Pass 3 applies `ai-reviewing-code` / `ai-reviewing-sec` immediately *before* spawning, and the skill spells out why: a reviewer only labels at the *end*, so from spawn until then the labels are indistinguishable from "nobody has started". **The fix round has exactly that gap and no guard.** From the implementer's spawn until it pushes, the PR reads `ai-changes` with no claim, which is precisely the selector Pass 3 dispatches on.

A fix round takes longer than a 15-minute tick. Observed on #565 today: `ai-changes` applied 17:35, implementer pushed 17:38 — a tick at either boundary would have matched. The failure is worse than a duplicated reviewer, because two implementers share one worktree and one branch: they race each other's commits and `git -C` operations, rather than merely posting two comments.

## What to change

- Add an `ai-fixing` label to the canonical table in `src/base/labels.ts` and to the bootstrap block in `skills/ai-issue-loop/SKILL.md` (a test asserts the two match).
- Pass 3: apply it immediately before the fix-round spawn, the same shape as the reviewer claims.
- The implementer's existing relabel command clears it, alongside the removals it already does.
- Pass 2 stall reaping: add the matching row, so a dead implementer's claim is dropped rather than wedging the PR forever.

## Why this only bites unattended

Driven by hand, the operator knows an implementer is live and skips the re-spawn — which is exactly the conversation-held state the skill's own opening rule forbids relying on ("All state lives in GitHub labels ... Never keep pipeline state in the conversation"). Under `/loop 15m` there is no operator, so nothing stops it.

Related: the verdict-marker read has an adjacent scoping defect, filed separately.

rtorcato:

Pass 3 dispatched the fix-round implementer off `ai-changes` with no claim label, so from spawn until the implementer pushed, the PR read exactly the selector Pass 3 dispatches on. A stateless tick landing in that window spawned a second implementer into the same worktree and branch — the two race each other's commits and `git -C` operations, which is worse than the duplicated reviewers the `ai-reviewing-*` claims already prevent.

## What changed

- **`src/base/labels.ts`** — `ai-fixing` added to the canonical table, colour `006b75`, deliberately distinct from every existing label (#446 is the precedent for what indistinguishable colours cost). The two "thirteen labels" comments now say fourteen.
- **`skills/ai-issue-loop/SKILL.md`** — the four places that must agree: the bootstrap `gh label create` line and Labels-table row; the Pass 3 claim applied immediately before the spawn, mirroring the reviewer claims including the `${AGENT_USER:+--add-assignee "$AGENT_USER"}` suffix; `--remove-label ai-fixing` in the implementer's relabel-on-push; and a Pass 2 stall-reaping row so a dead implementer's claim is dropped rather than wedging the PR. The Pass 3 selector now skips a PR already carrying `ai-fixing`.
- **`apps/docs/docs/guides/ai-issue-loop.md`** — the public mirror of the Labels table and the claim-step paragraph, kept in step.
- Both ASCII state diagrams: `fix round (max 2)` becomes `ai-fixing (max 2)`, an equal-length swap that preserves the alignment. The reviewer arm already named its claim label there; the fix arm now does too.

## Verification

`biome check`, `tsc --noEmit`, and the full suite pass (1143 tests) — including `tests/base/labels.test.ts`, which asserts the skill's bootstrap block matches `LOOP_LABELS` exactly.

`ai-fixing` was created on this repo before pushing, so `scripts/dogfood.mjs` stays green — a new entry in the canonical table is a bootstrap gap on this repo until the label exists (the #543 failure mode). `node scripts/dogfood.mjs` now reports clean over 63 checks.

Out of scope: #567 covers the verdict-marker read not being scoped to the head commit, in the same file. Left alone deliberately to avoid a conflict.

Closes #566

### DIFF
--- a/apps/docs/docs/guides/ai-issue-loop.md
+++ b/apps/docs/docs/guides/ai-issue-loop.md
@@ -132,6 +132,7 @@ that state, so a missed tick, a crash, or a restart costs nothing.
 | `ai-ok-code` | PR | `code-reviewer` passed. In-flight only — the handoff strips it. |
 | `ai-ok-sec` | PR | `security-expert` passed. In-flight only — the handoff strips it. |
 | `ai-changes` | PR | A reviewer requested changes, or Pass 1 sent the PR back over red CI. |
+| `ai-fixing` | PR | Fix-round implementer claimed and running. Cleared with its push. |
 | `ai-notes` | PR | Passed, but a reviewer left something to read before merging. |
 | `merge-ready` | PR | Both agent reviews passed and the PR is mergeable — waiting on a human. Supersedes the `ai-ok-*` pair rather than joining it. |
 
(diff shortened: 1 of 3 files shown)


rtorcato (commented):

Diff is labels-table and skill-doc only; no runtime code path handles user/issue input. New `gh pr edit` snippets use fixed label-name literals and the existing `${AGENT_USER:+--add-assignee "$AGENT_USER"}` guarded-expansion form, not bare interpolation of anything issue/PR-derived. No hardcoded `@rtorcato`-specific values leak into the shipped skill or label table. No secrets, no dependency changes.

### Before merging
Nothing.

rtorcato (commented):

Checked the four places for character-for-character agreement (`src/base/labels.ts` canonical entry, `SKILL.md` bootstrap line, Pass 3 claim command, implementer relabel, Pass 2 stall-reap row) — all consistent. Claim is applied immediately before the fix-round spawn with `${AGENT_USER:+--add-assignee "$AGENT_USER"}`, mirroring the reviewer claims. The implementer's relabel command clears `ai-fixing`. Colour `006b75` doesn't collide with any colour visible in this diff's context. Both ASCII diagrams do an equal-length swap (`fix round`→`ai-fixing`, 9 chars) so alignment holds. No drift into #567's territory.

### Before merging
Nothing.

PR #569, opened 2026-08-27
fix(skills): scope the verdict-marker read to the PR's head commit

### LINKED ISSUE #567: ai-issue-loop: verdict-marker read is not scoped to the head commit, so a fix round adopts stale verdicts

rtorcato:

Pass 3's verdict-marker read was unscoped to a commit, so `last` returned the newest marker on the PR regardless of which diff it was written about. After a fix round that is still the *pre-fix* marker, and the tick adopts it:

- a stale `CHANGES` re-applies `ai-changes` for a finding the fix round already resolved — a round of two spent on nothing, pushing the PR toward `ai-blocked`;
- a stale `PASS` is worse: it marks a rewritten diff reviewed when nothing read it.

## Change

`skills/ai-issue-loop/SKILL.md` only. The read now takes the PR's head commit and filters on it:

```bash
HEAD=$(gh pr view <N> --json headRefOid --jq .headRefOid)
… | select(.user.login==$me and .commit_id==$head) | …
```

An older marker reads as absent, which re-spawns that arm — already the behaviour for an arm that never posted. `--paginate --slurp` + `add`, `(.body // "")`, `// empty` and the `.user.login` author gate are all unchanged; a fifth bullet documents the head gate alongside them, including why it does not cost the #497 recovery case (a reviewer that died between posting and labelling posted against the head that is still current, so its marker still matches).

## Verification

`commit_id` confirmed present on `pulls/<N>/reviews`, and the exact filter run against #568: current head yields `code=PASS`, `sec=PASS`; a non-matching head yields empty for both, i.e. the re-spawn path.

`apps/docs/docs/guides/ai-issue-loop.md` does not carry this snippet, so it needed no change. `src/base/labels.ts` and the SKILL.md bootstrap label block are untouched; `tests/base/labels.test.ts`, `tests/cli/generators/claude-skills.test.ts` and `tests/cli/commands/loop-guard.test.ts` pass (80 tests).

Closes #567

### DIFF
--- a/skills/ai-issue-loop/SKILL.md
+++ b/skills/ai-issue-loop/SKILL.md
@@ -964,14 +964,15 @@ review that already exists — `<ARM>` is `code` or `sec`:
 
 ```bash
 ME=$(gh api user --jq .login)   # the identity every loop agent posts as
+HEAD=$(gh pr view <N> --json headRefOid --jq .headRefOid)
 VERDICT=$(gh api "repos/$OWNER_REPO/pulls/<N>/reviews" --paginate --slurp \
-  | jq -r --arg me "$ME" '[add[]
-            | select(.user.login==$me)
+  | jq -r --arg me "$ME" --arg head "$HEAD" '[add[]
+            | select(.user.login==$me and .commit_id==$head)
             | (.body // "")
             | capture("<!-- ai-issue-loop:verdict:<ARM>:(?<v>[A-Z-]+) -->").v] | last // empty')
 ```
 
-Four details there are load-bearing:
+Five details there are load-bearing:
 
 - **`pulls/<N>/reviews`, because the prompt posts with `gh pr review --comment`.**
   That creates a *review*, which never appears under `issues/<N>/comments`. The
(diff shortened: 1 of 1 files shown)


rtorcato (commented):

`$HEAD` is captured via command substitution but passed into the filter through `--arg head "$HEAD"`, so it's jq data, not interpolated filter syntax — same pattern as `$ME`. The author gate (`.user.login==$me`) is untouched and combined with `and`, so head-scoping is additive, not a replacement — a stranger's review still can't forge a marker. No shell-interpolated API output, no secrets, no dependency changes. Doc-only edit to `SKILL.md`; no `@rtorcato`-specific values introduced.

### Before merging
Nothing.

rtorcato (commented):

Filter is correct: `.commit_id==$head` passed via `--arg head "$HEAD"` (data, not interpolated), and `commit_id` is a real field on `pulls/<N>/reviews` entries. All four prior guards (`(.body // "")`, `// empty`, `.user.login==$me`, `--slurp`+`add`) are preserved unchanged; the head gate is ANDed into the existing `select`, not a replacement. `last` still makes sense post-scoping (newest marker among same-head reviews wins). The #497 recovery-case prose is accurate — a reviewer that died between posting and labelling posted against the still-current head, so it still matches.

Docs guide: correctly left alone. `apps/docs/docs/guides/ai-issue-loop.md` documents labels/state-machine, not jq mechanics — confirmed it has no `capture(`/`jq -r`/`commit_id` content, unlike #568's `ai-fixing` label which *is* guide-level. `src/base/labels.ts` untouched, confirmed.

### Before merging
Nothing.

Issue #575, opened 2026-08-27
doctor: flag sources under Sources/ that belong to no declared target


rtorcato:
🤖 *Filed by an agent via ai-issue-loop.*

Split out of #574, which fixed the write side only: `setup --preset swift-library` no longer rewrites an existing `Package.swift`. This is the detection side the issue floated, deliberately left out of that PR because it is a new doctor check rather than a preserve rule.

**Why it is worth having independently.** A directory under `Sources/` that no target declares is invisible to the toolchain: SwiftPM ignores it, so `swift build` and `swift test` both exit 0 while the library those files define is not being built at all. There is no red X anywhere. Any hand-edit of the manifest — a renamed target, a deleted `.target(...)` line, a merge that drops one — leaves the same silent hole, so the check earns its keep beyond the scaffold bug that surfaced it.

**Sketch.** `parsePackageSwift` in `src/languages/swift/ci.ts` already regex-parses the manifest for products and platforms; it would need target names too. Compare those against the directory names under `Sources/` and report the difference.

Worth settling first:
- Custom `path:` on a target moves its sources somewhere other than `Sources/<Name>/`, so a naive directory comparison false-positives. A manifest using `path:` may be better skipped than guessed at.
- `Tests/` has the same shape via `.testTarget`.
- Status: `drift` (a real defect, silently) versus advisory. No safe autofix exists — the fixer cannot know whether the right repair is to add a target or delete the directory — so this is likely a check with a hint and no `fix` target.

Note that Swift support is still being built out; this may be better folded into that work than done standalone.

PR #576, opened 2026-08-27
fix(swift): preserve an existing Package.swift instead of overwriting it

### LINKED ISSUE #574: setup --preset swift-library overwrites an existing Package.swift and orphans its sources

**Context: Swift support is still being built out, so this is a note for that work rather than a regression report.** Recording it because the failure is silent and easy to miss once the preset sees real repos.

`setup --preset swift-library` overwrites an existing `Package.swift` and orphans the package's real sources. Found dogfooding on a greenfield Swift package.

Given a committed manifest declaring target `Widget`:

```
-    name: "Widget",
-    products: [.library(name: "Widget", targets: ["Widget"])],
-    targets: [.target(name: "Widget")]
+    name: "Swiftlib",
+    platforms: [.iOS(.v17), .macOS(.v14)],
+    products: [.library(name: "Swiftlib", targets: ["Swiftlib"])],
+    targets: [.target(name: "Swiftlib"), .testTarget(name: "SwiftlibTests", ...)]
```

Three losses in one write: the package is **renamed after the directory** (`swiftlib` → `Swiftlib`), tools-version is bumped 5.9 → 6.0, and platform floors are invented. `Sources/Widget/Widget.swift` stays on disk declared by no target.

**The part worth designing around: `swift build` still succeeds.** SwiftPM ignores the undeclared directory, so the package compiles green while no longer building the user's library, and `doctor` reports it clean with exit 0. A build error would be noticed; this is not.

## Worth settling while the preset is still in flight

`Package.swift` is source, not config — the JS presets already treat `package.json` that way, merging and preserving `name`/`version`/`private`/scripts/deps.

- If `Package.swift` exists, do not rewrite it; at most add what is missing.
- Never derive the package name from the directory when the manifest already states one.
- Never scaffold `Sources/<Name>/` or `Tests/<Name>Tests/` for a target the manifest does not declare.
- Candidate doctor check: sources under `Sources/` belonging to no declared target.

rtorcato:

**Swift support is still being built out, so this is deliberately scoped to the preserve rule and nothing else.** No restructuring of the preset, no new doctor check — just the one guard that stops the data loss, so it should not collide with work in progress.

Closes #574

## The defect

`setup --preset swift-library` rewrote any `Package.swift` already on disk. Three losses in one write: the package was renamed after the *directory*, tools-version was bumped, and platform floors were invented — leaving the real `Sources/<Target>/` declared by no target.

The part that makes it worth a guard rather than a note: **SwiftPM ignores an undeclared directory, so `swift build` still exits 0** and `doctor` reports clean. A build error would be noticed; this is not.

## The fix

`Package.swift` is source, not config. `generateSwiftProject` now reads the manifest before writing anything, and when one exists:

- it is left **byte-identical** — not merged, not extended;
- the module name is read back off the manifest (`.library(name:)`) rather than derived from the directory;
- nothing is scaffolded under `Sources/` or `Tests/`, since those would be directories for a target the manifest does not declare.

Everything else in the standard still lands — SwiftLint config, Periphery, CI, hooks, security workflows — so running the preset over an existing package remains useful. This mirrors how the JS presets already treat `package.json`. Reuses the existing `parsePackageSwift`; no new parser.

## Tests

Four new cases in `tests/languages/swift/scaffold.test.ts`, against a directory holding a tools-version 5.9 manifest declaring target `Widget` plus `Sources/Widget/Widget.swift`. The first three fail on `main`:

- manifest is byte-identical after the run
- no `Sources/Swiftlib` or `Tests/SwiftlibTests` appears, and `Sources/Widget/Widget.swift` survives
- the README names `Widget`, not the directory
- `.swiftlint.yml` and CI are still written

Full suite green (1147 passed), biome and tsc clean.

## Not done here

The doctor check for sources under `Sources/` belonging to no declared target is a larger, separate piece of work — filed rather than built.

Follow-up: #575

### DIFF
--- a/src/languages/swift/scaffold.ts
+++ b/src/languages/swift/scaffold.ts
@@ -19,7 +19,12 @@ import { generateEditorConfig } from '../../cli/generators/misc.js'
 import { generateCodeQLWorkflow, generateDependabotConfig } from '../../cli/generators/security.js'
 import { copyPreset } from '../../cli/utils/copy-preset.js'
 import { LANGUAGES } from '../registry.js'
-import { readSwiftPackage, renderSwiftReleaseWorkflow, renderSwiftWorkflow } from './ci.js'
+import {
+	parsePackageSwift,
+	readSwiftPackage,
+	renderSwiftReleaseWorkflow,
+	renderSwiftWorkflow,
+} from './ci.js'
 import { SWIFT_HOOKS_DIR, installSwiftGitHooks } from './git-hooks.js'
 import { ensureSwiftGitignore } from './gitignore.js'
 
@@ -205,6 +210,10 @@ git tag 1.0.0 && git push origin 1.0.0
 /**
  * Relative paths `setup --dry-run` reports for a Swift config, in write order.
  * Kept beside the writer so the two can't drift.
+ *
+ * This is the greenfield list. Against a directory that already has a
+ * Package.swift the writer preserves it and skips the target scaffolding, so the
+ * preview over-reports there — the safe direction for a dry run.
  */
 export function swiftFileList(config: ProjectConfig): string[] {
 	const module = swiftModuleName(config.projectName)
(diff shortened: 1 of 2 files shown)


rtorcato (commented):

Checked the module-name fallback (`existingManifest?.products[0] ?? swiftModuleName(...)`) — when an existing manifest is preserved, that name only lands in `readme()` markdown text, not in any `fs.outputFile`/`path.join` call, so a crafted `.library(name:)` value can't cause path traversal. The Sources/Tests scaffold paths still use `swiftModuleName` only in the greenfield branch.

Also checked the read-before-write path for the silent-overwrite failure mode raised in the task: `fs.readFile` after `pathExists` throws on a permission error rather than being caught, so a read failure aborts the write instead of falling back to clobbering the manifest — no silent data loss reintroduced.

### Before merging
Nothing.

rtorcato (commented):

Traced the only write path: `existingManifest` gates the sole `fs.writeFile(manifestPath, ...)` plus the Sources/Tests scaffolding in one `if (!existingManifest)` block, so preservation holds on every branch, not just the common one. Greenfield path is untouched (same `packageSwift`/`sourceFile`/`testFile` calls). Tests assert the actual invariant (`toBe(EXISTING)` byte-identical), not just existence or name. CI workflow correctly re-reads the preserved manifest via `readSwiftPackage`. Scope stays inside scaffold.ts/its test — no #575 doctor check, no preset restructuring.

### Before merging
- `module = existingManifest?.products[0] ?? ...` picks the first `.library` product; for a manifest with multiple products, or a product name that differs from its target folder, the README's `Sources/${module}/` line can point at the wrong path. Package.swift itself is unaffected (still byte-identical) — cosmetic only.

rtorcato:

Ready to merge — both agent reviews passed. One thing to read first:
`code-reviewer` left a cosmetic note about the README module path on a
multi-product manifest. See its [### Before merging](https://github.com/rtorcato/repo-tooling/pull/576#pullrequestreview-5044827363).
`Package.swift` itself is byte-identical either way, which is the invariant this PR exists for.

Issue #573, opened 2026-08-27
setup --preset library leaves doctor at exit 1 (missing minimumReleaseAgeExclude for its own scope)


rtorcato:
🤖 *Filed by an agent via ai-issue-loop.*

`setup --preset library` leaves `doctor` **non-zero on a repo it just created**. Found dogfooding on a greenfield library named `@acme/widget`.

```
$ setup --preset library --yes    # exit 0
$ doctor                          # exit 1
[drift] pnpm settings
   detail: missing minimumReleaseAgeExclude: @acme/*
   hint: Run `npx @rtorcato/repo-tooling fix pnpm-workspace` to merge them in
```

`fix pnpm-workspace` resolves it and doctor then exits 0, so the fixer is right — setup just didn't write it.

The value is derivable at setup time: the scope comes straight from the `name` in `package.json`, which setup has already read (it preserves it). Only scoped packages trip this, which is why the `nextjs-app` preset (unscoped `my-next-app`) came out clean.

Low severity — the generated CI does not run `doctor`, so nothing goes red. But "run setup, you're aligned" should hold, and a non-zero exit is what a user's own CI or pre-commit would key on.

## What to change

Have the `library` preset derive `pnpm.minimumReleaseAgeExclude` from the package's own scope when the name is scoped, so `setup` and `doctor` agree immediately.

Reproduced with `3.11.0` (`dist/` built from `05750f8`).

PR #577, opened 2026-08-27
fix(setup): make the library preset own the build script it writes exports for

### LINKED ISSUE #570: setup --preset library: exports point at .cjs/.d.cts that the preserved `build: tsc` never emits

`setup --preset library` writes a **dual-format publish contract that the build script it preserves cannot produce**. Found dogfooding on a greenfield library.

Starting from `{"main": "dist/index.js", "scripts": {"build": "tsc"}}`, setup wrote:

```json
"main": "./dist/index.cjs",
"module": "./dist/index.js",
"exports": { ".": { "types": { "require": "./dist/index.d.cts" }, "require": "./dist/index.cjs" } },
"scripts": { "build": "tsc", "build:watch": "tsup --watch" }
```

plus a `tsup.config.ts`. But `build` is still `tsc`, and `tsc` with `outDir: ./dist` emits **only** `dist/index.js` and `dist/index.d.ts` — never `.cjs` or `.d.cts`.

So `pnpm build && npm publish` ships a package whose `main` and every `require` condition point at files that do not exist. Every CJS consumer breaks. `doctor` reports the repo clean.

**The cause is the merge policy, not a typo.** Preserving existing scripts is what makes the JS presets safer than `swift-library` (see the sibling issue) — but here setup changed the *contract* (`exports`) without changing the *producer* (`build`). It even added `build:watch: tsup --watch`, so the intent to use tsup is clear; only the real `build` was left behind.

## What to change

Pick one and make it coherent:

- Repoint `build` to `tsup` when the preset writes a tsup config and dual-format `exports` — and say so, since overwriting an existing `build` script is exactly the clobbering the merge policy avoids elsewhere.
- Or leave `build` alone and emit single-format `exports` matching what `tsc` produces.

Either way a doctor check would catch the general case: `exports`/`main` naming artefacts no build script can emit. `publint` (already in the generated `verify`) catches it at release time; nothing catches it at setup time.

rtorcato:

Closes #570

## The defect

`setup --preset library` writes the publish contract unconditionally — `main`, `module`, `types`, `exports`, `files`, `publishConfig` are all assigned after the merge, clobbering whatever was there. But `scripts` merged with **existing values winning**, so a repo's pre-existing `"build": "tsc"` survived.

`tsc` with `outDir: ./dist` emits `dist/index.js` and `dist/index.d.ts` and nothing else. The contract named `./dist/index.cjs` as `main` and `./dist/index.d.cts` as the `require` types condition. `pnpm build && npm publish` therefore shipped a package whose every CJS entry point 404s.

## Which resolution, and why

The issue sketched two: repoint `build` at tsup, or emit single-format `exports` matching `tsc`. **This takes the first**, scoped to `projectType === 'library'` when the preset generates a `build` at all.

The merge-don't-clobber policy is real and worth keeping — it is what makes re-running `setup` on a live repo safe, and what separates the JS presets from `swift-library`. But the policy was already not being applied to the publish contract: the library branch overwrites `main`/`module`/`types`/`exports`/`files`/`publishConfig` without asking. Preserving `build` alone, while overwriting every field that describes what `build` must produce, is the inconsistency. If the preset owns the contract, it has to own the producer.

The alternative — deriving single-format `exports` from the preserved `build` — is a bigger and leakier change. `attw` would need the `esm-only` profile (there is already an `isEsmOnly` branch in the `attw` fixer), the shipped `tsup.config.ts` would still say `format: ['cjs', 'esm']`, and `build:watch: tsup --watch` would still point at a bundler the `build` script does not use. It trades one incoherence for three smaller ones.

Scope is deliberately one script. Everything else — `test`, `lint`, custom targets — still merges under existing-wins, and the replacement is announced on stdout rather than done silently:

> ⚠️  Replaced "build": "tsc" with "tsup" — the library exports map names files only tsup emits.

## The regression test

The reason this shipped is that nothing asserted setup's own output was internally coherent. `tests/cli/generators/package-json.test.ts` now runs the real `library` preset (via `buildPresetConfig`) and asserts the invariant: **every path named in `main`/`module`/`types`/`exports` must be producible by the resolved `build` script**, checked both greenfield and against a package.json that arrived with `build: tsc`.

Verified it fails on the old behaviour before the fix:

```
AssertionError: expected [ './dist/index.d.cts', …(1) ] to deeply equal []
+ [ "./dist/index.d.cts", "./dist/index.cjs" ]
```

Two further assertions guard the blast radius: a sibling `my:script` and `test` still survive the merge, and a non-library project keeps its own `build` untouched.

## Checks

`pnpm typecheck`, `biome check` on the changed files, and the full suite (1146 passed) are green. Snapshot output for the greenfield library is unchanged — the fix only bites when an existing `build` differs.

Out of scope, as queued: #573 (`pnpm.minimumReleaseAgeExclude` on the same preset).

### DIFF
--- a/src/cli/generators/package-json.ts
+++ b/src/cli/generators/package-json.ts
@@ -1,3 +1,4 @@
+import chalk from 'chalk'
 import fs from 'fs-extra'
 import path from 'node:path'
 import type { ProjectConfig } from '../commands/setup.js'
@@ -13,6 +14,9 @@ export async function generatePackageJson(config: ProjectConfig, targetDir: stri
 
 	const includeTreeshake = Boolean(config.treeshakeCheck && config.projectType === 'library')
 
+	const generatedScripts = getScripts(config, { includeTreeshake })
+	const existingScripts: Record<string, string> = (existingPackageJson as any)?.scripts ?? {}
+
 	const packageJson: any = {
 		name: config.projectName,
 		version: '0.1.0',
@@ -22,10 +26,7 @@ export async function generatePackageJson(config: ProjectConfig, targetDir: stri
 		// generated workflows carry no `version:` input (#364).
 		packageManager: `pnpm@${detectPnpmVersion() ?? PNPM_FALLBACK_VERSION}`,
 		...existingPackageJson,
-		scripts: {
-			...getScripts(config, { includeTreeshake }),
-			...(existingPackageJson as any)?.scripts,
-		},
+		scripts: resolveScripts(config, generatedScripts, existingScripts),
 		dependencies: {
 			...(existingPackageJson as any)?.dependencies,
 		},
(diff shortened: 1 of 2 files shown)


rtorcato (commented):

Reviewed for injection, secrets, unsafe shell/SQL construction, and supply-chain impact. The only interpolation (`config.bundler` into a `console.warn` string) is internally generated, not user/attacker input, and never reaches a shell or file write beyond the warning text itself. `scripts.build` is set from the preset's own generated value, not from external input, so it can't be used to inject arbitrary shell content into the consumer's `package.json`. No widening of `files`/`exports`/`main` — the change narrows the contract to what the build actually emits. No new dependency added in this diff; `chalk` import is the only new symbol and is a print-only formatting call. No secrets present.

### Before merging
Nothing.

rtorcato (commented):

Fix is scoped correctly: `resolveScripts` only touches `build` when `config.projectType === 'library'`, and the non-library merge path is unchanged (verified against `package-json.ts` and the "leaves a non-library build alone" test). The new tests assert the actual invariant — every path in `main`/`module`/`types`/`exports` must be in the emitted-file list for the resolved `build` script — not just that `build === 'tsup'`, so they'd have caught the original bug. No stray touch to #573.

Follow-up: #578

### Before merging
- The library branch overwrites *any* pre-existing `build` script unconditionally (not just `tsc`) once `projectType === 'library'` — e.g. a hand-rolled `"build": "make dist"` on a library repo gets silently repointed to tsup too, just with a stdout warning. That's a defensible reading of the issue's chosen resolution, but it's a policy call with reach beyond this bug, shipped under a `fix(...)` subject (patch semver) that also changes what `setup` does to a user's existing scripts. Worth the maintainer's eye on whether that's `fix` or `feat`/`BREAKING`.

rtorcato:

Ready to merge, but **one decision is yours**: the library branch overwrites
*any* pre-existing `build` script, not just `tsc`. `code-reviewer` flags that
as a policy call shipped under a `fix(...)` subject — i.e. a patch release that
changes what `setup` does to an existing repo. See its
[### Before merging](https://github.com/rtorcato/repo-tooling/pull/577#pullrequestreview-5044864471).

Follow-up filed as #578.

PR #580, opened 2026-08-27
fix(setup): derive the pnpm release-age exemption from package.json name

### LINKED ISSUE #573: setup --preset library leaves doctor at exit 1 (missing minimumReleaseAgeExclude for its own scope)

rtorcato:

`setup --preset library` left `doctor` at exit 1 on a repo it had just written, whenever that repo's package name is scoped.

## Root cause

`generateConfigs` derived the release-age glob from `config.projectName`:

```ts
await ensurePnpmSettings(targetDir, bundlerNeedsEsbuild(config), familyGlob(config.projectName))
```

Under `--preset`, `projectName` is only the directory basename (`setup.ts:207` — `path.basename(path.resolve(options.directory))`). But `generatePackageJson` spreads `...existingPackageJson` *after* its own `name`, so an existing scoped `name` is preserved — and `doctor` derives its glob from that field (`checks.ts:1191` — `familyGlob(pkg?.name)`).

So `@acme/widget` in a `widget/` directory gave setup `familyGlob('widget')` → `null` → no setting written, while doctor asked for `@acme/*`. The two disagreed on every scoped repo. An unscoped name made both agree on `null`, which is why the `nextjs-app` preset came out clean.

The generator itself was already correct — `familyGlob` and the fixer needed no change, which matches the report that `fix pnpm-workspace` resolves it.

## Fix

Read the name back off the emitted `package.json` — the same field doctor reads — so the two agree by construction rather than by coincidence. Nothing scope-specific is hardcoded; an unscoped package still gets no exemption.

## Tests

Two regression cases in `tests/cli/commands/setup.test.ts`, both driving the real `generateConfigs`: a pre-seeded `@acme/widget` package.json in a `widget`-named config yields `- '@acme/*'`, and an unscoped package yields no `minimumReleaseAgeExclude` at all. The first fails on `main`.

Full suite green (1145 passed), plus `biome check` and `tsc --noEmit`.

## Verification note

This worktree arrived without the `node_modules` symlink its siblings have, so per instructions I did not run `pnpm install`; I symlinked to the main checkout's `node_modules` (the established convention here) and ran vitest, biome and tsc directly. I did **not** run `pnpm test:integration` — it shells out to `pnpm build-cli` and the integration script, which was outside what I could exercise safely from a shared-modules worktree.

Closes #573

### DIFF
--- a/src/cli/generators/index.ts
+++ b/src/cli/generators/index.ts
@@ -113,7 +113,14 @@ export async function generateConfigs(config: ProjectConfig, targetDir: string) 
 
 	// Family-wide pnpm settings (#314). Runs last of the workspace writers so it
 	// merges into whatever they wrote rather than racing them for the file.
-	await ensurePnpmSettings(targetDir, bundlerNeedsEsbuild(config), familyGlob(config.projectName))
+	//
+	// The scope is read back off the emitted package.json rather than taken from
+	// config.projectName: under `--preset` that name is only the directory
+	// basename, while generatePackageJson preserves an existing scoped `name`.
+	// doctor derives the glob from that same field, so reading the basename left
+	// the two disagreeing on every scoped repo (#573).
+	const { name } = (await fs.readJson(path.join(targetDir, 'package.json'))) as { name?: unknown }
+	await ensurePnpmSettings(targetDir, bundlerNeedsEsbuild(config), familyGlob(name))
 
 	// Turborepo task pipeline (pnpm-workspace monorepos, when opted-in)
 	if (config.turborepo) {
--- a/src/cli/generators/pnpm-workspace.ts
+++ b/src/cli/generators/pnpm-workspace.ts
@@ -26,10 +26,16 @@ export const WORKSPACE_FILE = 'pnpm-workspace.yaml'
  *
  * One glob covers a whole scope: pnpm matches these entries with
  * `@pnpm/config.matcher`, so there's no package list to keep in sync.
+ *
+ * The scope is matched against npm's own name charset, not just "anything up
+ * to the slash": the name comes verbatim from a pre-existing `package.json`
+ * and is never validated as an npm name on the way here, so a scope of a lone
+ * wildcard would otherwise be taken as a glob and exempt every scoped package
+ * from the release-age delay. An unparseable scope gets no setting at all.
  */
 export function familyGlob(packageName: unknown): string | null {
 	const name = typeof packageName === 'string' ? packageName : ''
-	const scope = /^(@[^/]+)\//.exec(name)?.[1]
+	const scope = /^(@[a-z0-9-][a-z0-9._-]*)\//i.exec(name)?.[1]
 	return scope ? `${scope}/*` : null
 }
 
--- a/tests/cli/commands/setup.test.ts
+++ b/tests/cli/commands/setup.test.ts
@@ -43,6 +43,32 @@ describe('generateConfigs size-limit budget', () => {
 	})
 })
 
+// #573: under `--preset` the config's projectName is only the directory
+// basename, so a repo whose package.json is already scoped used to get no
+// minimumReleaseAgeExclude — leaving doctor at exit 1 on a tree setup just wrote.
+describe('generateConfigs release-age exemption', () => {
+	it('derives the scope from an existing package.json name, not the directory', async () => {
+		const dir = newTmpDir()
+		await fs.writeJson(join(dir, 'package.json'), { name: '@acme/widget' })
+		await generateConfigs(
+			{ ...buildPresetConfig('library', 'widget'), aiSetup: false, securityAutomation: false },
+			dir
+		)
+		const yaml = await fs.readFile(join(dir, 'pnpm-workspace.yaml'), 'utf-8')
+		expect(yaml).toContain("- '@acme/*'")
+	})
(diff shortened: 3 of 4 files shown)


rtorcato (commented):

Fix reads the scope back off the emitted `package.json` (the same field `doctor` reads) instead of `config.projectName`, so the two agree by construction — matches the root cause in #573 and the public-library rule of deriving scope from the consuming repo rather than hardcoding it. Two regression tests cover the scoped and unscoped cases and both, plus the full suite/typecheck/lint, are green on the PR's CI run.

### Before merging
Nothing.

rtorcato (commented):

`familyGlob`'s scope regex (`src/cli/generators/pnpm-workspace.ts:32`, unchanged by this diff) is `/^(@[^/]+)\//` — it accepts any non-slash characters as the scope, including pnpm-matcher metacharacters like `*`. Before this fix, `config.projectName` fed into it was always a bare directory basename (no `/`), so the regex could never match under `--preset` — this glob-injection path was dead code, which is exactly bug #573.

This PR fixes that by reading the *emitted* `package.json` `name` back — which does contain `/` for scoped names, and which `generatePackageJson` preserves verbatim from any pre-existing `package.json` on disk, with no npm-name-format validation anywhere in this path (`setup.ts` only checks non-empty). So a template/repo with a pre-seeded `package.json` name of `"@*/x"` now causes `familyGlob` to write `minimumReleaseAgeExclude: ['@*/*']` into the victim's own `pnpm-workspace.yaml` on `setup --preset library` — silently exempting *every* scoped package from the minimum-release-age delay, not just the repo's own scope. That's the exact supply-chain guard this feature exists to preserve, and this PR is what makes it reachable via an unattended write.

Fix: validate the captured scope against npm's actual scope character set (e.g. `/^(@[a-z0-9-][a-z0-9._-]*)\//i`, or reject anything containing `*`/other glob metachars) before treating it as a family glob.

### Before merging
- Validate the scope regex in `familyGlob` against npm's actual name charset before writing it as a glob — a crafted `package.json` name (e.g. `@*/x`) now reaches this path via `setup` and can exempt every scoped package, not just the repo's own.

rtorcato (commented):

Fix is correct: reading the scope back off the emitted `package.json` makes setup and doctor derive `familyGlob` from the same field, closing the drift in #573. The new scope regex `/^(@[a-z0-9-][a-z0-9._-]*)\/i` rejects all the glob-metacharacter cases from the earlier finding (`@*/x`, `@!(a)/x`, `@ac me/x`, `@[a-z]/x`, `@.acme/x` all return `null`), and the failure path is deliberate and tested (`upsertPnpmSettings` confirmed to omit `minimumReleaseAgeExclude` when `familyGlob` returns `null`) — no accidental silent success.

### Before merging
- The `i` flag makes the regex accept uppercase in the scope (e.g. `@ACME/x`), which the comment above it describes as matching "npm's name charset" — npm scopes are lowercase-only for new packages. Not exploitable (still rejects every glob metacharacter), just slightly looser than the stated intent.

rtorcato (commented):

Verified the fix closes the hole. The new scope regex `/^(@[a-z0-9-][a-z0-9._-]*)\/i` excludes every pnpm/picomatch metacharacter (`*?[]!(){}+,` and space) from both the leading and body character classes, and `/` itself isn't in the charset, so the match can't span past the first slash. Traced `familyGlob`'s only caller in `generators/index.ts`: it reads `name` from the freshly emitted `package.json` (typed `unknown`, guarded by the existing `typeof === 'string'` check), so there's no path using the pre-validation value instead. Case-insensitivity admits non-lowercase scopes but that's over-permissive, not exploitable — an uppercase scope just fails to match the real package's lowercase scope, it doesn't parse as a glob. New tests cover the crafted-metacharacter cases from the prior finding plus the end-to-end `@acme/widget` case.

### Before merging
Nothing.

rtorcato:

Ready to merge — both arms passed, all 5 required checks green. One thing to read first: [`code-reviewer`'s **Before merging**](https://github.com/rtorcato/repo-tooling/pull/580#pullrequestreview-5045080394) — the `i` flag makes the scope regex accept uppercase, which is looser than the comment above it states. Not exploitable; your call whether to tighten it here or let it ride.

PR #581, opened 2026-08-27
feat(lockfile): write an empty rules scaffold so a new repo has rules to edit

### LINKED ISSUE #571: setup never writes a `rules` subtree, so new repos have no stated rules to edit or compare

`setup` never writes a `rules` subtree, so **the repos this tool creates are exactly the ones with no stated rules**.

Three freshly-`setup` repos (nextjs / library / swift) all produced:

```
keys = ["$schema","record","version"]     # rules: ABSENT
```

`rules` is only ever populated by *migration* from a v1–v3 flat lockfile (`src/cli/utils/lockfile.ts:288`), and the writer omits the key entirely when empty (`:303`). A greenfield repo gets `record` only.

That inverts the intent. `rules` is the human-written half — the repo's stated intent, edited by hand and reviewed in PRs (#559) — and the whole point of #563 is comparing it against a reference repo. On a new repo there is nothing to edit and nothing local to compare:

```
$ doctor --rules-from rtorcato/repo-tooling
  not applicable — no rules.aiLoop.agentUser in .repo-tooling.json
  📐 Rules vs rtorcato/repo-tooling — 8 difference(s)
```

The comparison runs, but every difference is "they have rules, we have none". A user following `setup` → edit your rules has no starting point; they have to know the shape and write it from scratch.

## What to change

Have `setup` write a `rules` subtree — even a near-empty one with the keys commented or defaulted, so the file teaches its own shape. `setup --from <owner/repo>` already seeds *config* defaults from another repo; seeding `rules` the same way is the natural companion, and would make "start from another repo's rules" true at creation rather than only at comparison time.

Worth deciding explicitly: whether an absent `rules` should be a doctor `optional-missing`, so a repo with no stated rules says so rather than looking configured.

rtorcato:

`rules` was only ever populated by v1–v3 migration, so a greenfield `setup` produced `record` only — nothing for a human to edit, and nothing local for `doctor --rules-from` to compare against. The repos this tool creates were exactly the ones with no stated rules.

## What changed

`writeLockfile` now emits `DEFAULT_RULES` when the file has no `rules` of its own; an existing `rules` is still carried forward verbatim, untouched and unstamped.

```json
"rules": { "aiLoop": {}, "requiredSkills": [], "mcp": { "recommended": [] }, "exceptions": {} }
```

Every value is empty on purpose. A populated one would be this tool asserting a rule on a repo whose humans have stated none. The containers plus the stamped `$schema` are what teach the shape — an editor autocompletes `agentUser` and the `mcp.recommended` item fields from there.

One consequence, followed through: every repo now carries an `aiLoop` key, so its *presence* no longer means "this repo uses the pipeline". `doctor`'s `requiredSkills` gate reads `rules.aiLoop.agentUser` instead — the login the loop skills have always read (`jq '.rules.aiLoop.agentUser'`). Behaviour only changes for an `aiLoop: {}`, where skipping is the right answer. Docs updated to match.

## Decisions

- **Not seeding `rules` from `setup --from`.** The issue floats it; it would copy another repo's `agentUser` and exceptions into a new repo, which is a stated rule the new repo's humans never stated. `--from` still seeds config defaults only.
- **No doctor `optional-missing` for an absent `rules`.** With the scaffold written on every save, absence now only happens when someone deletes the key by hand — a check for that would fire on nothing.

## Verification

- `pnpm typecheck` ✅
- `pnpm test` ✅ — 1161 passed, 1 expected fail (69 files)
- `pnpm build` ✅
- `pnpm check` — 62 pre-existing CSS `noImportantStyles` warnings under `apps/docs/`, untouched by this branch; `biome check` on the four changed source files is clean

Two existing assertions flipped, both deliberately: `writeLockfile` "omits the key entirely when nothing set it" now asserts the scaffold, and the `setup --from` test asserts the new repo gets the empty scaffold rather than the reference's `agentUser`.

The `runChecks` gate itself has no direct test (it needs `gh`); `checkRequiredSkills` remains unit-tested in `tests/base/required-skills.test.ts`.

Closes #571

### DIFF
--- a/apps/docs/docs/reference/repo-tooling-json.mdx
+++ b/apps/docs/docs/reference/repo-tooling-json.mdx
@@ -89,8 +89,18 @@ enum-valued records — see the schema for their exact shapes.
 ## `rules` — what the humans wrote
 
 Everything under `rules` is edited by hand and reviewed in PRs. It is deliberately outside
-the `writtenBy`/`writtenAt` stamps: the tool never writes here, so it never claims authorship
-of a rule — a hand edit to `rules` leaves the record's provenance true.
+the `writtenBy`/`writtenAt` stamps: the tool never writes a rule here, so it never claims
+authorship of one — a hand edit to `rules` leaves the record's provenance true.
+
+`setup` does write the **empty containers**, so a new repo has the shape in front of it rather
+than a blank page — and `doctor --rules-from` has something local to compare:
+
+```json
+"rules": { "aiLoop": {}, "requiredSkills": [], "mcp": { "recommended": [] }, "exceptions": {} }
+```
+
+Every value is empty on purpose. A default here would be this tool asserting a rule on a repo
+whose humans have not stated one. An existing `rules` is carried forward untouched.
 
 ### `rules.aiLoop` — ai-issue-loop settings
 
@@ -120,7 +130,7 @@ Two rules make it safe to commit:
 - **Never `drift`, never `missing`.** The check probes your machine, not the repo, so a
   contributor with no Claude installed must not fail its `doctor`. It only ever reports
   "not configured", which leaves the exit code alone. It is also skipped entirely unless the
-  file has a `rules.aiLoop` key — that key is already the "this repo uses the pipeline" signal.
+  file sets `rules.aiLoop.agentUser` — that login is the "this repo uses the pipeline" signal.
 - **Check and hint only.** `doctor` names `fix claude-skills`; it never runs it. Committed
   repo config that directs writes into your home directory is the shape of a supply-chain
   attack even when the content is benign, so that step stays a human's.
--- a/src/cli/commands/doctor.ts
+++ b/src/cli/commands/doctor.ts
@@ -326,9 +326,11 @@ async function runBaseChecks(
 	results.push(await checkAiSetup(dir))
 	// User-global, not repo state — see checkClaudeSkills on why it never returns drift.
 	results.push(await checkClaudeSkills(opts.skillsDir))
-	// #533: gated on `aiLoop`, which is already the "this repo uses the pipeline"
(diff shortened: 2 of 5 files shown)


rtorcato (commented):

Scanned for injection, secrets, unsafe shell/SQL, and supply-chain risk in this diff.

`DEFAULT_RULES` is a pure empty-data scaffold (`{}`/`[]` leaves) written as JSON — none of it flows into a shell command, `gh` argument, or glob anywhere in this diff. The gating fix in `doctor.ts` (checking `rules.aiLoop.agentUser` instead of just the `aiLoop` key's presence) correctly avoids the exact assertion risk called out: an empty scaffolded `aiLoop: {}` no longer reads as "this repo uses the pipeline," so no false pipeline-eligibility signal gets created by this scaffold. No dependency or lockfile changes.

### Before merging
Nothing.

rtorcato (commented):

Traced the doctor.ts gate specifically: `lock?.rules?.aiLoop?.agentUser` on a repo with no `rules` at all still short-circuits to `undefined` via optional chaining, same as the old `lock?.rules?.aiLoop` check — an absent `rules` subtree does not newly fail. Confirmed against the diff, not run.

`feat` is earned: this adds real new behavior (scaffolded rules on every new repo) and is non-breaking for existing repos (`existing?.rules` still carried forward verbatim). The one compatibility wrinkle — `rules.aiLoop`'s mere *presence* no longer means "uses the pipeline" — is called out in the PR body and the one known reader (doctor.ts) is fixed to check `.agentUser` instead.

Checked whether `DEFAULT_RULES` being a shared object literal risks cross-repo mutation: no, `writeLockfile` returns only the filepath, and the constructed `lockfile` object is local and discarded after `fs.writeJson` serializes it — not retained or returned.

### Before merging
Nothing.

PR #582, opened 2026-08-27
feat(doctor): flag Swift sources under Sources/ that no declared target claims

### LINKED ISSUE #575: doctor: flag sources under Sources/ that belong to no declared target

rtorcato:

SwiftPM silently ignores a directory under `Sources/` that no target declares:
it is never compiled, its tests never run, and `swift build` / `swift test`
both still exit 0. That is the whole point of this check — it catches a real
defect that is otherwise green, with no red X anywhere. #574 fixed the write
side (the preset no longer rewrites an existing `Package.swift`); this is the
detection side, and it also covers the hand-edit cases — a renamed target, a
dropped `.target(...)` line, a merge that loses one.

## What changed

- `parsePackageSwift` (`src/languages/swift/ci.ts`) now also returns `targets`,
  the declared target names, alongside the products and platforms it already
  extracted. One parser for the manifest rather than a second one.
- New `checkSwiftTargets` in `src/languages/swift/checks.ts` compares those
  names against the subdirectories of `Sources/` and `Tests/`. Anything
  undeclared reports as **`drift`**, naming each orphan.
- `swiftSourceTargets` now delegates to a small `subdirectories(dir, root)`
  helper, which is what lets the check read `Tests/` with the same code.
- Docs: a `Swift targets` row and a prose paragraph in
  `apps/docs/docs/reference/swift.md`, including the limits below.

`Tests/` is included deliberately: an orphaned test directory is the same
failure with a worse consequence — the tests silently never run.

## No fixer, on purpose

The repair is either *add a target* or *delete the directory*, and nothing
outside the project can tell which. So the check carries a hint naming both and
stays out of `fix-targets.ts`, next to `Swift tests` and `Package.swift`, which
are absent for the same reason. `tests/languages/swift/fixers.test.ts` asserts
the uncovered list, so this is enforced rather than just documented.

## What the parser does *not* handle

This is a regex over Swift source, not `swift package dump-package` — running
that would need a Swift toolchain on a machine that is usually just Node. So:

- **Literal names only.** `.target(name: "Widget")` is seen. A target built in
  a loop, behind `#if`, or from a variable is **not**, and its directory would
  be reported as an orphan. This is the one false-positive shape that remains.
- **Any `path:` skips the check entirely.** A custom `path:` decouples
  directory names from target names, so the comparison would be a guess. The
  test is crude — `.package(path: "../Local")` for a local dependency trips it
  too — which costs a skipped check on those repos. A missed finding beats a
  false accusation, and the skip says so in its `detail`.
- `binaryTarget` names are collected even though a binary target owns no source
  directory, for the same reason: the set only ever *suppresses* a report.

Both limits are stated in the source comments, the docs page, and the commit
body rather than implied away.

## Fit with the in-flight Swift work

This layers onto the existing module (`runSwiftChecks`, `SWIFT_FIX_TARGETS`,
the docs check table) and introduces no new structure, so it should not
prejudge how the rest of Swift support lands. The one thing a future
`swift package dump-package`-based parser would replace is `TARGET_DECL` in
`ci.ts`; `checkSwiftTargets` would keep working against a better-populated
`SwiftPackage.targets`.

## Verification

- `vitest run` — 1162 passed, 0 failed (69 files)
- `biome check .` — 0 errors (62 pre-existing `noImportantStyles` warnings in
  `apps/docs` CSS, untouched)

Six new tests cover the orphan case, the orphaned-`Tests/` case, the all-clean
case, an empty package, the `path:` skip, and target-name extraction across
`.target` / `.executableTarget` / `.testTarget` (asserting a `.library` product
name is *not* mistaken for a target). Notably `tests/languages/swift/scaffold.test.ts`
now shows `Swift targets: ok` for the `swift-library` preset — the scaffold it
produces is self-consistent.

Per repo convention I did not run `pnpm install`; `node_modules` was symlinked
from the main checkout to run the commands above.

Closes #575

### DIFF
--- a/apps/docs/docs/reference/swift.md
+++ b/apps/docs/docs/reference/swift.md
@@ -20,6 +20,7 @@ The standard here is the one [`swift-common`](https://github.com/rtorcato/swift-
 | Periphery | `optional-missing` | `periphery` |
 | swift-format | `optional-missing` | `swift-format` |
 | Swift `.gitignore` | `missing` | `swift-gitignore` |
+| Swift targets | `drift` | — (manifest edit, or delete the directory) |
 | Swift tests | `missing` | — (manifest edit, or `swift-ci`) |
 | DocC | `optional-missing` | `docc` |
 | Release automation | `optional-missing` | `swift-release` |
@@ -27,6 +28,20 @@ The standard here is the one [`swift-common`](https://github.com/rtorcato/swift-
 | Pre-push hook | `optional-missing` | `swift-git-hooks` |
 
 `Package.swift` is checked for two things SwiftPM will not infer: a `// swift-tools-version:` comment (without it the manifest doesn't parse) and an explicit `platforms:` clause (without it SwiftPM assumes its oldest supported deployment target, which rejects modern APIs at build time). There's no fixer — rewriting someone's manifest isn't safe, so `doctor` reports and you edit.
+
+`Swift targets` compares the directories under `Sources/` and `Tests/` against
+the target names the manifest declares. A directory no target names is invisible
+to SwiftPM: it is never compiled, its tests never run, and `swift build` /
+`swift test` both still exit 0 — a real defect with no red X anywhere. A renamed
+target, a dropped `.target(...)` line, or a generator that rewrote the manifest
+all produce it.
+
+The manifest is read with a regex, not a Swift toolchain, so the check only sees
+targets declared with a **literal** name — `.target(name: "Widget")`. Targets
+built in a loop, behind `#if`, or from a variable are invisible to it and their
+directories would be reported as orphans. It also skips the whole check on any
+manifest containing a `path:` argument, since a custom path decouples directory
+names from target names entirely.
 
 `Swift tests` has two halves: the manifest must declare a `.testTarget(`, and
 some pipeline (`.github/workflows/*` or `.gitlab-ci.yml`) must actually run
--- a/src/cli/commands/fix-targets.ts
+++ b/src/cli/commands/fix-targets.ts
@@ -80,6 +80,8 @@ const SWIFT_FIX_TARGETS: Record<string, string> = {
 	// `Swift tests` is deliberately absent: when it fails for the manifest half
 	// (no `.testTarget(`) there's nothing to run, and rewriting Package.swift
 	// isn't safe. The check's own hint covers both halves.
+	// `Swift targets` too (#575): the repair is either adding a target or
+	// deleting the directory, and only the project knows which.
 }
 
 /** The same shadowing for the Python module (#290). */
--- a/src/languages/swift/checks.ts
+++ b/src/languages/swift/checks.ts
@@ -17,6 +17,7 @@ import path from 'node:path'
 import fs from 'fs-extra'
 import { type FileCheck, type GitHooksProfile, checkFile } from '../../base/checks.js'
 import type { CheckResult } from '../../base/types.js'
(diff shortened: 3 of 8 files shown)


rtorcato (commented):

Traced target names from `parsePackageSwift` end to end: they only ever land in a `Set.has()` membership check (`checkSwiftTargets` in `src/languages/swift/checks.ts`) and in display strings (`detail`/`hint`). No path join, glob, or shell interpolation uses manifest-derived data — unlike #580's `package.json` name reaching a glob writer, directory names here come from `fs.readdir` on the already-cloned tree, not from the manifest. The `path:` skip regex and `TARGET_DECL` regex are both bounded (no nested quantifiers), so no ReDoS surface. Confirmed no `swift package dump-package` or other execution of the manifest — parsing is pure regex, as stated. The `ci.ts` diff here is limited to the parser; it doesn't touch workflow YAML emission, so no `run:` block injection surface is introduced.

### Before merging
Nothing.

rtorcato (commented):

Confirmed `ci.ts`'s generated-CI code (`swiftGithubJobs`, `renderSwiftWorkflow`, GitLab/release specs) is untouched — this only extends `parsePackageSwift`'s return shape, so no consumer pipeline changes. `feat(doctor):` matches the precedent of the two prior doctor-check commits, so the minor bump is fine.

One stated limit is wrong: the regex scans the whole file, not just `targets:`, so `.target(name: "X")` behind `#if os(...)` (or in a comment) *is* matched, contrary to docs/PR claiming `#if`-guarded targets go undetected. Harmless, errs false-negative — just inaccurate wording.

### Before merging
- Docs/comments say `#if`-guarded `.target()` calls go undetected; they don't (regex isn't scoped to the `targets:` array), so that wording in `ci.ts` and `swift.md` overstates the limitation.

rtorcato:

Ready to merge — both arms passed, all 5 required checks green. Read first: [`code-reviewer`'s **Before merging**](https://github.com/rtorcato/repo-tooling/pull/582#pullrequestreview-5045358486) — the docs and comments overstate a limitation (`#if`-guarded targets *are* matched, since the regex isn't scoped to the `targets:` array). Wording only, errs false-negative.

PR #583, opened 2026-08-27
feat(doctor): flag exports naming files the repo's build cannot emit

### LINKED ISSUE #578: doctor: flag package.json exports/main naming files no build script emits

#570/#577 fixed this for `setup`'s own library preset output, but the general case — any repo's `main`/`module`/`types`/`exports` naming a file its `build` script can't produce — has no doctor check, as #570's body itself notes. `publint` catches it at release time; nothing catches it at setup/doctor time for a repo that later edits `build` or the contract by hand.

Sketch: reuse the emitted-file-per-bundler mapping introduced in `tests/cli/generators/package-json.test.ts` (#577) as a doctor rule, advisory since there's no safe autofix for an arbitrary custom `build`.

rtorcato:

Closes #578

#577 fixed the *preset* that produced the pair. This is the general check for however else it arises — a repo editing `build` or the contract by hand, long after setup ran.

## The check

A new JS check, **Exports buildable**, reports `drift` when a path named in `main`/`module`/`types`/`exports` carries an extension the repo's own `build` script cannot produce. The motivating case: `main: ./dist/index.cjs` and a `require` condition on `./dist/index.d.cts`, while `build` was still plain `tsc`, which emits only `.js` and `.d.ts`. That publishes green and 404s for every CJS consumer.

```
⚠️  Exports buildable — drift
   `tsc` emits .js, .d.ts; package.json names 2 entry point(s) nothing
   produces: ./dist/index.cjs, ./dist/index.d.cts
```

**Judged from the build script, not from `dist/` on disk.** A clean checkout has no `dist/`, and a stale one has whatever the last build left, so neither answers the question.

**Git is what separates the two kinds of path a contract can name.** A tracked `./tooling/preset.mjs` ships as committed source and needs no build; an untracked `./dist/index.cjs` has to come out of `build` or it will not exist at publish time. This is load-bearing, not incidental — without it the check flags **46 valid paths on this repo alone**, whose `exports` point at committed `.mjs`/`.d.mts` presets. Verified against the real checkout: it resolves `build` → `pnpm build-cli` → `rimraf ./dist && tsc …` and reports `ok`.

## Stated limits — where it deliberately says nothing

Inferring emitted extensions from an arbitrary build command is inherently partial, so the check stands down with a reason in `detail` rather than guess. Each of these reports `ok`, never a finding:

| Situation | Why |
|---|---|
| Build command not in the table | Only `tsc`, `tsup`, and the cleaners `rimraf`/`rm`/`del` are recognised |
| `\|`, `\|\|`, `;`, backticks, `$(…)`, redirects | Only a plain `a && b` chain is read |
| Not a git repository | Committed assets and build output become indistinguishable |
| `git ls-files` unavailable | Same |
| Repo has `.cts`/`.mts` sources | `tsc` *does* emit `.cjs`/`.d.cts` from those, so its narrow entry would be wrong |
| Path extension outside `.js`/`.cjs`/`.mjs`/`.d.ts`/`.d.cts`/`.d.mts` | A package publishing `./src/index.ts` directly is not measured against a JS emit table |

**On `tsup` specifically:** only `tsc` narrows the emitted set. `tsup`'s real set is whatever `format`/`dts` say in a config file this does not parse, so it claims the broadest set it *could* emit — a tsup build is effectively passed rather than guessed at, while `rimraf dist && tsup` still resolves instead of standing down. Regex-parsing `tsup.config.ts` to do better is exactly the fragile step that would produce false positives, so it is not done. The table takes one line per new command if that changes.

The bias throughout is under-reporting: a doctor check that cries wolf on a valid config is worse than one with a stated blind spot.

## Not a duplicate of publint

The generated `verify` already runs publint — but at **release** time, against a `dist/` that was just built on a machine where it happens to work. This runs at **setup/doctor** time, on a checkout, before anything is built. Same class, different moment; that gap is the whole point of #578.

## No fixer

Deliberately absent from `FIX_TARGETS`. Both resolutions are valid — point `build` at a bundler that emits those formats, or narrow the contract to what the current build produces — and only the package author knows which one the package promises. The hint names both.

## Shared with the #570 regression test

`declaredEntryPoints` is now exported from `src/languages/js/checks.ts` and the #577 test imports it, deleting its local copy. That test and this check can no longer drift on what the publish contract even is.

## Checks

- `vitest run` — 1168 passed, 1 expected fail, 18 skipped (71 files). Also run as part of pre-push `verify`.
- `tsc --noEmit --project src/cli/tsconfig.json` — clean.
- `biome check .` — exit 0. The 62 warnings are pre-existing `noImportantStyles` hits in `apps/docs` CSS, untouched here.
- `rimraf ./dist && tsc --project src/cli/tsconfig.json` — clean, via pre-push.
- Dogfooded against this repo, as described above.

13 new tests in `tests/languages/js/exports-buildable.test.ts` cover the #570 case, the same contract under tsup, committed assets, the cleaner chain and `pnpm <script>` indirection, and every stand-down path.

Co-authored-by: Claude <noreply@anthropic.com>

### DIFF
--- a/src/cli/commands/doctor.ts
+++ b/src/cli/commands/doctor.ts
@@ -53,6 +53,7 @@ import {
 	checkConfigSchemaVersions,
 	checkDocsSite,
 	checkEnginesNode,
+	checkExportsBuildable,
 	checkGitDependencies,
 	checkKnip,
 	checkLintStaged,
@@ -489,6 +490,8 @@ export async function runDoctor(dir: string, skillsDir?: string): Promise<CheckR
 	results.push(await checkTypedoc(targetDir, pkg))
 	results.push(await checkAreTheTypesWrong(targetDir, pkg))
 	results.push(await checkPublint(targetDir, pkg))
+	// The setup/doctor-time half of what publint catches at release time (#578).
+	results.push(await checkExportsBuildable(targetDir, pkg))
 	results.push(await checkTreeshakeSetup(targetDir, pkg))
 	results.push(await checkPnpmWorkspace(targetDir, pkg))
 	results.push(await checkBuildApprovals(targetDir, pkg))
--- a/src/languages/js/checks.ts
+++ b/src/languages/js/checks.ts
@@ -2,6 +2,7 @@ import path from 'node:path'
 import fs from 'fs-extra'
 import type { BadgeAudience, FileCheck, GitHooksProfile } from '../../base/checks.js'
 import { checkFile, hookHasUncommented } from '../../base/checks.js'
+import { type GitExec, realGitExec } from '../../base/git-identity.js'
 import {
 	CLAUDE_SETTINGS_FILE,
 	readClaudeSettings,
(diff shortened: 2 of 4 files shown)


rtorcato (commented):

Confirmed the build command is only ever parsed (regex/split), never spawned — no path in `emittedExtensions` reaches `child_process`. Confirmed `realGitExec` uses `spawn(..., { shell: false })` with an args array, so shell metacharacters in package.json-derived strings are inert; the `--` before the pathspecs also blocks flag injection into `git ls-files` for a path starting with `-`. Script-indirection following (`build: "pnpm build-cli"`) has cycle detection (`seen` set), so a self-referential `scripts` object can't recurse forever. No ReDoS: `artefactExtension` uses `.endsWith()` over a literal list, not a regex.

One stale-but-harmless note: the comment on `realGitExec` ("Args are internal constants, never user free-text") is no longer accurate now that this check passes package.json-derived strings as args — still safe today because `shell: false` makes argv content-independent, but the comment could mislead a future edit toward `exec`/`shell: true`.

### Before merging
Nothing.

rtorcato (commented):

Test deletion in package-json.test.ts is legitimate dedup: the removed 13-line local declaredEntryPoints helper is replaced by an import of the same (slightly stricter) function now exported from checks.ts; no assertions dropped. Check infers emitted extensions from the build script only, never touches disk. False-positive guards hold: git-tracked paths excluded, unrecognised commands/operators and .cts/.mts sources stand the check down, only plain a && b chains parsed. Verified git ls-files glob (*.cts) matches nested paths, so the .cts/.mts stand-down works as intended.

### Before merging
Nothing.

